PR #23, opened 2015-01-06
Bump up `config.{guess,sub}` to latest

akiss77:
Versions from 2009 & 2011 are way too old, latest versions recognize newer architectures as well, e.g., aarch64.


jdm:
I reverted this change in https://github.com/servo/mozjs/commit/39ed76a3a9d3834a6e1f106d1043efff9b5d6608 because I'm pretty sure it broke our android build automation.

akiss77:
Can you share any info on what that break was? I wouldn't have expected any side effects from the bump-up.

jdm:
This is the failure log from our android builder: http://build.servo.org/builders/android/builds/571/steps/compile/logs/stdio

akiss77:
Thanks for the log! Oh my, I think I found it... `js/src/configure:1385:1398` (and ultimately, `js/src/build/autoconf/android.m4:38:51`) seem to be the cause of the problem. And a missing `*` in the `arm-linux*-android*` pattern.

Since the newer `config.sub` returns `arm-unknown-linux-androideabi` for `$target` (note that extra `-unknown`), the pattern above does not match, so `$android_tool_prefix` will get an incorrect value (that of `$target_os`, which will miss the `arm-` cpu part), which in turn will find a bad compiler (somewhere around `configure:1757`).

If I fix `configure` and `android.m4`, may `config.{guess,sub}` get landed again?

jdm:
Yes.

akiss77:
new PR (with pattern fixes) here: https://github.com/servo/mozjs/pull/26

PR #24, opened 2015-01-06
Add AArch64 as an architecture supported by Double-Conversion.

akiss77:
Tested for correctness as described in the comment.

### DIFF
--- a/mfbt/double-conversion/more-architectures.patch
+++ b/mfbt/double-conversion/more-architectures.patch
@@ -1,7 +1,7 @@
 diff --git a/mfbt/double-conversion/utils.h b/mfbt/double-conversion/utils.h
 --- a/mfbt/double-conversion/utils.h
 +++ b/mfbt/double-conversion/utils.h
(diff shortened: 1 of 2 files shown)


Ms2ger:
Would it work to use <https://mxr.mozilla.org/mozilla-central/source/mfbt/double-conversion/fix-aarch64-macro.patch> instead?

akiss77:
If I see it right, the patch would not apply per se, but the idea (using the macro __AARCH64EL__) can be used. However, I'm not sure why support is limited for little endian variants only. (I do have to admint that my box is little endian as well and have no access to a big endian variant, and have no deep knowledge of the code base either -- but after a quick look at the code, I see nothing against EB...)

akiss77:
So, the question is: shall I try to adapt the above-referenced patch (which is mostly replacing `__aarch64__` with `__AARCH64EL__` only)?

jdm:
Yes, I think I would prefer adapting that patch.

akiss77:
What do you think about the adapted patch?

PR #25, opened 2015-01-07
extern c blocks created around uint8ClampedArray method declarations

ebalint:
It is required if we want to call the uint8ClampedArray methods which are in jsfriendapi.rs


Ms2ger:
You could've done the entire block, but r+

PR #26, opened 2015-01-15
Bump up `config.{guess,sub}` to latest and fix the Android-on-ARM pattern in `configure` and in `android.m4

akiss77:
Versions of `config.{guess,sub}` from 2009 & 2011 are way too old, latest versions
recognize newer architectures as well, e.g., aarch64, so a bump-up is timely. However,
a pattern used in `configure` (originating from `android.m4`) is too restrictive:
 `arm-linux*-android*` does not match the `arm-unknown-linux-androideabi` string
returned by recent `config.sub`, so it needs fixing.


jdm:
Please also submit a pull request to servo/servo that makes use of this too, so we avoid the problem we had last time and this gets tested immediately :)

akiss77:
Is that really servo/servo (where I cannot find anything similar), or should it be https://mxr.mozilla.org/mozilla-central/source/build/autoconf/android.m4 ? (If the latter, I'll have to find out how to propose a patch there.)

jdm:
What I mean is that the servo repository is locked to particular revisions of dependencies (like mozjs) via the Cargo.lock file. You can update these revisions using `./mach update-cargo -p mozjs-sys --precise <hash>` and make a PR for the changes.

akiss77:
FYI: contributed the `android.m4` fix upstream ( https://bugzilla.mozilla.org/show_bug.cgi?id=1122111 )

jdm:
Thanks!

PR #27, opened 2015-02-10
Fix Windows build

klutzy:
This patch fixes three issues on Windows:

-   Set `$(CC)` properly and pass it to `js/src/configure`.
-   Set dummy `$(MOZ_TOOLS)` since `configure` requires non-empty
    value.
-   Fix `$(OUT_DIR)` due to incompatibility between cargo (which uses
    Windows native path e.g. `C:\path`) and msys make (which uses
    `/c/path`).

With such modifications, it succeeded to build on Windows.
(Tested with: rustc 134e00be7 2015-02-09, cargo 9404539 2015-02-09)

### DIFF
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -1,6 +1,7 @@
 ifeq (androideabi,$(findstring androideabi,$(TARGET)))
 
 CC := $(TARGET)-gcc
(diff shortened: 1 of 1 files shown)


jdm:
Looks good!

PR #29, opened 2015-02-19
enable all typed array function

yodalee:
Follow PR #25, enable all typed array function


yodalee:
@Ms2ger in servo `dom/ /codegen/TypedArray.h`, except these 9 array function, there are another two function `ArrayBufferView` and `ArrayBuffer`. Are these two function need to be added too?

PR #37, opened 2015-05-21
Upgrade to Spidermonkey 39

michaelwu:
Well, not really spidermonkey 39. It's just whatever I had pulled from master at the time. Maybe https://github.com/mozilla/gecko-dev/commit/2f60231832aaf5078490bfe56f35bc70ecdc7271 ?

All the imported code goes into the mozjs directory now. There's a linker hack to disable some weak linking in MFBT which doesn't work with the way servo links things together. I'll probably have to discuss that with glandium. There's also a hack to remove PR_GetCurrentThread since that's totally broken if you're not using NSPR. That just needs to be cleaned up a bit and landed upstream.

I'm not entirely sure what to do with the library naming. I would actually prefer if we didn't use a shared mozjs on Android.


jdm:
Note - I included this version of JS_GetAddressableObject in my original mozjs upgrade:

```c++
/*
 * Returns the object that the given candidate pointer points to, if it points
 * to a valid object in this runtime.
 */
JS_FRIEND_API(JSObject *)
JS_GetAddressableObject(JSRuntime *rt, uintptr_t candidateObj)
{
    gc::AllocKind kind;
    void *thing;
    if (IsAddressableGCThing(rt,
                             candidateObj,
                             false,
                             &kind,
                             NULL,
                             &thing) != CGCT_VALID) {
        return NULL;
    }
    if (MapAllocToTraceKind(kind) != JSTRACE_OBJECT) {
        return NULL;
    }
    return reinterpret_cast<JSObject *>(thing);
}
```

jdm:
This existed at the time:
```
+/*
+ * Tests whether w is a (possibly dead) GC thing. Returns CGCT_VALID and
+ * details about the thing if so. On failure, returns the reason for rejection.
+ */
+static inline ConservativeGCTest
+IsAddressableGCThing(JSRuntime *rt, uintptr_t w,
+                     bool skipUncollectedCompartments,
+                     gc::AllocKind *thingKindPtr,
+                     ArenaHeader **arenaHeader,
+                     void **thing)
+{
+    /*
+     * We assume that the compiler never uses sub-word alignment to store
+     * pointers and does not tag pointers on its own. Additionally, the value
+     * representation for all values and the jsid representation for GC-things
+     * do not touch the low two bits. Thus any word with the low two bits set
+     * is not a valid GC-thing.
+     */
+    JS_STATIC_ASSERT(JSID_TYPE_STRING == 0 && JSID_TYPE_OBJECT == 4);
+    if (w & 0x3)
+        return CGCT_LOWBITSET;
+
+    /*
+     * An object jsid has its low bits tagged. In the value representation on
+     * 64-bit, the high bits are tagged.
+     */
+    const uintptr_t JSID_PAYLOAD_MASK = ~uintptr_t(JSID_TYPE_MASK);
+#if JS_BITS_PER_WORD == 32
+    uintptr_t addr = w & JSID_PAYLOAD_MASK;
+#elif JS_BITS_PER_WORD == 64
+    uintptr_t addr = w & JSID_PAYLOAD_MASK & JSVAL_PAYLOAD_MASK;
+#endif
+
+    Chunk *chunk = Chunk::fromAddress(addr);
+
+    if (!rt->gcChunkSet.has(chunk))
+        return CGCT_NOTCHUNK;
+
+    /*
+     * We query for pointers outside the arena array after checking for an
+     * allocated chunk. Such pointers are rare and we want to reject them
+     * after doing more likely rejections.
+     */
+    if (!Chunk::withinArenasRange(addr))
+        return CGCT_NOTARENA;
+
+    /* If the arena is not currently allocated, don't access the header. */
+    size_t arenaOffset = Chunk::arenaIndex(addr);
+    if (chunk->decommittedArenas.get(arenaOffset))
+        return CGCT_FREEARENA;
+
+    ArenaHeader *aheader = &chunk->arenas[arenaOffset].aheader;
+
+    if (!aheader->allocated())
+        return CGCT_FREEARENA;
+
+    if (skipUncollectedCompartments && !aheader->zone->isCollecting())
+        return CGCT_OTHERCOMPARTMENT;
+
+    AllocKind thingKind = aheader->getAllocKind();
+    uintptr_t offset = addr & ArenaMask;
+    uintptr_t minOffset = Arena::firstThingOffset(thingKind);
+    if (offset < minOffset)
+        return CGCT_NOTARENA;
+
+    /* addr can point inside the thing so we must align the address. */
+    uintptr_t shift = (offset - minOffset) % Arena::thingSize(thingKind);
+    addr -= shift;
+
+    if (thing)
+        *thing = reinterpret_cast<void *>(addr);
+    if (arenaHeader)
+        *arenaHeader = aheader;
+    if (thingKindPtr)
+        *thingKindPtr = thingKind;
+    return CGCT_VALID;
+}
```

metajack:
@bors-servo r=larsbergstrom,metajack

I reviewed the final commit.

michaelwu:
Merging directly since the default build won't work. -j4 appears to cause us to run out of resources on the build machine, and lower numbers cause the build to die by taking too long.

PR #48, opened 2015-07-02
Improve ARM cross compilation

mmatyas:
The first patch fixes some configuration flags for ARM, which also means `jemalloc` is now disabled for all targets.
The second patch fixes zlib flags, which caused moz.build errors during cross compilation. I'm not sure if this is the correct way to solve this of if this works for all platforms, could you check it?

Requires servo/servo#6531
Part of servo/servo#6327

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/48)


mmatyas:
Updated as described in servo/servo#6537

PR #54, opened 2015-09-22
Fix build under MSYS2/MINGW64

vvuk:
This is a set of changes needed to build mozjs under msys2/mingw64.

Core changes:
- Allow building with msys make
- Don't strip "lib" prefix from library names
- Use system nspr with correct lib names (nspr-config gives incorrect values)
- Force using native win32 python instead of the mingw64/msys python

I'm not really happy with these changes since these are all basically core m-c js files, and I don't think we could take some of these changes in m-c.  But, the majority of changes are correct or are resolving deeper issues in the m-c build system (e.g. the need to use the native win32 python instead of the msys one).

The ICU libprefix stuff can probably be resolved a different way, I just wasn't sure how to go about it.

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/54)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,6 +7,7 @@ use std::process::{Command, Stdio};
 
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
+    let target = env::var("TARGET").unwrap();
     let result = Command::new("make")
         .args(&["-R", "-f", "makefile.cargo"])
         .stdout(Stdio::inherit())
@@ -15,7 +16,14 @@ fn main() {
         .unwrap();
     assert!(result.success());
     println!("cargo:rustc-link-search=native={}/dist/lib", out_dir);
-    println!("cargo:rustc-link-lib=static=js_static");
+    if target.contains("windows") {
+        // On Windows, because dynamic libs and static libs end up
+        // with different symbols for the import, we have to build
+        // with the shared mozjs DLL.
+        println!("cargo:rustc-link-lib=mozjs");
+    } else {
+        println!("cargo:rustc-link-lib=static=js_static");
+    }
     println!("cargo:rustc-link-lib=stdc++");
     println!("cargo:outdir={}", out_dir);
 }
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -28,26 +28,67 @@ ifneq (,$(CARGO_FEATURE_DEBUGMOZJS))
 	CONFIGURE_FLAGS += --enable-debug=-g --enable-optimize=-O0
 endif
 
-ifeq (windows,$(findstring windows,$(TARGET)))
+ifeq ($(MSYSTEM),MINGW64)
 	# msys2 sets CC=cc as default. however, there is no `cc.exe`.
 	# overwrite it here.
 	ifeq ($(CC),cc)
 		CC = gcc
 		CPP = gcc -E
 	endif
+
 	# cargo uses Windows native path. msys2 make unfortunately doesn't understand it.
 	OUT_DIR:=$(shell cygpath "$(OUT_DIR)")
+
 	# on windows, SM requires moztools which contains prebuilt libraries like glib and libIDL.
-	# we don't need them, so just set any fake path.
-	MOZ_TOOLS=$(OUT_DIR)
+	# we don't need them, so just set / -- configure will check that $(MOZ_TOOLS)/bin exists
+	# and that it's in the path.
+	MOZ_TOOLS=/
+
+	# This is a mess.  m-c/spidermonkey includes its own copy of virtualenv.py,
+	# which doesn't have the fixes needed to make it work with msys2/mingw64.
+	# The only way we can successfully build under msys2 is to use the win32-native
+	# python.
+
+	ifneq (,$(wildcard c:/python27/python.exe))
+	FORCED_PYTHON := "c:/python27/python.exe"
+	else ifneq(,$(NATIVE_WIN32_PYTHON))
+	FORCED_PYTHON := "$(NATIVE_WIN32_PYTHON)"
+	else
+	$(message You must either have the Native Win32 python installed in C:/python27, or set NATIVE_WIN32_PYTHON to point to the appropriate python.exe.)
+	$(message Download the Python installer from  https://www.python.org/downloads/release/python-2710/)
+	$(error Can't find native Win32 python)
+	endif
+
+	# on windows/mingw64, we need to not use virtualenv -- the virtualenv.py script
+	# that ships with m-c doesn't work on mingw.
+	ifeq ($(MSYSTEM),MINGW64)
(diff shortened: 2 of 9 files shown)


metajack:
I'm mostly concerned about the upstream build system modifications. Have you talked to the JS team about these changes and issues?

cc @tschneidereit

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[makefile.cargo, line 49 \[r1\]](https://reviewable.io:443/reviews/servo/mozjs/54#-JzpjfhO-vp2uEHhKOxU)** ([raw file](https://github.com/servo/mozjs/blob/297406ecc54591a44ef83ad81a2518a82e00556d/makefile.cargo#L49)):</sup>
Is this path guaranteed to exist or just where you happened to install the tool?

---

<sup>**[makefile.cargo, line 58 \[r1\]](https://reviewable.io:443/reviews/servo/mozjs/54#-Jzpjmxx2uiGQuBrAOUF)** ([raw file](https://github.com/servo/mozjs/blob/297406ecc54591a44ef83ad81a2518a82e00556d/makefile.cargo#L58)):</sup>
Our other tools are tolerant of python3 vs. python2 stuff. Is it possible not to hardcode a python version here?

Also this is the only difference in the two branches. Could we pull that out and then only have one all rule?

---

<sup>**[mozjs/config/baseconfig.mk, line 34 \[r1\]](https://reviewable.io:443/reviews/servo/mozjs/54#-JzpkFkWhYx3XtiHga5R)** ([raw file](https://github.com/servo/mozjs/blob/297406ecc54591a44ef83ad81a2518a82e00556d/mozjs/config/baseconfig.mk#L34)):</sup>
We don't typically change these files since they are from upstream. Does upstream specifically forbid msys?

---

<sup>**[mozjs/config/external/icu/Makefile.in, line 30 \[r1\]](https://reviewable.io:443/reviews/servo/mozjs/54#-JzpkTLPiz5kI2wdpuiZ)** ([raw file](https://github.com/servo/mozjs/blob/297406ecc54591a44ef83ad81a2518a82e00556d/mozjs/config/external/icu/Makefile.in#L30)):</sup>
is the `libs` intentional? This looks like a typo. For example it's `lib/s` in the else branch.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/mozjs/54)

tschneidereit:
It'd be great if you could split these changes into upstreamable and hacky bits. Then we can take the former on mozilla-central and deal with the hacky bits later. It's not like we don't have build system hacks already.

Ms2ger:
And rebase, please.

vvuk:
Rebased and commits split out.

The missing #! to milestone.py, and the allowing of msys make for mingw64/msys2 could be upstreamed.  The latter patch would ideally need a reliable way of actually detecting MSYS2 -- mozilla-build currently has $MSYSTEM = MINGW32 for historical raisins, even though that's mostly a lie.

The ICU changes are... correct, but it depends on what you're trying to accomplish.  If we're trying to use mingw to create windows-style libraries, then they're not correct (we don't want lib* as a prefix).  But if we're trying to make static libs, then they are.  The right thing here might be to have an entirely separate platform config file that's something like mingw64-msys2, and then find a way to explicitly tell ICU to use that config.

The makefile.cargo fixes are unfortunate, for the python change.  But I couldn't find any way to get around it, given that the spidermonkey tree has its own copy of virtualenv.py that it really wants to use.

adamncasey:
Would be nice to see this added soon, so servo on windows can move forward!

jdm:
I'd like @metajack to look at the new changes, but he's out all week :/

metajack:
I appreciate the extra comments in the weird spots. There are still two comments that appear unanswered.

In makefile.cargo on line 49 there is a hardcoded path, which I asked if that was guaranteed to be there.

In the icu makefile there seems to be a typo between "lib" and "libs" in the two branches. Is that intentional or an error?


---

Reviewed 5 of 6 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[build.rs, line 23 \[r3\]](https://reviewable.io:443/reviews/servo/mozjs/54#-K10Mv56A_M6Ff694Zge)** ([raw file](https://github.com/servo/mozjs/blob/265c176471e5df0a9025add3ebf6e42394cad10b/build.rs#L23)):</sup>
Why do we only need to link nspr on windows?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/mozjs/54)

vvuk:
> In makefile.cargo on line 49 there is a hardcoded path, which I asked if that was guaranteed to be there.

The default install location is now hardcoded, along with a NATIVE_WIN32_PYTHON env var fallback.  There's a message that gets printed if python can't be found using either of the two.

> In the icu makefile there seems to be a typo between "lib" and "libs" in the two branches. Is that intentional or an error?

Intentional -- both branches are moving libraries that have an "s" prefix to ones that don't  (libsicu -> libicu, or sicu -> icu).

> build.rs, line 23 [r3] (raw file): Why do we only need to link nspr on windows?

JS doesn't have the "mini nspr" for Windows that posixy platforms have in vm/PosixNSPR.{cpp,h}  We need to link with real nspr until someone writes that code... it shouldn't be difficult to do, other than the lack of native condvars on Windows.

metajack:
The path I am talking about is `/mingw64/include/nspr` which is not related to python. What about that path?

Other than that, you've answered the questions I had.

vvuk:
Oh, sorry, was looking at the wrong line.  That path is where the nspr includes will be from the mingw64 nspr package.  They will always be in that location (/mingw64/include/nspr will get mapped to wherever the base msys2 install happened in the actual filesystem).

metajack:
Squash and r=me


---

Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/mozjs/54)

vvuk:
I'm working on changing the linking here a bit -- figured out how to link with nspr statically, to avoid the DLL deps.  I'll update this PR once things check out.

vvuk:
static nspr linking didn't work out -- it has issues with threading, since some of the nspr threading functionality depends on DllMain being notified of thread attach/detach.

metajack:
@vvuk Is this ready for re-review?

vvuk:
Sorry, yes -- this should be reasy for re-review.

metajack:
@bors-servo r+


---

Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/mozjs/54)

PR #55, opened 2015-09-24
Update SM headers to reflect the headers used in bindings generation

michaelwu:
This isn't necessary to use the new bindings, but allow people other than myself to generate the bindings.

CC @tschneidereit

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/55)


mozjs/js/public/StructuredClone.h
```diff
@@ -29,8 +29,10 @@ enum TransferableOwnership {
     // Structured clone buffer does not yet own the data
     SCTAG_TMO_UNOWNED = 1,
 
+#ifndef RUST_BINDGEN
```
tschneidereit: It'd be good to have comments on stuff like this. We can do that during upstreaming, though.

michaelwu:
I think this is useful as a reference, but we don't have to merge it. It should be obsolete at some point since we want to upstream it.

PR #56, opened 2015-10-01
Updates to make mozjs work with newer SM versions

tschneidereit:
Not quite ready to land, but close.

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/56)


metajack:
@tschneidereit Does this still need more work?

PR #65, opened 2015-11-24
Backport of bug 1224722 - Enable executing multiple embedding-provided inputs as self-hosted JS during runtime startup

tschneidereit:
@michaelwu wanted to use this in his self-hosting experiments.

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/65)

### DIFF
--- a/mozjs/js/src/builtin/TestingFunctions.cpp
+++ b/mozjs/js/src/builtin/TestingFunctions.cpp
@@ -2440,6 +2440,22 @@ DumpStringRepresentation(JSContext *cx, unsigned argc, Value *vp)
 }
 #endif
 
+static bool
+GetSelfHostedValue(JSContext* cx, unsigned argc, Value* vp)
+{
+    CallArgs args = CallArgsFromVp(argc, vp);
+
+    if (args.length() != 1 || !args[0].isString()) {
+        JS_ReportError(cx, "First argument should be a string");
+        return false;
+    }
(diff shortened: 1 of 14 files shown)


Ms2ger:
We should just do a full update, except nobody seems to want to do that or explain how to do it.

tschneidereit:
> We should just do a full update, except nobody seems to want to do that or explain how to do it.

I agree in principle. However, applying the hacks that make the current mozjs work for Servo on top of current mozilla-central is non-trivial, and won't ever get easier, really. Hence, I'd like to instead get rid of as many of these hacks as possible, partly by working around the need for them, partly by upstreaming proper fixes. I started looking into this a few weeks back, and it seems feasible. In fact, it's one of my Q4 goals to get to a point where we can use pristine upstream and do away with mozjs in its current form entirely. (Whether I can manage to actually get to that point this quarter remains to be seen. I *did* lose a few weeks because of that stupid accident.)

For the time being, backporting this patch took about 10 minutes, so seemed to be well-worth it.

michaelwu:
I'm getting an assertion at MOZ_ASSERT(hasContexts()); in JSRuntime::evaluateSelfHosted. Should it be the opposite?

michaelwu:
MOZ_ASSERT(selfHostingGlobal_); is also failing. Guess I'm probably doing something wrong..

michaelwu:
Ok, seems to work.

PR #71, opened 2016-01-06
Fix build for win32, remove --enable-gczeal that snuck back in

vvuk:
For some reason, gcc isn't generating public symbols for setTracingLocation if they're defined inline in the class declaration in the header.  There's probably an attribute we could apply or something, but this was the simplest way to get this to work.

[Review on Reviewable](https://reviewable.io/reviews/servo/mozjs/71)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,13 +16,11 @@ fn main() {
         .unwrap();
     assert!(result.success());
     println!("cargo:rustc-link-search=native={}/dist/lib", out_dir);
+    println!("cargo:rustc-link-lib=static=js_static");
     if target.contains("windows") {
-        // On Windows, because dynamic libs and static libs end up
-        // with different symbols for the import, we have to build
-        // with the shared mozjs DLL.
-        println!("cargo:rustc-link-lib=mozjs");
-    } else {
-        println!("cargo:rustc-link-lib=static=js_static");
+        println!("cargo:rustc-link-lib=static=mozglue");
+        println!("cargo:rustc-link-lib=winmm");
+        println!("cargo:rustc-link-lib=psapi");
     }
     println!("cargo:rustc-link-lib=stdc++");
     println!("cargo:outdir={}", out_dir);
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -63,19 +63,8 @@ ifeq ($(MSYSTEM),MINGW64)
 	$(error Can't find native Win32 python)
 	endif
 
-	# on windows/mingw64, we need to not use virtualenv -- the virtualenv.py script
-	# that ships with m-c doesn't work on mingw.
-	ifeq ($(MSYSTEM),MINGW64)
-	DONT_POPULATE_VIRTUALENV=1
-	endif
+	CONFIGURE_FLAGS += --disable-shared-js --disable-export-js
 
-	# we build with system nspr, expecting a msys2 mingw64 build
-	# The mingw64 nspr-config gives -lplc4 for example, but only libplc4.dll.a and libplc4_s.a exist.
-	# We can't use static NSPR because it doesn't properly work with threads.  This would all
-	# be fixed once a Windows equivalent to PosixNSPR.{cpp,h} is written.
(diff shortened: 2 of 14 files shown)


vvuk:
Someone should test this on linux/osx with the removing of --enable-gczeal, but without the second patch -- we should see if this is a windows-specific gcc issue, or if we're doing something we shouldn't be.

jdm:
I'd like @michaelwu to take a look at the second commit.

vvuk:
I think the original linking problem may be due to -fno-keep-inline-dllexport that is added to the gcc flags in configure... but haven't tested without, would rather get static linking working properly.

asajeffrey:
Should this be two separate PRs? One to remove --enable-gc-zeal and one to fix building on windows?

vvuk:
Sure? Note that I'm working through a bunch of this stuff in

https://bugzilla.mozilla.org/show_bug.cgi?id=1238203
https://bugzilla.mozilla.org/show_bug.cgi?id=1238023
https://bugzilla.mozilla.org/show_bug.cgi?id=1237761

so that we get it all in upstream.

michaelwu:
An alternative to moving setTracingLocation and other functions is to remove those from the bindings - I don't think we need those functions, and they appear to be gone from upstream SM anyway. I'm fine with either solution though, since these functions are going to disappear in the future.

larsbergstrom:
Currently failing on linux with:
```
In file included from /home/larsberg/servo/target/debug/build/mozjs_sys-6e42204b6e19049d/out/js/src/Unified_cpp_js_src7.cpp:110:0:
/home/larsberg/mozjs/mozjs/js/src/jsapi.cpp: In function ‘bool JS_CallOnce(JSCallOnceType*, JSInitCallback)’:
/home/larsberg/mozjs/mozjs/js/src/jsapi.cpp:5830:80: error: ‘PR_CallOnceWithArg’ was not declared in this scope
     return PR_CallOnceWithArg(once, CallOnce, JS_FUNC_TO_DATA_PTR(void *, func)) == PR_SUCCESS;
                                                                                ^
make[4]: *** [Unified_cpp_js_src7.o] Error 1
make[3]: *** [js/src/target] Error 2
make[2]: *** [compile] Error 2
make[1]: *** [default] Error 2
make: *** [all] Error 2
thread '<main>' panicked at 'assertion failed: result.success()', build.rs:17
```

larsbergstrom:
@vvuk You may already know, but if not, here's a note on the hackery so you don't need to allocate/delete critical sections - http://joeduffyblog.com/2006/11/28/windows-keyed-events-critical-sections-and-new-vista-synchronization-features/

vvuk:
Arg, PR_CallOnceWithArg went away in m-c sm, but I guess not here.  I'll fix shortly.  Same with just removing those methods from the tracer.

As for critical sections... I had no idea about SRWLs.  Since we're already depending on >= Vista, I'll switch to using them.

larsbergstrom:
It looks like this is passing CI builds now. @vvuk Is this ready to be reviewed/landed?

larsbergstrom:
@vvuk and @michaelwu Is there anything more that needs to be done here to land this? I think it's the last piece we need before we land the Win32 changes to Servo master.

michaelwu:
I'm happy with the approach here.

larsbergstrom:
@vvuk, @jayflux and I are now getting missing `setTracingLocation` and friends with this and the landed rust-mozjs fixes. Is there something we're missing?

vvuk:
Did rust-mozjs changes land, that had the removal of that in the bindings?

vvuk:
https://github.com/servo/rust-mozjs/pull/236

larsbergstrom:
Great, thanks! That did it for me :-)

I'm opening the servo/servo PR, since it can land independently of these changes (which switch to static linking). That will let me start working on the CI/builder stuff.

larsbergstrom:
@bors-servo r=michaelwu

(confirmed on IRC with @jdm and @Ms2ger)

PR #74, opened 2016-03-31
Compile targeting modern ARM processors on Android for hardfloat

larsbergstrom:
r? @Ms2ger 

This ensures that when SM has any floating point operations on Android in the C code, they're compiled to use SIMD instructions and/or FP registers instead of integer registers or fallback code. Big perf win.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/74)


Ms2ger:
If you say so.

@bors-servo r+

PR #80, opened 2016-06-05
Compile successfully on FreeBSD

danlrobertson:
- Add static MAKE_CMD variable for the make command
 - replace instances of `make` with $(MAKE)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/80)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,12 +3,22 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use std::env;
+use std::ffi::OsStr;
 use std::process::{Command, Stdio};
+
+
+fn find_make<'a>() -> &'a str {
+    match Command::new("gmake").status() {
+        Ok(_) => "gmake",
+        Err(_) => "make",
+    }
+}
 
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     let target = env::var("TARGET").unwrap();
-    let result = Command::new("make")
+    let result = Command::new(env::var_os("MAKE")
+            .unwrap_or_else(|| OsStr::new(find_make()).to_os_string()))
         .args(&["-R", "-f", "makefile.cargo"])
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -84,13 +84,13 @@ all:
 	PYTHON="$(FORCED_PYTHON)" \
 	MOZ_TOOLS="$(MOZ_TOOLS)" CC="$(CC)" CPP="$(CPP)" CXX="$(CXX)" AR="$(AR)" \
 	$(SRC_DIR)/mozjs/js/src/configure $(strip $(CONFIGURE_FLAGS)) || (cat config.log && exit 1)
-	cd $(OUT_DIR) && make -f Makefile -j$(NUM_JOBS)
+	cd $(OUT_DIR) && $(MAKE) -f Makefile -j$(NUM_JOBS)
 else
 all:
 	touch $(SRC_DIR)/mozjs/js/src/configure
 	cd $(OUT_DIR) && \
 	MOZ_TOOLS="$(MOZ_TOOLS)" CC="$(CC)" CPP="$(CPP)" CXX="$(CXX)" AR="$(AR)" \
 	$(SRC_DIR)/mozjs/js/src/configure $(strip $(CONFIGURE_FLAGS)) || (cat config.log && exit 1)
-	cd $(OUT_DIR) && make -f Makefile -j$(NUM_JOBS)
+	cd $(OUT_DIR) && $(MAKE) -f Makefile -j$(NUM_JOBS)
 endif
 


build.rs
```diff
@@ -5,10 +5,17 @@
...
+static MAKE_CMD: &'static str = "gmake";
+#[cfg(not(target_os = "freebsd"))]
+static MAKE_CMD: &'static str = "make";
+
+
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     let target = env::var("TARGET").unwrap();
-    let result = Command::new("make")
+    let result = Command::new(MAKE_CMD)
```
mbrubeck: Maybe this should use the `$MAKE` env var if it's set (and fall back to the constant if it's not), so people on other BSDs or similar systems can more easily build without patching the build script.

build.rs
```diff
@@ -5,10 +5,17 @@
 use std::env;
 use std::process::{Command, Stdio};
 
+
+fn find_make() -> String {
+    Command::new("gmake").status()
+                         .map(|_| "gmake".to_string())
+                         .unwrap_or("make".to_string())
```
danlrobertson: Is this the best way to find if `gmake` exists?

emilio: I'll let @mbrubeck do the final review, though if it works it's probably ok.

For reference, [this](https://github.com/servo/servo/blob/master/components/style/build.rs#L26) is what it's used to find python in the style crate, though your version looks way cleaner.

build.rs
```diff
@@ -5,10 +5,17 @@
...
+}
+
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     let target = env::var("TARGET").unwrap();
-    let result = Command::new("make")
+    let result = Command::new(env::var("MAKE").unwrap_or(find_make()))
```
emilio: nit: `unwrap_or_else(find_make)` won't call the function if `MAKE` is found.

danlrobertson: Ah! good call

build.rs
```diff
@@ -5,10 +5,17 @@
...
+    Command::new("gmake").status()
+                         .map(|_| "gmake".to_string())
+                         .unwrap_or("make".to_string())
+}
+
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
     let target = env::var("TARGET").unwrap();
-    let result = Command::new("make")
+    let result = Command::new(env::var("MAKE").unwrap_or_else(find_make()))
```
emilio: nit again: you're still calling the function, remove the two parens because it's not building now :P

danlrobertson:
Updated

mbrubeck:
@bors-servo delegate+

r=mbrubeck with or without some optional suggestions below

<details><summary><i>Previously, danlrobertson (Dan Robertson) wrote…</i></summary>
> Updated
</details>


---

Reviewed 1 of 2 files at r1, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[build.rs, line 12 \[r2\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJXM_EwZC6uD_zdVOAq-r2-12:-KJaqJJaBu9ilSBAIFWP:-163488015) ([raw file](https://github.com/servo/mozjs/blob/d5114d83e196730ce8d965026580ef81c01e55dc/build.rs#L12)):*
<details><summary><i>Previously, emilio (Emilio Cobos Álvarez) wrote…</i></summary>
> I'll let @mbrubeck do the final review, though if it works it's probably ok.
> 
> 
> 
> For reference, [this](https://github.com/servo/servo/blob/master/components/style/build.rs#L26) is what it's used to find python in the style crate, though your version looks way cleaner.
</details>
This seems fine to me. My only other idea is to iterate through `["gmake", "make"]` and try running the actual command with each one, continuing if it returns an `std::io::ErrorKind::NotFound` and breaking on any other result.

---

*[build.rs, line 18 \[r3\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJXM_EwZC6uD_zdVOAq-r3-18:-KJarEdh0iochUXnlbGg:747376210) ([raw file](https://github.com/servo/mozjs/blob/834cf116c4e7ff35fc363462c19530130ef55906/build.rs#L18)):*
<details><summary><i>Previously, emilio (Emilio Cobos Álvarez) wrote…</i></summary>
> nit again: you're still calling the function, remove the two parens because it's not building now :P
</details>
nit: Could use `env::var_os` here, I think.

---

*[build.rs, line 11 \[r4\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJaqAu7rPrlx-8BTxDf:-KJaqAu8qZ0P6Yh8kQR_:1000544798) ([raw file](https://github.com/servo/mozjs/blob/d294feb044c707ab26352f1331df29da21a5c7cb/build.rs#L11)):*
> ```Rust
> fn find_make() -> String {
>     Command::new("gmake").status()
>                          .map(|_| "gmake".to_string())
> ```

Personally I'd find a `match` slightly more readable than `map`/`unwrap_or` here, but it's up to you.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/mozjs/80#-:-KJaqxsA3bUokvA_VXj5:-643861574)*

danlrobertson:
Thanks for the good suggestions!

<details><summary><i>Previously, mbrubeck (Matt Brubeck) wrote…</i></summary>
> @bors-servo delegate+
> 
> r=mbrubeck with or without some optional suggestions below
</details>


---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[build.rs, line 12 \[r2\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJXM_EwZC6uD_zdVOAq-r2-12:-KJrmuY8AObW3zoE14HD:-1124475087) ([raw file](https://github.com/servo/mozjs/blob/d5114d83e196730ce8d965026580ef81c01e55dc/build.rs#L12)):*
<details><summary><i>Previously, mbrubeck (Matt Brubeck) wrote…</i></summary>
> This seems fine to me. My only other idea is to iterate through `["gmake", "make"]` and try running the actual command with each one, continuing if it returns an `std::io::ErrorKind::NotFound` and breaking on any other result.
</details>
This would add more information in the case that neither one is present (although this is almost certainly very rare). I like the idea, but I think I'll stick with a `match` and `Command::status`

---

*[build.rs, line 18 \[r3\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJXM_EwZC6uD_zdVOAq-r3-18:-KJrnXq0TppMR25UF9pp:361414833) ([raw file](https://github.com/servo/mozjs/blob/834cf116c4e7ff35fc363462c19530130ef55906/build.rs#L18)):*
<details><summary><i>Previously, mbrubeck (Matt Brubeck) wrote…</i></summary>
> nit: Could use `env::var_os` here, I think.
</details>
I'll try it out and see what happens!

---

*[build.rs, line 11 \[r4\]](https://reviewable.io:443/reviews/servo/mozjs/80#-KJaqAu7rPrlx-8BTxDf:-KJrmOae0QKxiLyH9zWq:-161872713) ([raw file](https://github.com/servo/mozjs/blob/d294feb044c707ab26352f1331df29da21a5c7cb/build.rs#L11)):*
<details><summary><i>Previously, mbrubeck (Matt Brubeck) wrote…</i></summary>
> Personally I'd find a `match` slightly more readable than `map`/`unwrap_or` here, but it's up to you.
</details>
Not that I think about it, I agree.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/mozjs/80#-:-KJrnaSxl3C5_DIpsQIG:-1957612003)*

danlrobertson:
^ highfive bug?

jdm:
Homu instead of highfive, but yes.

danlrobertson:
@bors-servo r=mbrubeck

<details><summary><i>Previously, jdm (Josh Matthews) wrote…</i></summary>
> Homu instead of highfive, but yes.
</details>


---

Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/mozjs/80#-:-KJw6d_vSOde1woiUD7G:436296857)*

danlrobertson:
servo/angle#8 failed on appveyor... currently investigating

PR #88, opened 2016-08-09
Prepend MOZTOOLS_BIN to PATH, not append

vvuk:
We need the moz tools to be at the start of the path, to not pick up random things along the way.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/88)


vvuk:
@bors-servo r+

(no-one uses MOZTOOLS_BIN yet, hope the self-r+ is ok)

larsbergstrom:
Passed: https://travis-ci.org/servo/mozjs/builds/150921266

PR #89, opened 2016-08-12
Enable native promises.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/89)


nox:
Please put the promise thing behind a feature flag (so that we don't tie urgent smup fixes to promises) and add the cherry-picked commit to https://github.com/servo/mozjs/tree/master/etc/patches. r=me when that is done.

makefile.cargo
```diff
@@ -4,6 +4,7 @@ SHELL := /bin/bash
 
 # Default flags
 CONFIGURE_FLAGS := \
+        --enable-sm-promise \
```
nox: Indentation. That should be a tab no?

PR #98, opened 2016-08-24
Add instructions for testing that the bindings actually build correctly with the upgraded mozjs lib

fitzgen:
r? @nox

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/98)


Ms2ger:
Obsoleted by#100.

PR #99, opened 2016-09-08
smup fixes for Windows

vvuk:
Two of these apply to upstream; I've filed bug 1300925 and a comment on bug 1288313 for the WINNT issue (I should really file a separate bug).  The other two are rust-build-related fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/99)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -20,7 +20,7 @@ fn find_make() -> OsString {
 }
 
 fn main() {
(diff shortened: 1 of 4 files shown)


Ms2ger:
WINNT is fixed upstream in https://bugzilla.mozilla.org/show_bug.cgi?id=1289946

PR #109, opened 2016-12-22
More portable way to set SHELL to bash.

eddyb:
This is the only problem I found while trying to build on NixOS (the usual problem from the unpatched virtualenv that SpiderMonkey bundles seems to have fixed itself so 🎉).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/109)


Ms2ger:
> `<ted>` Ms2ger: FYI it looks like gmake will do a PATH search for you if you just do `SHELL := bash`

PR #128, opened 2017-11-02
Add `exclude` to `Cargo.toml`

fitzgen:
Excludes tests and documentation, which isn't necessary to build, and therefore not necessary for the .crate tarball.

Excluding ICU causes the build to fail. We can investigate this later, if we feel the need. We now have a 50M limit for our crate on crates.io, so we should be able to publish now either way.

r? @SimonSapin

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/128)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,13 @@ authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
 license = "MPL-2.0"
+exclude = [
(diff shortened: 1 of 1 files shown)


fitzgen:
This brought the .crate tarball down from 35M down to 28M, for the record.

fitzgen:
0.50.0 @ https://crates.io/crates/mozjs_sys 🎉

PR #119, opened 2017-06-02
WIP smup

fitzgen:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/119)

PR #130, opened 2017-12-18
Update SpiderMonkey

jdm:
Rebased from #119. Does not incorporate any newer changes from rust-mozjs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/130)


jdm:
The configure script now complains that GCC needs to be >=4.9.

atouchet:
Can this be closed now?

PR #131, opened 2018-01-12
Upgrade to newer Android NDK versions

ricochet1k:
Since r12 I think, the libcxx folder doesn't exist anymore. This lets mozjs_sys build with cargo-apk.

I didn't see any obvious way to predicate on the NDK revision.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/131)


jdm:
This is a bit complicated, since Servo is still using a version prior to r12, and merging this would mean that we couldn't update mozjs until we update.

jdm:
Closing; #138 should support newer android NDKs.

PR #132, opened 2018-02-26
Use python2 for the build whenever possible and bump the version.

### LINKED ISSUE #129: Build fails on Arch Linux unless PYTHON env. variable is set
`cargo build` bails out with 
```
--- stderr
  File "/home/marcin/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.50.0/mozjs/js/src/../../configure.py", line 83
    os.chmod('config.status', 0755)
                                 ^
SyntaxError: invalid token
cat: config.log: No such file or directory
make: *** [makefile.cargo:140: maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', /home/marcin/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.50.0/build.rs:53
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
No such error occurs if the invocation is `PYTHON=python2 cargo build`.

marmistrz:
If python2 seems to be a valid Python 2 interpreter, the configure
script will use it. Otherwise, it will fall back to python, as
previously.

We also bump the crate version so that this change can be propagated
further the dependency tree.

Closes #129.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/132)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
-version = "0.50.0"
+version = "0.50.1"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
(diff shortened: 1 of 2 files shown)


jdm:
@bors-servo r+
Thanks!

PR #134, opened 2018-05-09
Run bindgen in build.rs

asajeffrey:
This PR uses the bindgen crate to generate the Rust bindings for the `mozjs_sys` crate.

In its current form, this means requiring llvm to be installed to build `mozjs_sys`. We could check those bindings in if we want to avoid that requirement, but this has its own problems.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/134)


build.rs
```diff
@@ -66,3 +74,179 @@ fn main() {
...
+        .clang_arg("-x").clang_arg("c++")
+        .clang_arg("-std=c++11")
+        .clang_arg("-fno-sized-deallocation")
+        .clang_arg("-DRUST_BINDGEN");
+
+    if cfg!(feature = "debugmozjs") {
+        builder = builder
+            .clang_arg("-DJS_GC_ZEAL")
+            .clang_arg("-DDEBUG")
+            .clang_arg("-DJS_DEBUG");
```
jdm: These defines are missing from the jsglue compilation if debugmozjs is enabled. We should probably make some kind of common setup to avoid flags going out of sync.

asajeffrey: Oh, good point.

asajeffrey: Rather annoyingly, the flags are set in both `makefile.cargo` and `build.rs`, so we're stuck with some duplication. The thing to do eventually is move everything out of `makefile.cargo`, but I don't think that should be part of this PR.

jdm:
I'm enjoying these commit messages.

asajeffrey:
You can spot my level of annoyance with travis and appveyor increasing as the day goes on :)

nox:
FWIW, I still don't think we should run bindgen in build.rs and force everyone to have a clang setup etc. There are no particular problems with checking in the bindings, apart from generating them in the first place.

asajeffrey:
@nox yes, there's an annoying trade-off of checking generated artefacts in vs building them. If this were the only use of libclang, I'd say we should just swallow the pain and check them in,  but it looks like osmesaup may require llvm too.

asajeffrey:
Look at all that green!

asajeffrey:
Green! Lovely green!

asajeffrey:
Squashed.

PR #135, opened 2018-05-22
Version bump

asajeffrey:
This is the version bump for bindgenup, PR #134 which can land once it does.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/135)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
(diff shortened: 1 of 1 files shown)


asajeffrey:
One annoyance is that according to semver, v0.50.2 is meant to be back-compatible with v0.50.1, which this almost certainly isn't, According to semver this should be v0.51.0, but that conflicts with naming mozjs_sys versions after their SM release number.

jdm:
I don't think it's worth trying to maintain the version number to match the underlying SM version number. It's better to have a realistic semver story.

asajeffrey:
OK, v0.51.0 it is!

PR #139, opened 2018-06-18
Fix macOS/Android cross-compile.

jdm:
These changes allow my local Servo android build to complete.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/139)


jdm:
cc @paulrouget

asajeffrey:
Sigh, we have to a) specify the -include twice to get round flags being named different things on msvc and clang, and b) use an unreleased version of libloading. But it works! Let's see what homu has to say! @bors-servo r+

PR #145, opened 2018-07-26
Correct MOZTOOLS_PATH

mcduke:
The proposed `MOZTOOLS_PATH` no longer seems to be correct, as the current [MozillaBuild Package](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Windows_Prerequisites#MozillaBuild) doesn't create either `C:\mozilla-build\mozmake` or `C:\mozilla-build\yasm` directories. `mozmake.exe` and `yasm.exe` do exist in `C:\mozilla-build\bin`, though.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/145)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Under Windows:
 
 3. Set the `MOZTOOLS_PATH` environment variable to point to the tools from the Mozilla Build Package:
 ```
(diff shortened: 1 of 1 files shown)


chrmod:
In my case, python2.7 is missing as well. Adding `C:\mozilla-build\python2` to `MOZTOOLS_PATH` solved the issue

emilio:
Let's larsbors it since the failure looks like a flake and this is a doc-only update anyway.

PR #146, opened 2018-08-21
Pass -stdlib=libc++ on Apple systems

nox:
```
/Users/nox/src/mozjs/target/debug/build/mozjs_sys-0b48e14e1603d6a2/out/dist/include/mozilla/Span.h:32:10: fatal error: 'array' file not found
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/146)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -66,6 +66,13 @@ fn cc_flags() -> Vec<&'static str> {
         ]);
     }
 
+    let is_apple = target.contains("apple");
+    let is_freebsd = target.contains("freebsd");
+
+    if is_apple || is_freebsd {
+        result.push("-stdlib=libc++");
+    }
+
     result
 }
 
--- a/build.rs
+++ b/build.rs
@@ -11,7 +11,8 @@ use std::ffi::{OsStr, OsString};
 use std::process::{Command, Stdio};
 
 fn main() {
-    if cfg!(feature = "debugmozjs") && cfg!(windows) {
+    let target = env::var("TARGET").unwrap();
+    if env::var_os("CARGO_FEATURE_DEBUGMOZJS").is_some() && target.contains("windows") {
         // https://github.com/rust-lang/rust/issues/39016
         panic!("Rustc doesn't support MSVC debug runtime.");
     }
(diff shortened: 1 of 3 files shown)


nox:
I have no damn clue why Travis is suddenly complaining about 10.7 when it didn't before the smup and it's driving me mad. Cc @asajeffrey @jdm

nox:
Note that it doesn't complain about that locally either.

build.rs
```diff
@@ -64,6 +66,11 @@ fn cc_flags() -> Vec<&'static str> {
         ]);
     }
 
+    if target.contains("apple") || target.contains("freebsd") {
+        result.push("-stdlib=libc++");
+        result.push("-mmacosx-version-min=10.7");
```
nox: Yeah I've seen that we can do that but I would like to understand why this is even needed on Travis when I don't need it locally.

Btw this block is also executed when `target` contains "freebsd".

nox:
I've cleaned up the commits and added `result.push("-mmacosx-version-min=10.7");`, but I still don't understand why it is needed, especially given it just works on Servo's CI and locally.

nox:
https://github.com/llvm-mirror/clang/blob/df6c8360ca465c84d5e84bb346465848bf80e816/lib/Driver/ToolChains/Darwin.cpp#L2047-L2060

The mystery deepens.

nox:
I don't understand how the code I linked can *ever* produce "requires OS X 10.7 or later", given it can only ever fail with `where` being "iOS 5.0".

nox:
Maybe Travis is making bindgen use an older clang or something like that.

nox:
bindgen is using "clang version 5.0.0 (tags/RELEASE_500/final)", which is definitely not what bindgen is using locally on my computer, nor what Servo's CI is using.

nox:
https://github.com/KyleMayes/clang-sys/blob/master/build.rs#L113-L118

nox:
> bindgen is using "clang version 5.0.0 (tags/RELEASE_500/final)", which is definitely not what bindgen is using locally on my computer, nor what Servo's CI is using.

I'm curious to know which libclang that is, and from which directory, given it managed to build without being explicitly told to use libc++, which sounds curious to me.

nox:
So I tried to build master with various LLVM versions on macOS, here is the result of my investigation:

- library from Xcode: doesn't build;
- `llvm`: doesn't build;
- `llvm@5`: doesn't build;
- `llvm@4`: builds, lolwat.

nox:
I'm dumb I wanted to check whether installing `llvm` made the `-mmacosx-version-min` flag unnecessary but I never pushed the removal…

jdm:
FWIW my local build is either using LLVM 4 or 6, which would match the results from your investigations.

PR #147, opened 2018-08-24
Update bindgen and bump the minimum Rust version to 1.25

SimonSapin:
This lets bindgen use `#[repr(align = …)]`, which removes the need for dummy `__bindgen_align` zero-size array fields.

This field was only present with some versions of libclang (presumably), which could cause build failures like:

```rust
error[E0560]: struct `generated::root::JS::Value` has no field named `__bindgen_align`sys
  --> /home/simon/projects/mozjs/src/jsval.rs:83:2
   |
83 |     __bindgen_align: [],
   |     ^^^^^^^^^^^^^^^ `generated::root::JS::Value` does not have this field
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
-version = "0.60.1"
+version = "1.60.0"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
@@ -26,5 +26,5 @@ libc = "0.2"
 libz-sys = "1.0"
 
 [build-dependencies]
-bindgen = "0.37"
+bindgen = "0.39"
 cc = "1.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
-version = "0.60.1"
+version = "1.60.0"
```
nox: `0.61`, it's nowhere near stable and then we would need to have version `2.60` etc. Let's not keep it in sync with SM upstream given that's not doable long-term (e.g. we already need a breaking bump now).

SimonSapin:
Oops, cancelled the wrong duplicate build.

@bors-servo r=nox

SimonSapin:
```
Moving /usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask to /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

@bors-servo retry

PR #148, opened 2018-09-04
Properly pass bindgen cflags to bindgen.

emilio:
This gets me a cross build without anything else.

It's probable that we can simplify a bunch of the stuff in the build.rs file
with this patch, but I haven't tried to do that yet.

Whitespace handling is a bit broken (I hope nobody tries to build this crate in
a directory with spaces) but all the other uses of split_whitespace are equally
broken in build.rs.


jdm:
How did you even find this?

emilio:
> How did you even find this?

I've poked too much at the moz.build files for cbindgen recently :)

emilio:
This should also allow us to cross-compile OS X builds from Linux, should we want that :)

emilio:
If we want to take this fix, this should probably get cleaned up so the replace is not needed and such, and so this is unified with Gecko's usage in:

  https://searchfox.org/mozilla-central/rev/c3fef66a5b211ea8038c1c132706d02db408093a/layout/style/bindgen.toml.in#3

And then upstreamed. But this probably works as an immediate solution to unblock stuff.

paulrouget:
Added `mozjs_sys = { git = "https://github.com/servo/mozjs", branch = "mac-cross" }` to Cargo.toml.

I get:

```

error[E0063]: missing field `__bindgen_padding_0` in initializer of `mozjs_sys::root::JS::ForOfIterator`
   --> /Users/paul/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs-0.9.0/src/conversions.rs:604:28
    |
604 |         let mut iterator = ForOfIterator {
    |                            ^^^^^^^^^^^^^ missing `__bindgen_padding_0`

error: aborting due to previous error

```

asajeffrey:
My goodness, that file was well-hidden. Should we add a `println!("cargo:rerun-if-changed=js/rust/bindgen-flags");`?

emilio:
@paulrouget was it a clobber build? This touches configure stuff and I have to clobber a few times while writing the patch, the makefile.cargo condition is pretty broken.

paulrouget:
> was it a clobber build?

Yes.

Gonna retry just in case.

emilio:
If you still see it, then it'd be really unfortunate, because that means I cannot reproduce it even on the OS X loaner that I took...

emilio:
Please also ensure that the override is getting used, if you're using an outdated build that doesn't match this mozjs version it silently won't get used.

jdm:
I also have the same issue that Paul sees.

paulrouget:
adding `mozjs = { git = "https://github.com/servo/rust-mozjs", branch = "android" }` to Cargo.toml too solves the issue

paulrouget:
Is this blocked on anything?

emilio:
Not that I know of (well, needs review). It does fix my build which needed a bunch of extra headers on OSX, but probably doesn't fix the bindgen issue.

jdm:
We hit https://github.com/servo/mozjs/issues/152.

PR #150, opened 2018-09-08
Support to get PromiseIsHandled internal slot in Promise

CYBAI:
When trying to implement unhandled rejection in servo/servo#20755, I'd need to check the internal slot of Promise which is the step 4.1.1 for the spec of [notify about rejected promises](https://html.spec.whatwg.org/multipage/webappapis.html#notify-about-rejected-promises).
So, this PR will add the API to add the `GetPromiseIsHandled` method.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/150)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
-version = "0.61.0"
+version = "0.61.1"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
--- a/etc/patches/is_handled.patch
+++ b/etc/patches/is_handled.patch
@@ -1,0 +1,36 @@
+diff --git a/mozjs/js/src/jsapi.cpp b/mozjs/js/src/jsapi.cpp
+index 94ef1b3ed..06be95062 100644
+--- a/mozjs/js/src/jsapi.cpp
++++ b/mozjs/js/src/jsapi.cpp
+@@ -5162,6 +5162,14 @@ JS::GetPromiseResult(JS::HandleObject promiseObj)
+     return promise->state() == JS::PromiseState::Fulfilled ? promise->value() : promise->reason();
+ }
+ 
++JS_PUBLIC_API(bool)
++JS::GetPromiseIsHandled(JS::HandleObject promiseObj)
++{
++    PromiseObject* promise = &promiseObj->as<PromiseObject>();
++    MOZ_ASSERT(promise->state() == JS::PromiseState::Rejected);
++    return !promise->isUnhandled();
++}
++
+ JS_PUBLIC_API(JSObject*)
+ JS::GetPromiseAllocationSite(JS::HandleObject promise)
+ {
+diff --git a/mozjs/js/src/jsapi.h b/mozjs/js/src/jsapi.h
+index c7f92e273..5c5e9abc5 100644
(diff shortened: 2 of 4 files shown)


tschneidereit:
@CYBAI looks like I didn't quite add enough to JSAPI to support the WHATWG spec when I implemented rejection tracking support :(

Could you by any chance also send a patch to mozilla-central? Otherwise this will be lost with the next SpiderMonkey update. I'm happy to review there.

nox:
Can't you just use https://doc.servo.org/mozjs/jsapi/fn.GetPromiseState.html?

CYBAI:
@tschneidereit Sure! I'm willing to send a patch to mozilla-central but I haven't send one. Could you give me some related documents? Then, I can try to send it! Thanks!

@nox No, we can't. With `GetPromiseState`, we can only know if the promise is rejected or not but we'd like to know if the rejected promise is handled or not.

nox:
@CYBAI Oh I see, disregard my comment then.

tschneidereit:
> Sure! I'm willing to send a patch to mozilla-central but I haven't send one. Could you give me some related documents? Then, I can try to send it! Thanks!

Thank you, much appreciated! You'll need a checkout of mozilla-central, either using [Mercurial](https://hg.mozilla.org/mozilla-unified/) or [Git](https://github.com/mozilla/gecko-dev)—which might be preferable given that you're already using it for Servo :)

Then follow [these instructions](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Build_Documentation), ideally after doing a `./mach bootstrap` from the checkout's top-level directory.

Ideally, you'd also add to [the existing jsapi tests for Promise](https://searchfox.org/mozilla-central/source/js/src/jsapi-tests/testPromise.cpp), but I'd understand if that's asking a bit much :)

The one thing that'd be really good to have is a [Bugzilla](http://bugzilla.mozilla.org/) account, but feel free to let me know here once you have the patch, and perhaps publish it as a gist, and I'll do the rest.

CYBAI:
@tschneidereit I just move the patch into [gist](https://gist.github.com/CYBAI/82317df9d1eb1aec2f7a1ce90ddad41a) and also added two test cases for it. Please help me review it!

BTW, due to almost same as this PR, I think I can close this PR when the patch is merged into mozilla-central and also merged it into `mozjs`? Thanks!

CYBAI:
ping for review @tschneidereit

asajeffrey:
Shouldn't these changes be in the rust-mozjs repository rather than here?

CYBAI:
@asajeffrey No, we won't need to have these changes in `mozjs` repository. I think changes in [the gist](https://gist.github.com/CYBAI/82317df9d1eb1aec2f7a1ce90ddad41a) needs to be addressed in `mozilla-central` instead.

Should I create a bugzilla to send the patch then?

For `rust-mozjs` part, I think I'll need to send PR after the patch merged into `mozilla-central`.

asajeffrey:
@CYBAI the master copy of the mozjs crate is the rust-mozjs repo. I'm concerned that if we start submitting changes to m-c then the Rust code there will end up out of sync with the mozjs crate.

jdm:
@CYBAI We will need the actual changes in the patch applied to the tree in this PR as well.

CYBAI:
@jdm Ah, I see. I misunderstood what @asajeffrey said before. I'll apply the patch and update PR later. Thanks for catching that!

jdm:
@bors-servo delegate+
Please update the Cargo.toml version to 0.61.1 so we can publish this.

PR #162, opened 2018-11-27
Pass the appropriate `-stdlib` C++ compiler flag to the configure script.

pcwalton:
We were already choosing the appropriate flag when compiling `jsglue.cpp`, but
we didn't pass the flag along to `configure`, so on macOS Mojave the configure
script would detect a seemingly-broken C++ compiler and abort.

r? @jdm


jdm:
@bors-servo r+
Aha, interesting. Thanks for tracking this down!

PR #164, opened 2018-11-30
Update bindgen and bump version

Eijebong:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/164)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rust:
   - nightly
   - beta
   - stable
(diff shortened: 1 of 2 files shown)


jdm:
There is a new dependency (`which`) which uses `dyn Trait`. Rust 1.26 does not like this; let's try upgrading the version of rustc.

Eijebong:
There, updated to 1.27 which should have dyn

PR #168, opened 2019-02-26
Fix build on clang++ (macOS/freebsd)

king6cong:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/168)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -71,6 +71,7 @@ fn cc_flags() -> Vec<&'static str> {
 
     if is_apple || is_freebsd {
         result.push("-stdlib=libc++");
+        result.push("-std=c++14");
     }
 
     result


build.rs
```diff
@@ -71,6 +71,7 @@ fn cc_flags() -> Vec<&'static str> {
 
     if is_apple || is_freebsd {
         result.push("-stdlib=libc++");
+        result.push("-std=c++14");
```
jbeich: FWIW, all supported FreeBSD versions ship Clang 6 or later which defaults to C++14 as is.

nox:
@king6cong Why was this needed?

nox:
Closing this, I don't think it is an issue anymore.

PR #170, opened 2019-02-27
Disable all shared library code

jdm:
This should work around the duplicate DllMain that's appearing in https://github.com/servo/servo/pull/22856.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/170)


asajeffrey:
LGTM. Needs squashing down to 2 commits, the second just doing the version bump, then r=me.

PR #175, opened 2019-04-09
Use Xenial for TravisCI.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/175)


asajeffrey:
Waits with baited breath...

PR #179, opened 2019-04-29
Try enabling lld

jdm:
This is an experiment.

### DIFF
--- a/makefile.cargo
+++ b/makefile.cargo
@@ -122,7 +122,7 @@ else ifeq ($(findstring i686,$(TARGET)),i686)
 	# This is the correct target for MSVC builds
 	CONFIGURE_FLAGS += --target=i686-pc-mingw32 --host=i686-pc-mingw32
 endif
-	CONFIGURE_FLAGS += --without-pthreads
+	CONFIGURE_FLAGS += --without-pthreads --enable-linker=lld
 	MOZ_TOOLS=/
 
 else ifeq ($(MSYSTEM),MINGW64)


nox:
Would that improve build speeds? If so, what are we waiting for? 😁

jdm:
We were waiting for CI to actually work.

nox:
What happens if this uses lld but not the rest of our stuff btw? Isn't that a problem?

jdm:
Fun fact we use lld on Windows for everything already!

nox:
I'm a bit confused but I guess that wasn't my business anyway: I'm just wondering what's going on if we hardcode that flag in, but cargo itself doesn't use lld for its own linking step.

jdm:
We already tell cargo to use lld on Windows in Servo's .cargo/config.

nox:
> We already tell cargo to use lld on Windows in Servo's .cargo/config.

Sure, what I meant to say is: `lld` probably shouldn't be hardcoded here, we should probably pass `RUSTC_LINKER` or something like that.

> RUSTC_LINKER - The path to the linker binary that Cargo has resolved to use for the current target, if specified. The linker can be changed by editing .cargo/config; see the documentation about cargo configuration for more information.

jdm:
Windows CI failed with:
```
Traceback (most recent call last):
  File "c:/projects/mozjs/mozjs/js/src/../../configure.py", line 127, in <module>
    sys.exit(main(sys.argv))
  File "c:/projects/mozjs/mozjs/js/src/../../configure.py", line 29, in main
    sandbox.run(os.path.join(os.path.dirname(__file__), 'moz.configure'))
  File "c:\projects\mozjs\mozjs\python\mozbuild\mozbuild\configure\__init__.py", line 409, in run
    self._value_for(option)
  File "c:\projects\mozjs\mozjs\python\mozbuild\mozbuild\configure\__init__.py", line 477, in _value_for
    return self._value_for_option(obj)
  File "c:\projects\mozjs\mozjs\python\mozbuild\mozbuild\util.py", line 944, in method_call
    cache[args] = self.func(instance, *args)
  File "c:\projects\mozjs\mozjs\python\mozbuild\mozbuild\configure\__init__.py", line 542, in _value_for_option
    % option_string.split('=', 1)[0])
mozbuild.configure.options.InvalidOptionError: --enable-linker is not available in this configuration
mozmake: *** [maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', build.rs:142:5
```
No idea why.

jdm:
Caused by https://github.com/servo/mozjs/blob/b2f83932fe9d361face14efd03f2465b9262e687/mozjs/build/moz.configure/toolchain.configure#L1907-L1910.

PR #180, opened 2019-05-03
Add _LIBCPP_INLINE_VISIBILITY environment variable

asajeffrey:
This patch allows the `_LIBCPP_INLINE_VISIBILITY` variable to be set by an environment variable, rather than just basing it on the compiler and target OS. It's needed for builds for Android targets which use the usual `libc++`, such as magicleap.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/180)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
       - llvm-toolchain-xenial-5.0
     packages:
       - autoconf2.13
+      - gcc-6
       - g++-6
       - clang-5.0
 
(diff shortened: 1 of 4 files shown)


asajeffrey:
Needed to fix https://github.com/servo/servo/issues/22069

asajeffrey:
@bors-servo: retry
 * Error: retrieving gpg key timed out.

jdm:
I have retried the individual jobs that failed.

PR #185, opened 2019-06-12
Don't build NSPR at all.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/185)


asajeffrey:
No more NSPR == good! @bors-servo r+

PR #195, opened 2019-08-19
Support Rust configure when UWP targets are present

jdm:
The existing code that attempts to detect if a given target will allow compiling Rust code assumes that if multiple Rust targets target the same CPU that the triple's OS value will be unique (ie. x86_64-pc-windows-msvc vs. x86_64-pc-windows-gnu). This is no longer true with the introduction of x86_64-uwp-windows-msvc, but it's not enough to include the vendor string in the checks - the configure script is passed a different target (x86_64-windows-mingw32), and the python configuration code needs to reverse-engineer a Rust triple from it to use when compiling Rust code as part of the mozjs build. We bypass this issue by storing the target and host triples that Cargo makes available in the build script and using those value when the python configuration heuristics fail.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/195)

### DIFF
--- a/etc/patches/uwp-pc-triple.patch
+++ b/etc/patches/uwp-pc-triple.patch
@@ -1,0 +1,116 @@
+commit 81ce1668dc07897b832af74f00a3a3dbc83b337c
+Author: Josh Matthews <josh@joshmatthews.net>
+Date:   Mon Aug 19 12:43:43 2019 -0400
+
+    Support Rust configure when UWP targets are present.
+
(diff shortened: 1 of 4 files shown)


SimonSapin:
```
ERROR: Don't know how to translate x86_64-pc-mingw32 for rustc

jdm:
Yay, my changes work with nightly rustc but fail with stable and beta. More fiddling is required.

PR #200, opened 2019-09-11
Add cargo feature to build with UWP support.

jdm:
These changes remove a bunch of forbidden API uses. I've tested a local build and it starts up and can successfully JIT some rAF code in a WebGL example.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/200)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ exclude = [
 [features]
 debugmozjs = []
 profilemozjs = []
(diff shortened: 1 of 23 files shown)


asajeffrey:
Wow, that's quite a big patch! LGTM, though I'm concerned about the process of upstreaming, especially if this changes the minimum windows SDK version. @bors-servo r+

PR #205, opened 2019-10-01
Support --features vslatestinstalled

### LINKED ISSUE #204: Add opt-in way to build on VS2019
From https://github.com/servo/servo/issues/24331

Since VS2019 supports VC++ 2017, forcing VS2017 makes me install another full copy of VC++ (and MSBuild etc.) because of the single line added by #197. Could we add an opt-in way to use VS2019?

image

Figure 1 - VS Build Tools 2019 provides VC++ 2017

saschanaz:
Closes #204 

This supports opting out `--with-visual-studio-version=2017` by `--features vslatestinstalled`.

The name is from the script behavior:

https://github.com/servo/mozjs/blob/b2f83932fe9d361face14efd03f2465b9262e687/mozjs/build/moz.configure/toolchain.configure#L650-L652

(My original idea was to do `--with-visual-studio-version=2019` but the option currently only support `2017` as its value.)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/205)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ exclude = [
 debugmozjs = []
 profilemozjs = []
 uwp = []
+vslatestinstalled = []
 
 [lib]
 name = "mozjs_sys"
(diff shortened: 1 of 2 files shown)


jdm:
https://github.com/servo/rust-mozjs/blob/8e99da18185a92a5c10bc7e4bd5249bb9f9c3d84/Cargo.toml#L55 will need to be updated next as well as adding the cargo feature to forward it to this crate.

Issue #169, opened 2019-02-26
Exception: Failed to create virtualenv: /workspace/servo/target/debug/build/mozjs_sys-1352731da3144e78/out/_virtualenv


jankeromnes:
Hello!

I tried creating 8 Servo workspaces on gitpod.io, and ran `./mach build -d` in all of them (for a stress test).

Most of these builds succeeded, but one of these (identical) builds failed with:

```
   Compiling mozjs_sys v0.61.6
   Compiling script v0.0.1 (/workspace/servo/components/script)
   Compiling compositing v0.0.1 (/workspace/servo/components/compositing)
   Compiling glutin v0.19.0
error: failed to run custom build command for `mozjs_sys v0.61.6`
process didn't exit successfully: `/workspace/servo/target/debug/build/mozjs_sys-c7ce3ec219d28561/build-script-build` (exit code: 101)
--- stdout
[[ /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure -ot /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure.in ]] && touch /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure || true
[[ /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/old-configure -ot /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/old-configure.in ]] && touch /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/old-configure || true
! [[ /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure.in -ot /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure ]] && touch /home/gitpod/.cargo/registry/src/github.com-1ecc6299db9ec823/mozjs_sys-0.61.6/mozjs/js/src/configure || true
[...]

PR #207, opened 2019-10-07
Make mozjs_sys cope with RLS (fixes #206)

### LINKED ISSUE #206: mozjs cannot build under RLS
Apparently, SpiderMonkey's build system itself invokes Cargo, except that it's the RLS executable and it fails to extract Cargo's version for obvious reasons.

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/207)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,9 @@ use std::process::{Command, Stdio};
 use std::str;
 
 fn main() {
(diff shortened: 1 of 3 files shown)


nox:
Oh it did the thing from #169.

PR #208, opened 2019-10-10
Handle more env variables from the build script

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/208)


nox:
So for some reason, all the tests on the `auto` branch passed but this didn't land, I'm just going to close it because I have more interesting commits to push.

PR #209, opened 2019-10-10
Copy all the sources to OUT_DIR before building them

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/209)


asajeffrey:
I'm a bit surprised we need to roll so much code for copying sources, is there not a "recursively copy" off-the-shelf function we can use? Apart from that LGTM, you can r=me.

nox:
@bors-servo r=asajeffrey

This needed some quoting for Windows, should be ok now, hopefully.

PR #211, opened 2019-10-11
Restore outdir build metadata

nox:
This is accessed by rust-mozjs so it can build its own jsglue.cpp.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/mozjs/211)


nox:
Will you ever land things here, Homu?

PR #219, opened 2020-01-02
Support profiling feature when building for UWP.

jdm:


jdm:
The TravisCI failures will be fixed by #220.

mozjs/js/src/builtin/Profilers.cpp
```diff
@@ -24,8 +24,13 @@
 #endif
 
 #ifdef XP_WIN
-#  include <process.h>
-#  define getpid _getpid
+#  ifndef JS_ENABLE_UWP
```
asajeffrey: Nit: can we make this an `ifdef` rather than an `ifndef`?

mozjs/js/src/builtin/Profilers.cpp
```diff
@@ -24,8 +24,13 @@
...
+#    include <process.h>
+#    define getpid _getpid
+#else
```
asajeffrey: Nit: indentation.

nox:
Why was it needed to disable vtune?

mozjs/js/moz.configure
```diff
@@ -271,13 +271,13 @@ def profiling(value):
...
-@depends(profiling, target)
-def imply_vtune(value, target):
+@depends(profiling, target, js_enable_uwp)
+def imply_vtune(value, target, uwp):
     ok_cpu    = target.cpu in ['x86', 'x86_64']
     ok_kernel = target.kernel == 'WINNT' or \
                 (target.kernel == 'Linux' and target.os == 'GNU')
 
-    if value and ok_cpu and ok_kernel:
+    if value and ok_cpu and ok_kernel and not uwp:
```
nox: Why is this patch needed if we explicitly disable `vtune` when invoking `configure`? I don't get it.

jdm: Without being patched, configute complained that --enable-profiling always enabled vtune and you could not disable it.

nox: Oh so `--disable-vtune` has less weight than this implied setting? Good to know.

PR #225, opened 2020-01-10
Update SpiderMonkey

nox:


nox:
```
      0 [main] us 0 init_cheap: VirtualAlloc pointer is null, Win32 error 487
AllocationBase 0x0, BaseAddress 0xE10000, RegionSize 0xD0000, State 0x10000
C:\mozilla-build\msys\bin\sh.exe: *** Couldn't reserve space for cygwin's heap, Win32 error 0
mozmake: *** [maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', build.rs:170:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
and

```
which: python2.7: unknown command
ERROR: The libclang located at c:/PROGRA~1/LLVM/bin/libclang.dll is too old (need at least 4.0).
Please make sure to update it or point to a newer libclang using
--with-libclang-path.
mozmake: *** [maybe-configure] Error 1
thread 'main' panicked at 'assertion failed: result.success()', build.rs:170:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

For the second part, maybe I can revert the bindgen bump. For the first part, I have no clue what's going on.

jdm:
The first part has historically been https://github.com/servo/servo/issues/22212.

jdm:
Also we use LLVM 8.0 on Windows in Servo, so we could try adding that to CI - https://servo-deps.s3.amazonaws.com/msvc-deps/llvm-8.0.0.zip

asajeffrey:
Can we upgrade llvm rather than downgrade bindgen?

nox:
> Can we upgrade llvm rather than downgrade bindgen?

I was suggesting that just to get some progress here, but it was a worthless suggestion anyway. :)

nox:
SM really *really* wants to invoke rust-bindgen on its own even though it's not its business at all, I'm really tempted to just kill that.

nox:
https://searchfox.org/mozilla-central/source/js/src/make-source-package.sh#107-111

AppVeyor cannot find dllservices, and for sure, it's not included in them tarballs. I'll make a patch upstream after lunch.

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository = "https://github.com/servo/mozjs/"
-version = "0.67.1"
+version = "0.68.0"
```
atouchet: Should this be 0.72.0?

nox: No. We shouldn't be following that version because then we end up in trouble when we want to introduce additional breaking changes.

asajeffrey: FWIW, when I did the smup, I'd set NEW_VERSION to be max(OLD_VERSION+1, SPIDERMONKEY_VERSION).

nox: With this scheme you can end up with a `0.74` that is actually from Firefox 72, and that's just too confusing to me, so let's just use version numbers as they were intended to be used.

nox:
How, why, I didn't change anything, why isn't this working anymore…

build.rs
```diff
@@ -405,6 +405,10 @@ const MODULE_RAW_LINES: &'static [(&'static str, &'static str)] = &[
 /// Rerun this build script if files under mozjs/ changed, unless this returns true.
 /// Keep this in sync with .gitignore
 fn ignore(path: &Path) -> bool {
+    if path.is_dir() {
```
asajeffrey: Missing `!`?

src/jsval.rs
```diff
@@ -179,20 +179,12 @@ pub fn ObjectOrNullValue(o: *mut JSObject) -> JSVal {
...
-}
-
-#[cfg(target_pointer_width = "32")]
 #[inline(always)]
 pub fn PrivateValue(o: *const c_void) -> JSVal {
     let ptrBits = o as usize as u64;
-    assert!((ptrBits & 1) == 0);
-    BuildJSVal(ValueTag::PRIVATE, ptrBits)
+    #[cfg(target_pointer_width = "64")]
+    assert_eq!(ptrBits & 0xFFFF000000000000, 0);
```
asajeffrey: `0xFFFF000000000001` to check the bottom bit?

build.rs
nox: Nope, I specifically don't care about directories because Python puts some `pyc` files in them and so Cargo thinks they are stale.

src/jsval.rs
nox: There is no bottom bit being shifted out anymore, that bitmask comes from SM itself.

https://searchfox.org/mozilla-central/source/js/public/Value.h#189

build.rs
```diff
@@ -237,12 +237,13 @@ fn build_jsapi_bindings(build_dir: &Path) {
     config &= !bindgen::CodegenConfig::METHODS;
 
     let mut builder = bindgen::builder()
-        .rust_target(bindgen::RustTarget::Stable_1_25)
+        .rust_target(bindgen::RustTarget::Stable_1_40)
```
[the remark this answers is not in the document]
nox: Not that I know of, but both Firefox and Servo are beyond 1.40 at that point, and this is a major bump anyway.

build.rs
asajeffrey: Might be worth adding a comment to that effect?

build.rs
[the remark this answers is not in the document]
asajeffrey: I don't really have a strong opinion, 1.40 is fine by me.

build.rs
nox: Well there is a comment here now! Just kidding, I'll add one. :)

src/jsval.rs
asajeffrey: TIL

asajeffrey:
At this point, everything is niggles, you can r=me.

nox:
Added the comment about `is_dir`.

@bors-servo r=asajeffrey

PR #235, opened 2020-03-10
Don't install clang-8 package on CI.

jdm:
Something changed in the TravisCI xenial environment recently that breaks bindgen when both the base clang-7 and clang-8 packages are installed. Compilation succeeds when clang-8 isn't installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,10 @@ addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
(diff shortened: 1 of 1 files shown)


nox:
Oh neat. Initially I needed that package but that may have been due to something else, and CI is green anyway.

@bors-servo r+

nox:
It timed out during fetch, I restarted the broken job.

PR #237, opened 2020-03-23
Add RegExp bindings

teapotd:
I finally got around to finishing [form validation](https://github.com/servo/servo/pull/25447) in Servo and it looks like RegExp bindings got lost after SpiderMonkey update. This PR adds `js/RegExp.h` to `jsglue.hpp`.

### DIFF
--- a/src/jsglue.hpp
+++ b/src/jsglue.hpp
@@ -25,6 +25,7 @@
 #include "js/Promise.h"
 #include "js/PropertySpec.h"
 #include "js/Realm.h"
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

jdm:
Sigh. We break travis ci because the logs are too long if we don't rustfmt the generated bindings, but if we try to merge on days when rustfmt isn't included in the nightly then the build breaks.

PR #239, opened 2020-03-26
Disable travis builds using nightly rustc.

jdm:
We're prevented from merging any changes as long as there's no rustfmt available on nightly (https://rust-lang.github.io/rustup-components-history/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - nightly
+  #- nightly
   - beta
   - stable
 sudo: 9000


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - nightly
+  #- nightly
```
asajeffrey: Might be worth using `if [ $TRAVIS_RUST_VERSION == "stable"]` on rustfmt rather than just disabling nightly?

jdm: rustc warns about things in the generated jsapi bindings. When these don't have rustfmt applied to them, the entire contents of the file appears on one line which exceeds the travisci log limit and the build fails. We can't build without rustfmt, sadly.

asajeffrey: Oh that's annoying!

PR #240, opened 2020-03-26
Backport arm64 JIT performance improvements

jdm:
https://phabricator.services.mozilla.com/D67485
https://phabricator.services.mozilla.com/D67742


nox:
Not sure those patches will apply with `git apply` in `update.py` but shrug.

@bors-servo r+

PR #245, opened 2020-04-20
Use CARGO_MAKEFLAGS on Windows.

### LINKED ISSUE #244: No parallel builds on windows
When I look at the process tree when building mozjs_sys, I only see one clang-cl process at a time, and there are no `-j` flags in the mozmake invocations. I suspect [this code](https://github.com/servo/mozjs/blob/d4370798ee17ad16d52ff2e83c9055d19a98b26f/build.rs#L159-L165) is to blame; I don't think we should need a special windows branch any more.

jdm:
This significantly improves local build times for me when mozjs is rebuilt (such as with special features); previously we end up building all the C++ files sequentially. Fixes #244.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -156,11 +156,7 @@ fn build_jsapi(build_dir: &Path) {
     cppflags.push(env::var_os("CPPFLAGS").unwrap_or_default());
     cmd.env("CPPFLAGS", cppflags);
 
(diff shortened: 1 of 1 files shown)


nox:
The reason we weren't doing this on Windows was that there is no jobserver support on Windows, so each crate using make ends up spawning a gazillion processes each and your computer slows down a lot. If that's not an issue anymore, r=me.

PR #253, opened 2020-06-23
Don't return mozilla::Variant by value.

jdm:
This yields invalid codegen on Windows arm64 and causes crashes.


asajeffrey:
Oh goodness. Windows calling conventions, sigh. You can r=me.

PR #254, opened 2020-06-24
Add GC barriers for JSString and JSScript heap values.

jdm:
The lack of these barriers can cause memory corruption when these values are stored on the heap while a minor GC occurs.


asajeffrey:
Sigh.  GC. Sigh. @bors-servo r+

PR #255, opened 2020-06-25
Avoid invoking destructor on uninitialized StackCapture vlaues.

jdm:
All the StackCapture variants are plain old data, so we can avoid invoking the destructor when assigning via the pointer and do a memcpy instead.


jdm:
r? @asajeffrey 
This is an unfortunate followup to #253 that was exposed by the unit tests in https://github.com/servo/rust-mozjs/pull/513.

jdm:
@bors-servo r=asajeffrey
Good call!

PR #265, opened 2021-02-09
Update to mozjs 84.

jdm:


mozjs/mozglue/misc/Mutex_posix.cpp
```diff
@@ -66,11 +66,9 @@ mozilla::detail::MutexImpl::MutexImpl() {
                     "mozilla::detail::MutexImpl::MutexImpl: "
                     "pthread_mutexattr_settype failed");
 #  elif defined(POLICY_KIND)
-  if (__builtin_available(macOS 10.14, *)) {
-    TRY_CALL_PTHREADS(pthread_mutexattr_setpolicy_np(&attr, POLICY_KIND),
-                      "mozilla::detail::MutexImpl::MutexImpl: "
-                      "pthread_mutexattr_setpolicy_np failed");
-  }
+  (pthread_mutexattr_setpolicy_np(&attr, POLICY_KIND),
```
emilio: Can you elaborate on why this patch is needed?

etc/patches/0019-Unused-Windows-code.patch
```diff
@@ -0,0 +1,26 @@
...
+
+diff --git a/mozglue/misc/moz.build b/mozglue/misc/moz.build
+index d3136755f5..7ab9fe2567 100644
+--- a/mozglue/misc/moz.build
++++ b/mozglue/misc/moz.build
+@@ -25,7 +25,6 @@ EXPORTS.mozilla.glue += [
+ 
+ if CONFIG["OS_ARCH"] == "WINNT":
+     EXPORTS.mozilla += [
+-        "PreXULSkeletonUI.h",
```
emilio, line 15: Can this be upstreamed?

emilio (approved):
With answers, r=me

etc/patches/0019-Unused-Windows-code.patch
jdm: Not in its current state - Gecko uses the file on Windows, and it only causes problems in UWO builds.

mozjs/mozglue/misc/Mutex_posix.cpp
jdm: __builtin_available causes linking errors when linked statically, and I'm not sure how to work around that. The mutex policy is invalid on earlier versions of macOS, like the ones we run our CI on, but that just returns an error code but is otherwise ignored. That means we can try applying the policy and ignore it if it fails.

.travis.yml
```diff
@@ -4,7 +4,7 @@ rust:
   - beta
   - stable
 sudo: 9000
-dist: xenial
+dist: focal
```
asajeffrey, line 7: \o/

asajeffrey (approved):
Yay!

PR #258, opened 2020-12-22
Updated to support macOS 11.1

avr1:
Based on [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1682608), it looks like we can update mozjs to support up to macOS 11.1.

PR #271, opened 2021-03-03
Restore support for macOS 11.1

ksqsf:
Originally introduce by #258, but seems to be overwritten by accident by #265.


jdm:
@bors-servo r+
Thanks!

PR #277, opened 2021-04-13
Update to SpiderMonkey 88

sagudev:
In line with Firefox again 🎉 

Note to self: Now forks are having trouble running CI builds if doing work on master. We should do something like that do not trigger builds on master from @bors-servo 
something like: 
`if: "!contains(github.event.head_commit.message, 'Auto merge') && !contains(github.event.head_commit.author, 'Bors Servo')"`


jdm:
```
  cargo:warning=In file included from src/jsglue.cpp:7:
  cargo:warning=src/jsglue.hpp:42:10: fatal error: 'js/shadow/ObjectGroup.h' file not found
```

jdm:
It was removed in part 8 of https://bugzilla.mozilla.org/show_bug.cgi?id=1689413.

jdm:
It will require more changes in https://github.com/servo/rust-mozjs/blob/fe104a26147a42f6267286cd856361a126ab733b/src/rust.rs#L1189-L1193.

jdm:
Ultimately we should be able to make the same change in get_object_class as https://phabricator.services.mozilla.com/D106975.

sagudev:
Ready for merge as https://github.com/servo/servo/pull/28365 passed all test.

build.rs
```diff
@@ -7,37 +7,16 @@ extern crate cc;
...
-    "CXX",
-    "CXXFLAGS",
-    "MAKE",
-    "MOZ_TOOLS",
-    "MOZJS_FORCE_RERUN",
-    "MOZTOOLS_PATH",
-    "PYTHON",
-    "STLPORT_LIBS",
-];
+const ENV_VARS: &'static [&'static str] = &["AR", "AS", "CC", "CFLAGS", "CLANGFLAGS", "CPP", "CPPFLAGS", "CXX", "CXXFLAGS", "MAKE", "MOZ_TOOLS", "MOZJS_FORCE_RERUN", "MOZTOOLS_PATH", "PYTHON", "STLPORT_LIBS"];
```
jdm, line 17: I don't understand what happened in this commit. Can we revert it?

sagudev: I ran `cargo fmt`

sagudev: It's due to https://github.com/servo/mozjs/blob/master/rustfmt.toml#L2. Should we do something like https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#configuration-file-resolution and make sure it is only applied on jsapi files?

sagudev:
Now I moved `rustfmt.toml` to src, so only files within folder will be formatted with its options. Others (tests and build.rs) will be formatted with default settings.

build.rs
jdm: Yeah, that would be sensible.

PR #278, opened 2021-04-15
Update Bindgen

sagudev:


emilio (approved):
Would be good to remove usage of deprecated methods etc (https://github.com/servo/mozjs/blob/d68e07271794eb22e893c0bf9ef839303cc6f943/build.rs#L316 should be `blocklist_type` etc), but that can be a follow-up, this looks good anyways.

Issue #151, opened 2018-09-17
Add a CI check that patching the download with`/etc/patches` produces `mozjs`


asajeffrey:
It would be nice to have a  check that `/etc/patches` "does the right thing". This probably means adding a CI script which downloads the mozjs master, applies the patches in `/etc/patches` then does a recursive diff against the `mozjs` directory.

sagudev:
Done in #281

PR #281, opened 2021-04-26
Check patches

sagudev:
Fixing #151


sagudev:
I only do not know if homu (@bors-servo) will know that it failed.

jdm:
This looks fantastic! All we need to to tie the new job into https://github.com/servo/mozjs/blob/3d80b6e53d266e0c323547b8078bf15dd073cbf1/.github/workflows/rust.yml#L80 to ensure that homu won't merge if the job fails.

PR #282, opened 2021-05-11
Update to support latest macOS SDK

jt9897253:
Compilation with SDK version 11.3 is successful on my machine.

Without this commit, you get this error:

`ERROR: SDK version "11.3" is unsupported. Please downgrade to version 11.1.`

As the Xcode and macOS SDK version are constantly changing, I also propose to remove the check for the maximum supported SDK completely.

The maximum version check seems to be a constant source of problems when building from source. See

 - https://github.com/servo/servo/issues/28244
 - https://github.com/servo/servo/issues/25919
 - https://github.com/servo/servo/issues/26040

This should also fix https://github.com/servo/servo/issues/28244


jdm:
I agree,  removing the check does sound like a preferable choice!

jdm:
This change will need to be turned into a patch in https://github.com/servo/mozjs/tree/master/etc/patches before this can be merged.

PR #292, opened 2021-12-26
Update deprecated python imports

mvforell:
The abstract base classes in Python's `collections` module were moved to `collections.abc`. Importing them from `collections` is deprecated. With Python 3.10, importing them from `collections` does not work anymore.


jdm:
Sorry, messed this up.

PR #295, opened 2022-02-13
Upgrade SpiderMonkey to 97

CYBAI:


CYBAI:
Failed with undefined references for linkers 👀 

```
  = note: /usr/bin/ld: /home/runner/work/mozjs/mozjs/target/debug/deps/libmozjs_sys-c6f7fada473b2a03.rlib(Unified_cpp_js_src19.o): in function `JS::detail::InitWithFailureDiagnostic(bool)':
          /home/runner/work/mozjs/mozjs/mozjs/js/src/vm/Initialization.cpp:133: undefined reference to `install_rust_panic_hook'
          /usr/bin/ld: /home/runner/work/mozjs/mozjs/mozjs/js/src/vm/Initialization.cpp:134: undefined reference to `install_rust_oom_hook'
          /usr/bin/ld: /home/runner/work/mozjs/mozjs/target/debug/deps/evaluate-05785f52a9d7fc7b: hidden symbol `install_rust_oom_hook' isn't defined
          /usr/bin/ld: final link failed: bad value
          collect2: error: ld returned 1 exit status
```

jdm:
Looks like we just need to define those functions somewhere in lib.rs. They should be able to just panic.

jdm:
https://searchfox.org/mozilla-central/source/mozglue/static/rust/lib.rs#100

sagudev:
I updated SpiderMonkey to 91 a while ago, but unfortunately I didn't have time to finish servo implementation, [mozjs](https://github.com/sagudev/mozjs/tree/sm91) and [rust-mozjs](https://github.com/sagudev/rust-mozjs/tree/sm91) were mostly done, just for some inspiration if needed. You can also check my closed PRs to servo with unfinished upgrading as there was some discussion there.

CYBAI:
@sagudev thanks for sharing your branch and pointing to your PR! They are very helpful for me to see how I can fix the test failures related to bindgen. (I will also copy your update for jsglue later).

However, I've fixed most of the bindgen test failures in bac85adcb8659f0e5906d though, there's one I'm not sure how I can fix it.

They are following errors:

```
---- generated::root::JS::dbg::bindgen_test_layout_Builder_Object stdout ----
thread 'generated::root::JS::dbg::bindgen_test_layout_Builder_Object' panicked at 'assertion failed: `(left == right)`
  left: `16`,
 right: `40`: Size of: Builder_Object', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:6407:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation_1 stdout ----
thread 'generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation_1' panicked at 'assertion failed: `(left == right)`
  left: `16`,
 right: `40`: Size of template specialization: root :: JS :: dbg :: Builder_BuiltThing', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:16611:9

---- generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation stdout ----
thread 'generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation' panicked at 'assertion failed: `(left == right)`
  left: `16`,
 right: `40`: Size of template specialization: root :: JS :: dbg :: Builder_BuiltThing', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:16606:9
```

With adding `__attribute__ ((__packed__))` to the `Builder`, `Builder::BuiltThing` and `Builder::Object` classes, I will get a smaller size instead

```
---- generated::root::JS::dbg::bindgen_test_layout_Builder_Object stdout ----
thread 'generated::root::JS::dbg::bindgen_test_layout_Builder_Object' panicked at 'assertion failed: `(left == right)`
  left: `9`,
 right: `40`: Size of: Builder_Object', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:6407:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation stdout ----
thread 'generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation' panicked at 'assertion failed: `(left == right)`
  left: `9`,
 right: `40`: Size of template specialization: root :: JS :: dbg :: Builder_BuiltThing', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:16606:9

---- generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation_1 stdout ----
thread 'generated::root::__bindgen_test_layout_Builder_BuiltThing_open0_ptr_JSObject_close0_instantiation_1' panicked at 'assertion failed: `(left == right)`
  left: `9`,
 right: `40`: Size of template specialization: root :: JS :: dbg :: Builder_BuiltThing', /Users/cybai/codespace/servo/mozjs/target/debug/build/mozjs_sys-a575a7c3d708b62b/out/build/jsapi.rs:16611:9
```

@jdm or @emilio Could you please guide me how to fix the bindgen test failures here? 🙏 Thank you!

jdm:
@CYBAI My recommendation would be to skip generating Rust types for it entirely, since it's not used in any APIs we care about: https://searchfox.org/mozilla-central/search?q=symbol:T_JS%3A%3Adbg%3A%3ABuilder&redirect=false

CYBAI:
> @CYBAI My recommendation would be to skip generating Rust types for it entirely, since it's not used in any APIs we care about: https://searchfox.org/mozilla-central/search?q=symbol:T_JS%3A%3Adbg%3A%3ABuilder&redirect=false

@jdm oh, yes! if it's not used in the APIs we'd use, I think it's good to skip generating them now!

sagudev:
@CYBAI set line endings to Unix style. If you are on windows I recommend you use `git config --global core.autocrlf true` this is so called "Checkout Windows-style, commit Unix−style line endings − CRLF".

CYBAI:
> @CYBAI set line endings to Unix style. If you are on windows I recommend you use `git config --global core.autocrlf true` this is so called "Checkout Windows-style, commit Unix−style line endings".

Thanks! I'm on macOS though, let me set the config and commit again to fix the CRLF issue.

sagudev:
Hm, apparently https://github.com/servo/mozjs/blob/master/.gitattributes should be forcing LF endings. Weird.

CYBAI:
hmm, it's weird. Even I setup the `autocrlf` config with git and extracted the mozjs-97 tarball again. These files end of line are still CRLF :/

As a workaround, I just set `git config core.eol lf` and rebase master to force git to re-save those files to fix this issue 🤔

CYBAI:
hmm, weird. I've confirmed some reported files are end with LF on my local though... the integrity job still got failed :/ Let me fix another issue (the JS_GetPropertyDescriptor & friends) first. I will look into the EOL issue later.

sagudev:
Maybe mozilla changed how tar files are packed. So maybe the solution would be to change all line endings to LF before diff checking (this should be done in update.py) or simply tell diff to not differentiate line endings.

CYBAI:
@sagudev I just figured out why Integrity job failed. Because we will run the `update.py` to extract files from the tarball, the files are CRLF in the tarball so we will get the diff in the job (even they are already converted to LF on my end). I wonder maybe we can just `git add .` after the update and then run `git diff --stage --exit-code` in the integrity checking. WDYT?

CYBAI:
> Maybe mozilla changed how tar files are packed. So maybe the solution would be to change all line endings to LF before diff checking (this should be done in update.py) or simply tell diff to not differentiate line endings.

oh, diff without checking EOL sounds good to me! let me try it.

sagudev:
I am more leaning towards your option. Although I am not sure if staged commits are already converted to LF.

CYBAI:
> I am more leaning towards your option. Although I am not sure if staged commits are already converted to LF.

I tried my option locally and it seems working fine on macOS. I'm a bit confused why `--ignore-cr-at-eol` doesn't work here that [b30dfb8](https://github.com/servo/mozjs/pull/295/commits/b30dfb8df2ae3938e69b1d7fb0575f1a1aa9e43b) failed in Integrity job 🤔 

Let me update it to my option to see if it works.

sagudev:
Apparently [update.py uses `git add --all mozjs`](https://github.com/servo/mozjs/blob/master/etc/update.py#L39)

CYBAI:
hmm, with `git add`, the CRLF issue is already solved but we have some files under renamed folder which is `Click -> click` 🤔 

```
D  mozjs/third_party/python/Click/click-7.1.2.dist-info/LICENSE.rst
D  mozjs/third_party/python/Click/click-7.1.2.dist-info/METADATA
D  mozjs/third_party/python/Click/click-7.1.2.dist-info/RECORD
D  mozjs/third_party/python/Click/click-7.1.2.dist-info/WHEEL
D  mozjs/third_party/python/Click/click-7.1.2.dist-info/top_level.txt
D  mozjs/third_party/python/Click/click/__init__.py
D  mozjs/third_party/python/Click/click/_bashcomplete.py
D  mozjs/third_party/python/Click/click/_compat.py
D  mozjs/third_party/python/Click/click/_termui_impl.py
D  mozjs/third_party/python/Click/click/_textwrap.py
D  mozjs/third_party/python/Click/click/_unicodefun.py
D  mozjs/third_party/python/Click/click/_winconsole.py
D  mozjs/third_party/python/Click/click/core.py
D  mozjs/third_party/python/Click/click/decorators.py
D  mozjs/third_party/python/Click/click/exceptions.py
D  mozjs/third_party/python/Click/click/formatting.py
D  mozjs/third_party/python/Click/click/globals.py
D  mozjs/third_party/python/Click/click/parser.py
D  mozjs/third_party/python/Click/click/termui.py
D  mozjs/third_party/python/Click/click/testing.py
D  mozjs/third_party/python/Click/click/types.py
D  mozjs/third_party/python/Click/click/utils.py
?? mozjs.tar.xz
?? mozjs/third_party/python/click/
```

sagudev:
to get rid of tar use `git add --all ./mozjs`

sagudev:
It should be lowercase: 
https://searchfox.org/mozilla-central/source/third_party/python/click
https://hg.mozilla.org/mozilla-central/file/tip/third_party/python/click

sagudev:
I rechecked on my computer (linux) tar contain lowercase folder. Maybe macOS is messing with files.

CYBAI:
> I rechecked on my computer (linux) tar contain lowercase folder. Maybe macOS is messing with files.

ah, possibly. Let me try to configure `git config core.ignorecase false` to see if it can help.

CYBAI:
@sagudev Yeah, with `git rm` the capitalized `Click` folder and re-add the lower cases `click` folder. The integrity job is fixed now 🎉 Thank you so much for your help and investigation!

sagudev:
@CYBAI no problem. Can we now omit: `git config core.ignorecase true` and `git status --short`

CYBAI:
> @CYBAI no problem. Can we now omit: `git config core.ignorecase true` and `git status --short`

oh, yes, let me remove them from the job. thanks for the reminder!

src/lib.rs
```diff
@@ -27,7 +27,7 @@ mod generated {
 }
 
 fn panic_hook(info: &std::panic::PanicInfo) {
-    panic!("{}", info);
+    eprint!("Panic from mozjs: {}", info);
```
CYBAI, line 30: @jdm @sagudev While I try to update rust-mozjs, `test-panic` failed with 

```
process didn't exit successfully (signal: 4, SIGILL: illegal instruction)
```

and I realized the SIGILL is caused by panic in panic which is related to this line.

After changing the `panic` to `eprint` in this panic hook, I can pass the `test-panic` in rust-mozjs. Do you think it's fine to just `eprint` here instead of `panic` in mozjs?

sagudev: What does stderr is saying?

jdm: Yep, eprint seems reasonable. Hopefully if the panic hook doesn't do anything special the rust continues handling the original panic oncethe handler returns.

CYBAI: > What is stderr saying?

I didn't get any stderr and only get the SIGILL when the program panicked in a panic.

> Yep, eprint seems reasonable. Hopefully if the panic hook doesn't do anything special then rust continues handling the original panic once the handler returns.

Based on the `test-panic` test in rust-mozjs, I think rust will continue handling the original panic after the hook is done. because I can get the test pass after I made this line as eprint.

https://github.com/servo/rust-mozjs/blob/6dad062b9386e75165d258078a772d7ad53e9837/tests/panic.rs#L51-L61

jdm:
I've created https://github.com/servo/mozjs/pull/298 for the windows failures.

CYBAI:
Based on @sagudev's suggestion on https://github.com/servo/servo/pull/28753#issuecomment-1107825960, we might need to add a patch to mozjs to fix the build failure.

jdm:
That's already part of the patch in https://github.com/servo/mozjs/blob/master/etc/patches/0011-Introduce-enable-uwp.patch#L407 . Did something change?

sagudev:
It looks like something is missing in @CYBAI patch. https://github.com/CYBAI/mozjs/blob/smup-97/etc/patches/0011-Introduce-enable-uwp.patch#L383

CYBAI:
> That's already part of the patch in https://github.com/servo/mozjs/blob/master/etc/patches/0011-Introduce-enable-uwp.patch#L407 . Did something change?

@jdm thanks for pointing that out! we have another `RtlGenRandom` used in `mfbt/RandomNum.cpp` in this smup. I will update the patch to these lines as well 🙏 

https://github.com/CYBAI/mozjs/blob/0ce5c85726f5f9d7eb79e8f3019514a250c71eab/mozjs/mfbt/RandomNum.cpp#L97-L99

CYBAI:
> It looks like something is missing in @CYBAI patch. https://github.com/CYBAI/mozjs/blob/smup-97/etc/patches/0011-Introduce-enable-uwp.patch#L383

ah, yes, you are right. possibly I accidentally removed that line when I tried to fix different versions of these lines.

jdm:
FYI: I have started on an upgrade to SM 107, reusing a lot of work from this PR.

CYBAI:
Let's close this PR as we're going to bump to 107 at #318. 🎉

PR #298, opened 2022-03-16
Fix windows llvm installation in CI.

jdm:
The CI environment changed recently to include a version of scoop that disallows running in admin environments. Following https://github.com/ScoopInstaller/Install#for-admin we should be able to get around this.


jdm:
Subsumed by #301.

PR #303, opened 2022-06-24
Even more portable way to set SHELL to bash

fgaz:
While a NixOS system doesn't have /bin/bash but does have
/usr/bin/env, a Nix _builder_ has neither!
As pointed out by @Ms2ger in
https://github.com/servo/mozjs/pull/109#issuecomment-268814475
make will do a path search for you if only the executable
name is specified.


fgaz:
Ping

PR #315, opened 2022-11-17
Implemented Methods for PropertyKey

Redfire75369:
Resolves servo/rust-mozjs#544

There are a few inconsistent names in `jsid.rs` because `PropertyKey` and `jsid` are both names for the same thing, so I can change any of those if requested.
I have also deprecated the old functions which are in `mozjs::glue`, in favour of the new methods.

This also serves as preparation for future versions as stated in the [Migration Guide](https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/8dd70287d6214f24f7fcd1abd9d4f326ba7b4f62/docs/Migration%20Guide.md#property-keys).

### DIFF
--- a/mozjs/src/jsgc.rs
+++ b/mozjs/src/jsgc.rs
@@ -9,12 +9,12 @@ use jsapi::JSScript;
 use jsapi::JSString;
 use jsapi::JSTracer;
 use jsapi::JS;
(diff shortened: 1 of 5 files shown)


jdm (approved):
Nice! Thank you for this.

PR #318, opened 2022-11-17
Upgrade to SM 107.

jdm:
Building on https://github.com/servo/mozjs/pull/295/.

Known issues:
* [x] FromPropertyDescriptor and JS_GetPropertyDescriptor have been removed; /property_descriptor.rs needs to be updated


Redfire75369:
I think some glue could be added to adapt the functions using `mozilla::Maybe` to functions just taking pointers. 
<https://searchfox.org/mozilla-central/source/mfbt/Maybe.h#349>

sagudev:
The best option would be to glue mozilla::Maybe` to `Option`.

Redfire75369:
Rust doesn't have a stable ABI so you can't do that. You can't guarantee the bit repesentation is the same.

sagudev:
As for the CI failure, `update.py` scripts expects to be run from `.` not from `./mozjs`, but `get_mozjs*.sh` expects to be run from `./mozjs`. Maybe changing all of them to be run from `.` would be the most convenient  solution.

jdm:
Once the UWP issues from #29079 are addressed we're ready to rebase and merge!

sagudev:
Is there for every auto merge two CI runs? 
slika

jdm:
Yeah, homu force pushes twice for some reason.

PR #322, opened 2022-11-21
UWP builds in CI

sagudev:
I saw that a lot of dummy commits were made to test UWP building in https://github.com/servo/servo/pull/29079 (read: to test if mozjs actually compiles in servo), so the solution was obvious. Add UWP testing to mozjs so we can be more confident when landing changes to servo.

This PR also separates build scripts per OS host and removed testing on beta. UWP builds are run from nightly which is currently hardcoded in yaml file.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -3,3 +3,12 @@ linker = "lld-link.exe"
 
 [target.i686-pc-windows-msvc]
 linker = "lld-link.exe"
+
+[target.x86_64-uwp-windows-msvc]
+linker = "lld-link.exe"
+
+[target.aarch64-pc-windows-msvc]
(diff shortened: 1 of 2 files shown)


sagudev:
Ops! Forget to mark it as WIP. And I cannot find a button to mark it as such.

jdm:
Galaxy_brain_male_stage_3

sagudev:
Do we even have working UWP testing in servo or just building?

sagudev:
Maybe for the future: investigate if we really need both aarch64-uwp and x86_64-uwp.

jdm:
We do not have working UWP testing for servo.

jdm (approved):
Nice!

Issue #327, opened 2022-12-05
UWP builds are broken


jdm:
https://github.com/servo/mozjs/actions/runs/3620237778/jobs/6102230493

```
 error: failed to add native library D:\a\mozjs\mozjs\target\aarch64-uwp-windows-msvc\debug\build\mozjs_sys-4605c0d63069ddd7\out\build/js/src/build\js_static.lib: Unsupported archive identifier
```

This error message happens when the `object` crate that rustc uses can't read an archive: https://github.com/gimli-rs/object/blob/ed03e05213421e74ba2ccd216efe0ca0e8e06d65/src/read/archive.rs#L64 . One way to investigate this would be to save the js_static.lib file from the github task and download it for further investigation.

Redfire75369:
You can use `actions/upload-artifact` to get your hands on the file. Something like this:
```yml
    - uses: actions/upload-artifact@v3
      with:
        name: debug-build
        path: target/debug/build/mozjs_sys-*/out/build/js/build/js_static.lib
```

jdm:
I haven't made time to look into this yet. I'm going to disable the builds in the meantime.

Psychpsyo:
Does this mean #328 is no longer blocked?

sagudev:
UWP is removed.

PR #328, opened 2022-12-11
Build on latest python.

jdm:
Now that we've upgraded to a recent spidermonkey, let's see if we still need to pin to python 3.9.


jdm:
This looks like it worked but #327 is blocking it.

PR #336, opened 2023-01-07
Added More Functions to Bindgen Blacklist

### LINKED ISSUE #335: Incorrect Definition of JS::CreateError
The 9th argument, `cause` is a `u8` instead of a `Handle<mozilla::Maybe<Value>>`. We should probably remove the generation for this and provide some glue so it is still usable.

Redfire75369:
Fixes #335 and also blacklists a lot of other functions that use `mozilla::Maybe`, `mozilla::Span` or `mozilla::UniqueTwoByteChars`.


Redfire75369:
Perhaps I shouldn't have run `generate_wrappers.sh`?

jdm:
Oh, yeah, something about generate_wrappers isn't working correctly right now and keeps removing APIs that should actually be there.

jdm:
Oh, actually using the script is ok, but there are a few manual changes made to its output right now. If you look at the changes in the file, you can copy the changes from the previous versions of the declarations that are reporting compile errors.

Redfire75369:
I think I've fixed the wrappers, but I don't trust my fixes on `jsapi_wrappers.in`, so can you check those?

jdm (approved):

PR #337, opened 2023-01-08
Added Methods for ObjectOpResult

Redfire75369:
Methods are implemented according to https://searchfox.org/mozilla-central/source/js/public/Class.h#79-155 and https://searchfox.org/mozilla-central/source/js/src/jsapi.cpp#212-282.

### DIFF
--- a/mozjs/src/jsimpls.rs
+++ b/mozjs/src/jsimpls.rs
@@ -393,7 +393,7 @@ impl JS::ObjectOpResult {
     }
 
     pub fn fail(&mut self, code: JSErrNum) -> bool {
-        assert_ne!(self.code_, JS::ObjectOpResult_SpecialCodes::OkCode as usize);
+        assert_ne!(code, JS::ObjectOpResult_SpecialCodes::OkCode as usize);
         self.code_ = code as usize;
         true
     }
@@ -478,9 +478,7 @@ impl JS::ObjectOpResult {
     #[deprecated]
     #[allow(non_snake_case)]
     pub fn failNoNamedSetter(&mut self) -> bool {
-        assert_ne!(self.code_, ObjectOpResult_SpecialCodes::OkCode as usize);
-        self.code_ = JSErrNum::JSMSG_NO_NAMED_SETTER as usize;
-        true
+        self.fail_no_named_setter()
     }
 }
 
--- a/mozjs/src/jsimpls.rs
+++ b/mozjs/src/jsimpls.rs
@@ -26,6 +26,7 @@ use std::ops::Deref;
 use std::ops::DerefMut;
 use std::os::raw::c_void;
 use std::ptr;
+use jsapi::JS::{ObjectOpResult, ObjectOpResult_SpecialCodes};
 
 impl<T> Deref for JS::Handle<T> {
     type Target = T;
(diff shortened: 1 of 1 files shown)


mozjs/src/jsimpls.rs
```diff
@@ -380,20 +381,117 @@ impl<T> JS::Rooted<T> {
 }
 
 impl JS::ObjectOpResult {
+    pub fn ok(&self) -> bool {
+        assert_ne!(self.code_, JS::ObjectOpResult_SpecialCodes::Uninitialized as usize);
+        self.code_ == JS::ObjectOpResult_SpecialCodes::OkCode as usize
+    }
+
```
mrobinson, line 388: How is this different than `succeed` other than the assertion?

mozjs/src/jsimpls.rs
```diff
@@ -380,20 +381,117 @@ impl<T> JS::Rooted<T> {
...
     pub fn succeed(&mut self) -> bool {
         self.code_ = JS::ObjectOpResult_SpecialCodes::OkCode as usize;
         true
     }
 
+    pub fn fail(&mut self, code: JSErrNum) -> bool {
+        assert_ne!(self.code_, JS::ObjectOpResult_SpecialCodes::OkCode as usize);
+        self.code_ = code as usize;
+        true
+    }
```
mrobinson, line 399: Let's just expose this API, which is a bit more flexible than exposing a method for each type of failure?

mozjs/src/jsimpls.rs
```diff
@@ -380,20 +381,117 @@ impl<T> JS::Rooted<T> {
...
+    pub fn failure_code(&self) -> u32 {
+        assert!(!self.ok());
+        self.code_ as u32
+    }
+
+    #[deprecated]
     #[allow(non_snake_case)]
     pub fn failNoNamedSetter(&mut self) -> bool {
-        assert!(self.code_ != JS::ObjectOpResult_SpecialCodes::OkCode as usize);
+        assert_ne!(self.code_, ObjectOpResult_SpecialCodes::OkCode as usize);
```
mrobinson, line 481: Should it really be an error to set the value more than once?

mrobinson (changes requested):

mozjs/src/jsimpls.rs
Redfire75369: `succeed` sets the code to `OkCode`. This just checks that its not unitialised and if its `OkCode`. It follows the same API as the C++ method.

mozjs/src/jsimpls.rs
Redfire75369: I don't have any problems with this, but my initial intention with this PR was porting over the C++ methods on it, not necessarily improving/changing it.

mrobinson: Okay. I see now that this is a mirror of the C++ interface. That makes sense to me!

mozjs/src/jsimpls.rs
```diff
@@ -380,20 +381,117 @@ impl<T> JS::Rooted<T> {
...
+
     /// Set this ObjectOpResult to true and return true.
     pub fn succeed(&mut self) -> bool {
         self.code_ = JS::ObjectOpResult_SpecialCodes::OkCode as usize;
         true
     }
 
+    pub fn fail(&mut self, code: JSErrNum) -> bool {
+        assert_ne!(self.code_, JS::ObjectOpResult_SpecialCodes::OkCode as usize);
+        self.code_ = code as usize;
```
mrobinson, line 397: This does something different than the C++ API which asserts that the argument isn't JS::ObjectOpResult_SpecialCodes::OkCode. Here you are asserting that the existing code isn't that.

From https://searchfox.org/mozilla-central/source/js/public/Class.h#123:
```c++
  bool fail(uint32_t msg) {
    MOZ_ASSERT(msg != OkCode);
    code_ = msg;
    return true;
  }
```

mozjs/src/jsimpls.rs
mrobinson, line 481: There's the same issue here. This function can just be:

```rust

    #[deprecated]
    #[allow(non_snake_case)]
    pub fn failNoNamedSetter(&mut self) -> bool {
         self.fail_no_named_setter()
    }
```

mrobinson (changes requested):

mrobinson (approved):
Thank you for the updates.

sagudev:
Is this normal to be merged without https://github.com/servo/mozjs/actions/runs/5890501581 finished and with PR checks failing?

PR #339, opened 2023-01-21
Fix UWP builds

### LINKED ISSUE #334: Builds on Windows with the Beta branch are broken
It's very similar to the issue in #327. rustc is unable to read the archive file.

Action Report on spiderfire: https://github.com/Redfire75369/spiderfire/actions/runs/3860919044/jobs/6581558820
Action Report on mozjs: https://github.com/Redfire75369/mozjs/actions/runs/3861092537/jobs/6581850000

```sh
error: failed to add native library D:\a\mozjs\mozjs\target\debug\build\mozjs_sys-4062d5cd7cbc7110\out\build/js/src/build\js_static.lib: Unsupported archive identifier
```

This really seems to imply that something recent in the object crate is causing this.
It also seems likely that the reason the UWP builds were failing is because they were built with nightly.

sagudev:
Only temporary fix of: #327
I think the main reason of fail was System LLVM and Rust LLVM become out of sync.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -61,8 +61,8 @@ jobs:
       fail-fast: false
       matrix:
         features: ["--features debugmozjs", ""]
-        target: [""]
-        #target: ["", "aarch64-uwp-windows-msvc", "x86_64-uwp-windows-msvc"]
+        #target: [""]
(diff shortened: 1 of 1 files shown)


jdm:
Thanks!
@bors-servo r+

PR #326, opened 2022-11-25
Upgrade to MozillaBuild 4.0

Redfire75369:
Added Support for MozillaBuild 4.0 to build SpiderMonkey on Windows

PR #340, opened 2023-01-25
Upgrade to SM 109

wusyong:
Here's my first attempt to upgrade to spider monkey 109. But there are some issues I might need some advice, so I keep it as a draft for now. Only `mozjs-sys` has been built and tested on Linux, macOS, and Windows.

- On Windows, there are some new config checks need `MOZILLABUILD` env variable. But I couldn't figure out why it didn't call its `mach_initialize` to set it. The easiest way is to set it manually like `$env:MOZILLABUILD="C:\mozilla-build"` for example. Could we add this in CI?

- [JS Streams have been removed](https://hg.mozilla.org/mozilla-central/rev/bf1bc477f95d). I removed the bindings in sys crate as well, but not `rust-mozjs`. Servo's dom implementation is going to be affected too. I'm not sure what I should do starting from here. Could we also remove all bindings and implementation?


Redfire75369:
The issue with the `MOZILLABUILD` environment variable should be addressed when #326 is merged. That's waiting on a new release of MozillaBuild currently.

Redfire75369:
I believe this is superceded by #408 so this can be closed.

PR #341, opened 2023-01-28
Fix broken offthread compilation on Windows

jdm:
https://github.com/servo/servo/issues/29214 reported a crash when compiling scripts off-thread on Windows. By using an outptr instead of a return value, we should avoid the ABI issue.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -108,7 +108,7 @@ jobs:
         PYTHON3: "C:\\mozilla-build\\python3\\python3.exe"
         LIBCLANG_PATH: "C:\\ProgramData\\scoop\\apps\\llvm\\current\\lib"
       run: |
-        cargo test --verbose ${{ matrix.features }} --lib
+        cargo test --verbose ${{ matrix.features }}
   Integrity:
     runs-on: ubuntu-latest
     steps:
--- a/mozjs/build.rs
+++ b/mozjs/build.rs
@@ -387,6 +387,7 @@ const BLACKLIST_FUNCTIONS: &'static [&'static str] = &[
     "JS_GetPropertyDescriptor",
     "JS_GetUCPropertyDescriptor",
     "js::SetPropertyIgnoringNamedGetter",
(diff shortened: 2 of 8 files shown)


jdm:
Hmm, I was trying to reproduce the same crash from https://github.com/servo/servo/issues/29214#issuecomment-1407229240 but all the Windows tests are working fine :(

.github/workflows/rust.yml
```diff
@@ -105,31 +41,14 @@ jobs:
...
-        git add --all mozjs
-    - name: Check patch integrity
-      working-directory: ./mozjs
-      # Because we've added files in previous run, we need to
-      # check diff with `--staged`.
-      run: |
-        git diff --staged --no-ext-diff --quiet --exit-code
+        rustup install nightly-2022-11-20
+        rustup default nightly-2022-11-20
+        cargo test --verbose ${{ matrix.features }} --release --lib
```
mukilan, line 46: The mozjs binary tests are being skipped on Windows due to the `--lib` option passed to `cargo test`. We can just remove it to test both mozjs & mozjs-sys:

Here are the logs from the [last successful windows job](https://github.com/servo/mozjs/actions/runs/4033746654/jobs/6934466205)
```
    Finished release [optimized] target(s) in 9m 46s
     Running `D:\a\mozjs\mozjs\target\release\deps\mozjs-0a886cf876b6e577.exe`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `D:\a\mozjs\mozjs\target\release\deps\mozjs_sys-0bd5cea302288402.exe`

running 694 tests
```

jdm: Woah. Great catch!

jdm:
Test successfully crashed on Windows \o/

jdm:
This is ready to merge—I verified that it fixes the problem in https://github.com/servo/servo/issues/29404 and the new test fails in CI without the workaround.

mukilan (approved):

PR #344, opened 2023-02-05
Add wasm example

sagudev:
Port of [spidermonkey-embedding-examples/examples/wasm.cpp](https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/esr102/examples/wasm.cpp)

### DIFF
--- a/rust-mozjs/examples/wasm.rs
+++ b/rust-mozjs/examples/wasm.rs
@@ -1,0 +1,177 @@
+//! This example illustrates usage of WebAssembly JS API
+//! as showcased in [spidermonkey-embedding-examples/examples/wasm.cpp](https://github.com/mozilla-spidermonkey/spidermonkey-embedding-examples/blob/esr102/examples/wasm.cpp)
+//! It does no error handling and simply exits if something goes wrong.
+//!
+//! To use the WebAssembly JIT you need to create a context and a global object,
+//! and do some setup on both of these. You also need to enter a "realm"
+//! (environment within one global object) before you can execute code.
+
+use std::ptr;
+use std::ptr::null_mut;
+
+#[macro_use]
+extern crate mozjs;
+use mozjs::jsapi::*;
+use mozjs::jsval::ObjectValue;
+use mozjs::jsval::UndefinedValue;
+use mozjs::rust::jsapi_wrapped::{Construct1, JS_GetProperty, JS_SetProperty};
+use mozjs::rust::SIMPLE_GLOBAL_CLASS;
+use mozjs::rust::{JSEngine, RealmOptions, Runtime};
+
+/// hi.wat:
+/// ```
+/// (module
+///  (import "env" "bar" (func $bar (param i32) (result i32)))
+///  (func (export "foo") (result i32)
+///    i32.const 42
+///    call $bar
+///  ))
+///```
+const HI_WASM: [libc::c_uchar; 56] = [
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x0a, 0x02, 0x60, 0x01, 0x7f, 0x01, 0x7f,
+    0x60, 0x00, 0x01, 0x7f, 0x02, 0x0b, 0x01, 0x03, 0x65, 0x6e, 0x76, 0x03, 0x62, 0x61, 0x72, 0x00,
+    0x00, 0x03, 0x02, 0x01, 0x01, 0x07, 0x07, 0x01, 0x03, 0x66, 0x6f, 0x6f, 0x00, 0x01, 0x0a, 0x08,
+    0x01, 0x06, 0x00, 0x41, 0x2a, 0x10, 0x00, 0x0b,
+];
+
+unsafe extern "C" fn bar(_cx: *mut JSContext, argc: u32, vp: *mut Value) -> bool {
+    let args = CallArgs::from_vp(vp, argc);
+    args.rval().set(args.get(0).get());
+    true
+}
+
+fn run(rt: Runtime) {
+    let options = RealmOptions::default();
+    rooted!(in(rt.cx()) let global = unsafe {
+        JS_NewGlobalObject(rt.cx(), &SIMPLE_GLOBAL_CLASS, ptr::null_mut(),
+                           OnNewGlobalHookOption::FireOnNewGlobalHook,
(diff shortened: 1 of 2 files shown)


rust-mozjs/examples/wasm.rs
```diff
@@ -0,0 +1,177 @@
...
+        rooted!(in(rt.cx()) let mut module = null_mut::<JSObject>());
+        {
+            let array_buffer = JS::NewArrayBufferWithUserOwnedContents(
+                rt.cx(),
+                HI_WASM.len(),
+                HI_WASM.as_ptr() as _,
+            );
+            assert!(!array_buffer.is_null());
+
+            let args = HandleValueArray::from_rooted_slice(&[ObjectValue(array_buffer)]);
```
jdm, line 88: This slice is not actually rooted. I'll try to provide useful suggestions later when I'm not on my phone.

rust-mozjs/examples/wasm.rs
```diff
@@ -0,0 +1,177 @@
...
+            rooted!(in(rt.cx()) let mut imports = JS_NewPlainObject(rt.cx()));
+            assert!(!imports.is_null());
+            assert!(JS_SetProperty(
+                rt.cx(),
+                imports.handle(),
+                b"env\0".as_ptr() as *const libc::c_char,
+                env_import.handle()
+            ));
+
+            let args = HandleValueArray::from_rooted_slice(&[
```
jdm, line 124: Same comment about this slice. from_rooted_slice is so easy to misuse ;_;

rust-mozjs/examples/wasm.rs
```diff
@@ -0,0 +1,177 @@
...
+        rooted!(in(rt.cx()) let mut exports_obj = exports.to_object());
+        rooted!(in(rt.cx()) let mut foo = UndefinedValue());
+        assert!(JS_GetProperty(
+            rt.cx(),
+            exports_obj.handle(),
+            b"foo\0".as_ptr() as *const libc::c_char,
+            &mut foo.handle_mut()
+        ));
+
+        // call foo ang get its result
```
jdm, line 156:
```suggestion
        // call foo and get its result
```

rust-mozjs/examples/wasm.rs
sagudev: Would something like this be ok:
```rust
rooted!(in(rt.cx()) let val = ObjectValue(array_buffer));
let args = HandleValueArray::from_rooted_slice(&[val.get()]);
```

jdm: Hmm, almost, but not quite. There's an awkward but correct way of doing that:
```rust
rooted!(in(rt.cx()) let val = ObjectValue(array_buffer));
let args = HandleValueArray { length_: 1, elements_: &*val };
```
In Servo we have the RootedVec type which could probably be used to build a cleaner binding here, but that's not part of this crate.

rust-mozjs/examples/wasm.rs
```diff
@@ -0,0 +1,179 @@
...
+            assert!(!imports.is_null());
+            assert!(JS_SetProperty(
+                rt.cx(),
+                imports.handle(),
+                b"env\0".as_ptr() as *const libc::c_char,
+                env_import.handle()
+            ));
+
+            let args_slice = [ObjectValue(module.get()), ObjectValue(imports.get())];
+            let args = HandleValueArray::from_rooted_slice(&args_slice);
```
sagudev, line 129: I am not 100% sure about this one, I did it like it is done in [servo/components/script/dom/paintworkletglobalscope.rs](https://github.com/servo/servo/blob/b1d90f871dd4ddddbd6f4fa66648265674338e29/components/script/dom/paintworkletglobalscope.rs#L344).

jdm: That code needs to be fixed as well ;_;

sagudev: Not a Rust unsafe experienced but wrapping `[ObjectValue(module.get()), ObjectValue(imports.get())]` in `ManuallyDrop` should be enough?

jdm: The safety risk here is that the garbage collector does not know about the slice that contains those JS values, so if they are moved for any reason like a GC occurs during the next spidermonkey API call), the JS values will not be updated with the new pointers. This means the handles passed as arguments would be invalid and could cause a crash.

jdm: I filed https://github.com/servo/mozjs/issues/346 about the missing pieces to make this safe.

sagudev: Hm, but we do have `CustomTrace` and `auto_root!` so 
```rust
auto_root!(in(rt.cx()) let vec = vec![ObjectValue(module.get()), ObjectValue(imports.get())]);
```
should do it, right?

mrobinson:
Oh, perhaps I reviewed this too hastily.  Maybe @jdm would look at this one as well since there were some comments before.

rust-mozjs/examples/wasm.rs
```diff
@@ -0,0 +1,180 @@
...
+            assert!(JS_SetProperty(
+                rt.cx(),
+                imports.handle(),
+                b"env\0".as_ptr() as *const c_char,
+                env_import.handle()
+            ));
+
+            rooted!(in(rt.cx()) let mut arg1 = ObjectValue(module.get()));
+            rooted!(in(rt.cx()) let mut arg2 = ObjectValue(imports.get()));
+            let args = HandleValueArray::from_rooted_slice(&[arg1.get(), arg2.get()]);
```
jdm, line 130: Let's use rooted_vec here; it will be safe to call `HandleValueArray::from_rooted_slice` using the slice from inside the vector.

jdm (approved):

rust-mozjs/examples/wasm.rs
sagudev: To put this in rooted_vec we would need to wrap JSval in Heap (beacuse it needs to implement Traceable).

sagudev: maybe we should port RootedValueArray

jdm (approved):
Nice.

sagudev:
Rebased and one new commit added.

jdm (approved):

PR #347, opened 2023-02-21
README: add instructions for building on NixOS

delan:
This patch makes it easier for NixOS users to work on mozjs by adding a shell.nix that should be exhaustive (it compiles ok under nix-shell --pure), and some advice to the README about Visual Studio Code and rust-analyzer.


mukilan (approved):
nice! I was able to build mozjs on my new machine by just entering nix-shell, installing rust nightly and running cargo test!

@bors-servo r+

PR #349, opened 2023-03-23
Use sccache & install llvm based on rustc version

### LINKED ISSUE #302: Windows CI breaks when the environment's LLVM doesn't match the version installed by scoop
Keeping up with this requires changing the explicit version installed by scoop (eg. `llvm@13.0.1`) whenever the Windows virtual environment updates with a new LLVM version: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md

sagudev:
fixes #302

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,7 @@ on:
 
 env:
   #CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: 1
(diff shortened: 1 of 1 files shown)


jdm:
Can we do the LLVM changes separate from sccache? I'm not convinced that this will solve #302.

sagudev:
Closed in favor of #350

PR #351, opened 2023-03-25
Use runners llvm on windows

### LINKED ISSUE #302: Windows CI breaks when the environment's LLVM doesn't match the version installed by scoop

sagudev:
If installed LLVM and runners LLVM need to be in sync, the solution is clear: use runners LLVM, that way we also remove installing scoop/LLVM and speeding up builds.

Why hasn't it been done before? IIRC due to missing `libclang` or `llvm-config` or some other tool from LLVM that was not included in runners LLVM.

Should fix #302

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,6 +65,14 @@ jobs:
         features: ["--features debugmozjs", ""]
         #target: [""]
         target: ["", "aarch64-uwp-windows-msvc", "x86_64-uwp-windows-msvc"]
+    env:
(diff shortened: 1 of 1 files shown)


jdm:
Huh. We always needed to install LLVM because libclang was not included in the preinstalled MSVC component, but apparently that is no longer the case.

PR #356, opened 2023-06-02
Upgrade to MozBuild 4.0

mrobinson:


.github/workflows/rust.yml
```diff
@@ -84,22 +84,49 @@ jobs:
         components: rust-src
     - name: Install deps
       run: |
-        Start-BitsTransfer -Source https://ftp.mozilla.org/pub/mozilla/libraries/win32/MozillaBuildSetup-3.4.exe -Destination ./MozillaBuildSetup.exe
+        Start-BitsTransfer -Source https://ftp.mozilla.org/pub/mozilla/libraries/win32/MozillaBuildSetup-4.0.2.exe -Destination ./MozillaBuildSetup.exe
         .\MozillaBuildSetup.exe /S | Out-Null
-        echo 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Tools\LLVM\bin' | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        iwr -useb get.scoop.sh -outfile 'install.ps1'
```
sagudev, line 89: This should not be here as it was removed in https://github.com/servo/mozjs/commit/16c72d786f623f6844c315c7fe8a9fb19d713f13

.github/workflows/rust.yml
```diff
@@ -84,22 +84,49 @@ jobs:
...
+      if: (!contains(matrix.target, 'uwp'))
+    - uses: dtolnay/rust-toolchain@master
+      if: contains(matrix.target, 'uwp')
+      with:
+        toolchain: "nightly-2023-02-02"
+        components: rust-src
     - name: Build uwp
       if: contains(matrix.target, 'uwp')
-      shell: cmd
+      env:
```
sagudev, line 111: This was also refactored in https://github.com/servo/mozjs/commit/16c72d786f623f6844c315c7fe8a9fb19d713f13

sagudev (commented):
BONUS: As a workaround until https://bugzilla.mozilla.org/show_bug.cgi?id=1803269 lands in tagged release I created my [own tagged release](https://github.com/sagudev/mozilla-build/releases/tag/MOZILLABUILD_4_0_42_RELEASE).

Redfire75369:
When I tried this without switching the makefile to tabs, I'm pretty sure it had failed. Something about mozmake didn't like the formatting in the makefile.

.github/workflows/rust.yml
```diff
@@ -132,4 +159,4 @@ jobs:
         if: success()
       - name: Mark the job as unsuccessful
         run: exit 1
-        if: "!success()"
```
sagudev, line 135: No workflow runs:
```
Invalid workflow file: .github/workflows/rust.yml#L1
The workflow is not valid. .github/workflows/rust.yml: Unexpected tag '!success()'
```

.github/workflows/rust.yml
mrobinson: Thanks. I've fixed this.

.github/workflows/rust.yml
mrobinson: I've also updated this as well.

mrobinson:
@sagudev Thanks for the review. Does this look good to you now? It seems like the CI is passing.

mrobinson:
> When I tried this without switching the makefile from tabs, I'm pretty sure it had failed. Something about mozmake didn't like the formatting in the makefile.

Huh...that's interesting. I wonder if it was some other issue. Makefiles traditionally use tabs.

sagudev (approved):
LGTM

Redfire75369:
This most likely shouldn't be merged yet. Either the next version of Mozbuild with the patch should be released or we should be using @sagudev's release.

mrobinson:
> This most likely shouldn't be merged yet. Either the next version of Mozbuild with the patch should be released or we should be using @sagudev's release.
> 
> It's a pretty simple change at least, but I'm not sure which to go with.

Would it make sense to merge this change and then switch to @sagudev's packaged version of MozBuild in a followup PR?

Redfire75369:
I'm mostly against merging this directly since the `start-shell.bat` and having it in the repository itself is a very hacky workaround. You should probably ask @jdm for his opinion though.

mrobinson:
Closing this one in favor of a switch to Servo's mozillatools zip bundle.

PR #357, opened 2023-06-03
Fix rustbindgen tags in documentation

leftmostcat:
Trailing solidus in non-void tags is [not valid HTML5](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-non-void-html-element-start-tag-with-trailing-solidus) and is specified to be treated as a start tag. This breaks the layout of the [generated Rust docs](https://doc.servo.org/mozjs/jsapi/index.html) and makes them very hard to read and use.

Replacing these with standard `div` tags fixes layout:
image

### DIFF
--- a/mozjs/src/jsglue.hpp
+++ b/mozjs/src/jsglue.hpp
@@ -176,7 +176,7 @@ class JSJitMethodCallArgsReplacement {
 #endif
 };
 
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #358, opened 2023-06-04
Upgrade mach

mrobinson:
This change upgrades the version of mach in the repository to match the one from the latest commit in the Gecko repository. This should allow mozjs to build with newer versions of Python. Notably, this change does not upgrade any other Mozilla python packages such as mozbuild so that they maintain compatibility with the build files of SpiderMonkey. This is meant as a stopgap to fix compilation on newer versions of Python until a proper SpiderMonkey upgrade, which will require a significant amount of work downstream in Servo.


sagudev:
Maybe porting only patch for py3.11 support would be easier.

mrobinson:
I've fixed the build and I think this is ready to go. It might be a smaller change to only include the fixes for Python 3.11, but to be honest I'm not sure exactly which change fixed things. Additionally, I think this will help us avoid problems in the future, because we pick up all the improvements that have been made to mach.

sagudev:
Well if it works, it works.

Maybe it would be useful to update mach in servo too (preferably to the same version).

jdm:
@bors-servo r+
Thanks for doing this!

PR #360, opened 2023-07-10
Enable the GitHub merge queue

mrobinson:


.github/workflows/rust.yml
```diff
@@ -2,8 +2,10 @@ name: Rust
 
 on: 
   push:
-    branches-ignore: "master"
+    branches: [master]
```
mukilan, line 5: Are we good with the loss of functionality here?  Previously the CI was being run on all branches except master, which could be useful to allow testing WIP branches that don't yet have a pull request. With this change we will only run these on master, which is already protected.

mrobinson: It's a good question. I think most folks uploading changes to origin will want to open a PR for them, which will run the workflows. This prevents the workflow from immediately running twice when pushing a branch and opening a PR.

mukilan (approved):

PR #362, opened 2023-08-10
Fix mismatched ProxyTraps signature

sagudev:
Will fix https://github.com/servo/servo/issues/30043


mukilan (approved):
Thank you for the fix!

PR #366, opened 2023-08-16
Wrap `JS_FORGET_STRING_LINEARNESS`

Redfire75369:


Redfire75369:
I'm not 100% sure if that `reinterpret_cast` is sound, but it's similar to what I found [here](https://searchfox.org/mozilla-central/source/js/public/Id.h#203), for `PropertyKey::toLinearString`.

mrobinson:
Would it make sense to expose `JSLinearString` as well?

Redfire75369:
`JSLinearString` is already exposed, it's just that it can't be used that much because most functions only take in `JSString`.
Refer to <https://doc.servo.org/servo/index.html?search=JSLinearString> under "In Parameters" and "In Return Types".
I guess it would probably be a good idea to expose `JS_FORGET_STRING_LINEARNESS` (maybe even just that, without this wrapper).

mrobinson:
> `JSLinearString` is already exposed, it's just that it can't be used that much because most functions only take in `JSString`. Refer to https://doc.servo.org/servo/index.html?search=JSLinearString under "In Parameters" and "In Return Types". I guess it would probably be a good idea to expose `JS_FORGET_STRING_LINEARNESS` (maybe even just that, without this wrapper).

That makes sense to me. Thanks!

rust-mozjs/src/jsglue.cpp
```diff
@@ -1018,6 +1018,14 @@ void EncodeStringToUTF8(JSContext* cx, JS::HandleString str,
...
 }
 
+JSString* JS_ForgetStringLinearness(JSLinearString* str) {
+  return JS_FORGET_STRING_LINEARNESS(str);
+}
+
+JSString* EnsureLinearString(JSContext* cx, JSString* str) {
+  return JS_FORGET_STRING_LINEARNESS(JS_EnsureLinearString(cx, str));
+}
+
```
mrobinson, line 1028: Can this API return a JSLinearString so that it works the same as the wrapped version?

Redfire75369: In that case, this API isn't needed at all. `JS_EnsureLinearString` can already be called from rust.

mrobinson (approved):
Looks great. Let me know when this is ready to merge.

Redfire75369:
It's ready to merge.

Redfire75369:
It's ready to be merged.

PR #367, opened 2023-08-17
Add support for Android NDK r21

mrobinson:
This is the essentially the NDK that the version of SpiderMonkey
supports. This makes a few changes to the SpiderMonkey source as well:

1. Pulls in a change from a more recent version so that the correct
   version of the clock functions are used.
2. Disables all attempts by SpiderMonkey to set the sysroot, Android
   platform includes, and the STL. Modern NDKs don't require these
   flags and they often just break mysteriously if not used perfectly.
   SpiderMonkey upstream is starting to use them less, but is still
   pretty far in the past. This makes everything about the Android build
   simpler.

In addition, some compilation errors for 32-bit machines are fixed.


mozjs/etc/patches/0026-Add-support-for-NDKr21.patch
```diff
@@ -0,0 +1,90 @@
...
+index d4967864d..30eba4ff5 100644
+--- a/build/autoconf/android.m4
++++ b/build/autoconf/android.m4
+@@ -8,12 +8,14 @@ AC_DEFUN([MOZ_ANDROID_NDK],
+ case "$target" in
+ *-android*|*-linuxandroid*)
+     dnl $android_* will be set for us by Python configure.
+-    directory_include_args="-isystem $android_system -isystem $android_sysroot/usr/include"
++    dnl
++    dnl See the mozjs note android-nkd.configure for why these lines are commented out.
```
jdm, line 11: nit: .ndk, not .nkd

jdm (approved):
Nice!

Issue #300, opened 2022-03-27
CI broken with mozilla-build 4.0


jdm:
```
     Running `D:\a\mozjs\mozjs\target\debug\build\mozjs_sys-429806b452f72395\build-script-build`
[mozjs_sys 0.68.2] cargo:outdir=D:\a\mozjs\mozjs\target\debug\build\mozjs_sys-c3e02811bd22dd0e\out\build
[mozjs_sys 0.68.2] gmake: *** No targets specified and no makefile found.  Stop.
[mozjs_sys 0.68.2] thread 'main' panicked at 'Failed to run `make`: Error { kind: NotFound, message: "program not found" }', build.rs:177:10
[mozjs_sys 0.68.2] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I believe this is because msys64 isn't placed in the path by default in the new environment (see note in https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md). The easiest way to deal with this is probably via https://github.com/msys2/setup-msys2

jdm:
This is impacting the windows-2019 environment as well for reasons that I don't understand.

jdm:
This is actually caused by the release of Mozilla-Build 4.0 last week: https://groups.google.com/a/mozilla.org/g/dev-platform/c/0GHhML7tY6c . If we update, we'll need to address that.

sagudev:
integrity check is also broken, either taskcluster changed their api (unlikly) or previous build is out of cache.

jdm:
That's a very old issue; the taskcluster API returns an error for something that it shouldn't, and the taskcluster team didn't look into when I raised it on Matrix.

sagudev:
It's actually treeherder. My bad.

sagudev:
Probably out of cache, their retention is 6 or 9 months I guess.

PR #368, opened 2023-08-17
Use moztools instead of mozilla-build

sagudev:
This makes mozjs on windows more standalone and allows to remove presetting variables in servo.

Servo's `./mach bootstrap` can still download moztools and mozjs will use downloaded moztools if they are right version.

WIP because I still need to make companion PR for servo.

Fixes https://github.com/servo/mozjs/issues/300
Supersedes https://github.com/servo/mozjs/pull/326 & https://github.com/servo/mozjs/pull/356


mrobinson:
This is a nice change!

Some suggestions:
1. Can you have `build.rs` download anything yet, but focus on whether or not to do that in a followup? Just download the Servo's Mozilla tools in the GitHub action. That way we have two clean changes:
   - Switch to Servo's packaged tools.
   - Decide if we want  build.rs to take care of that.
2. The documentation should be updated to reflect this change.
3. Let's move the changes to the actions unrelated to this change to a new PR (not running actions in forks).

sagudev:
1. This could be problematic as if we switch to moztools but do not implement downloading step the in this middle commit users needs to manually set `MOZ_TOOLS_PATH` or manually download moztools and extract it to `target/dependencies`
2. Will do!
3. #369

mrobinson:
> 1. This could be problematic as if we switch to moztools but do not implement downloading step the in this middle commit users needs to manually set `MOZ_TOOLS_PATH` or manually download moztools and extract it to `target/dependencies`, so I think it would be better to have it in same change.

This is the status quo with MozillaBuild so I think it's fine to keep this behavior for now. I think we should hold off on the download step...because we really haven't decided what to do with this dependency. For instance, maybe we could convert it into a crate and avoid having to do a manual download at all.

sagudev:
> > 1. This could be problematic as if we switch to moztools but do not implement downloading step the in this middle commit users needs to manually set `MOZ_TOOLS_PATH` or manually download moztools and extract it to `target/dependencies`, so I think it would be better to have it in same change.
> 
> This is the status quo with MozillaBuild so I think it's fine to keep this behavior for now. I think we should hold off on the download step...because we really haven't decided what to do with this dependency. For instance, maybe we could convert it into a crate and avoid having to do a manual download at all.

Using mozilla-build was easier as you can just simply install mozilla-build using installer and open mozilla-build shell (like you would do if you wanted to develop for firefox) and run commands from there. Manually setting `MOZILLA_BUILD` was just a trick to simulate mozilla-build env in github actions where you cannot open mozilla-build shell to run commands.

mozjs/build.rs
```diff
@@ -131,24 +138,102 @@ fn cc_flags() -> Vec<&'static str> {
...
+            return target_dir.to_path_buf();
+        }
+        dir = target_dir;
+    }
+    panic!("$OUT_DIR is not in target")
+}
+
+fn find_moztools() -> Option<PathBuf> {
+    let cargo_target_dir = cargo_target_dir();
+    let deps_dir = cargo_target_dir.join("depedencies");
```
Redfire75369, line 155: Small spelling mistake here, should be `"dependencies"`.

mozjs/build.rs
```diff
@@ -131,24 +138,102 @@ fn cc_flags() -> Vec<&'static str> {
...
+
+            // curl is on Win10+
+            let mut curl = Command::new("curl");
+
+            let moztools_zip = &deps_dir.join("moztools.zip");
+            let result = curl
+                .arg("-SL")
+                .arg(MOZ_TOOLS_DL)
+                .arg("--create-dirs")
+                .args(&["-o", &moztools_zip.display().to_string()])
```
Redfire75369, line 212: This can be replaced with `&[OsStr::new("-o"), moztools_zip.as_os_str()]`, which both looks nicer and doesn't have unnecessary allocations.

mozjs/build.rs
```diff
@@ -131,24 +138,102 @@ fn cc_flags() -> Vec<&'static str> {
...
+                .status()
+                .unwrap();
+            assert!(result.success());
+
+            // extract
+            let file = fs::File::open(moztools_zip).unwrap();
+            let mut archive = zip::ZipArchive::new(file).unwrap();
+            archive.extract(deps_dir).unwrap();
+
+            find_moztools().unwrap()
```
Redfire75369, line 222: The `.unwrap()`'s in these lines should be replaced with `.expect()` with an appropriate error message so errors can be diagnosed more easily.

mozjs/build.rs
```diff
@@ -131,24 +138,99 @@ fn cc_flags() -> Vec<&'static str> {
...
+
+    if moztools_path.exists() {
+        Some(moztools_path)
+    } else {
+        None
+    }
+
+    // to support many moztools versions
+    /*
+    // maybe moztools are already in targed/depedencies/moztools-*
```
Redfire75369, line 166: Minor spelling mistake, should be `target/dependencies/moztools-*`.

mozjs/build.rs
```diff
@@ -131,24 +138,99 @@ fn cc_flags() -> Vec<&'static str> {
...
-    // If MOZILLA_BUILD is specified, process that environment variable to put
-    // the build tools on the path and properly set the AUTOCONF environment
-    // variable.
-    if let Some(mozilla_build_path) = env::var_os("MOZILLA_BUILD") {
+    #[cfg(windows)]
+    {
+        let moztools = if let Some(moztools) = env::var_os("MOZ_TOOLS_PATH") {
+            PathBuf::from(moztools)
+        } else if let Some(moztools) = find_moztools() {
+            // moztools already in targed/depedencies/moztools-*
```
Redfire75369, line 197: Minor spelling mistake, should be `target/dependencies/moztools-*`.

mozjs/build.rs
```diff
@@ -131,24 +138,99 @@ fn cc_flags() -> Vec<&'static str> {
...
+            moztools
+        } else {
+            // download moztools
+            let cargo_target_dir = cargo_target_dir();
+            let deps_dir = cargo_target_dir.join("depedencies");
```
Redfire75369, line 202: Minor spelling mistake, should be `"dependencies"`.

Redfire75369 (changes requested):

mozjs/build.rs
```diff
@@ -131,24 +138,103 @@ fn cc_flags() -> Vec<&'static str> {
...
-    // If MOZILLA_BUILD is specified, process that environment variable to put
-    // the build tools on the path and properly set the AUTOCONF environment
-    // variable.
-    if let Some(mozilla_build_path) = env::var_os("MOZILLA_BUILD") {
+    #[cfg(windows)]
+    {
+        let moztools = if let Some(moztools) = env::var_os("MOZ_TOOLS_PATH") {
+            PathBuf::from(moztools)
+        } else if let Some(moztools) = find_moztools() {
+            // moztools already in targed/dependencies/moztools-*
```
Redfire75369, line 196: Minor spelling mistake, should be `target/dependencies/moztools-*`.

mozjs/build.rs
```diff
@@ -131,24 +138,103 @@ fn cc_flags() -> Vec<&'static str> {
...
+        None
+    }
+
+    // to support many moztools versions
+    /*
+    if let Ok(dirs) = std::fs::read_dir(deps_dir) {
+        dirs.into_iter().find_map(|p| {
+            if let Ok(pp) = p {
+                if let Some(name) = pp.path().file_name() {
+                    if name.to_string_lossy().contains("moztools") {
```
Redfire75369, line 170: This part is fine if you're still planning to implement it in this PR, but should probably be removed otherwise.

Redfire75369 (changes requested):

Redfire75369 (approved):

mrobinson (changes requested):
I think we should not download  moztools in `build.rs`. In addition the comments above, this introduces a hidden dependency on cURL to build this crate. Let's just do the minimal change here and require that it be unzipped ahead of time and an environment variable set.

sagudev:
> In addition the comments above, this introduces a hidden dependency on cURL to build this crate.
Curl is installed on every Win10+ system by default otherwise we could just depend on some rust crate for downloading.

I already have separation per commits, but now I would like to test with servo first and then do changes here.

Redfire75369:
I think this should accept either `MOZ_TOOLS_PATH` or `MOZILLA_BUILD` aa the env variable for backwards compatibility. Mainly outside of servo, this could be pretty annoying as a change.

sagudev:
> I think this should accept either MOZ_TOOLS_PATH or MOZILLA_BUILD aa the env variable for backwards compatibility. Mainly outside of servo, this could be pretty annoying as a change.

You are right, while we are still on moztools-3.2 we can still support MozillaBuild 3.2+

sagudev:
@mrobinson any progress?

mozjs/build.rs
```diff
@@ -39,6 +38,9 @@ const EXTRA_FILES: &'static [&'static str] = &[
     "src/jsglue.cpp",
 ];
 
+/// Which version of moztools we expect
+const MOZ_TOOLS_VER: &str = "3.2";
```
mrobinson, line 42:
```suggestion
const MOZ_TOOLS_VERSION: &str = "3.2";
```

mozjs/build.rs
```diff
@@ -25,9 +25,8 @@ const ENV_VARS: &'static [&'static str] = &[
     "CXX",
     "CXXFLAGS",
     "MAKE",
-    "MOZ_TOOLS",
+    "MOZ_TOOLS_PATH",
```
mrobinson, line 28:
```suggestion
    "MOZTOOLS_PATH",
```
This would make it consistent with the name `moztools`.

.github/workflows/rust.yml
```diff
@@ -84,9 +82,9 @@ jobs:
         components: rust-src
     - name: Install deps
       run: |
-        Start-BitsTransfer -Source https://ftp.mozilla.org/pub/mozilla/libraries/win32/MozillaBuildSetup-3.4.exe -Destination ./MozillaBuildSetup.exe
-        .\MozillaBuildSetup.exe /S | Out-Null
         echo 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Tools\LLVM\bin' | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        curl -SL "https://github.com/servo/servo-build-deps/releases/download/msvc-deps/moztools-3.2.zip" --create-dirs -o target/dependencies/moztools.zip
+        cd target/dependencies && unzip -qo moztools.zip -d . && mkdir moztools && mv moztools-3.2 moztools/3.2
```
mrobinson, line 87: I know this is a bit of a small thing, but I think we should stick to the environment variable here so that we are testing the method that most people will be using on the CI.

mozjs/build.rs
```diff
@@ -131,24 +133,57 @@ fn cc_flags() -> Vec<&'static str> {
...
+        let moztools = if let Some(moztools) = env::var_os("MOZ_TOOLS_PATH") {
+            PathBuf::from(moztools)
+        } else if let Some(moztools) = find_moztools() {
+            // moztools already in target/dependencies/moztools-*
+            moztools
+        } else if let Some(moz_build) = env::var_os("MOZILLA_BUILD") {
+            // For now we also support mozilla build
+            PathBuf::from(moz_build)
+        } else {
+            panic!("MozTools not found!");
```
mrobinson, line 175: It would be nice here to add a link to the build instructions.

mrobinson (changes requested):
Sorry for the delay in reviewing. This looks good to me. The one larger change I think we should have apart from the small things below is that we should update the README to explain downloading, unzipping, and setting the environment variable for moztools. Thanks!

.github/workflows/rust.yml
sagudev: You are right, this one will already be tested in servo.

mozjs/build.rs
```diff
@@ -39,7 +39,7 @@ const EXTRA_FILES: &'static [&'static str] = &[
 ];
 
 /// Which version of moztools we expect
-const MOZ_TOOLS_VER: &str = "3.2";
+const MOZ_TOOLS_VERSION: &str = "3.2";
```
mrobinson, line 42: Apologies for missing this, but:

```suggestion
const MOZTOOLS_VERSION: &str = "3.2";
```

mozjs/build.rs
```diff
@@ -172,7 +172,11 @@ fn build_jsapi(build_dir: &Path) {
             // For now we also support mozilla build
             PathBuf::from(moz_build)
         } else {
-            panic!("MozTools not found!");
+            panic!(
+                "MozTools or MozillaBuild not found!\n
+Follow instructions on: https://github.com/servo/mozjs?tab=readme-ov-file#windows
+            "
+            );
```
mrobinson, line 179:
```suggestion
            panic!(
                "MozTools or MozillaBuild not found!\n \
                 Follow instructions on: https://github.com/servo/mozjs#windows "
            );
```

I think that the slash escapes leading whitespace.

mrobinson (changes requested):
Sorry, a couple other tiny things.

mozjs/src/jsimpls.rs
```diff
@@ -484,9 +484,7 @@ impl JS::ObjectOpResult {
 
 impl Default for ObjectOpResult {
     fn default() -> ObjectOpResult {
-        ObjectOpResult {
-            code_: ObjectOpResult_SpecialCodes::Uninitialized as usize,
-        }
+        ObjectOpResult { code_: ObjectOpResult_SpecialCodes::Uninitialized as usize }
```
mrobinson, line 487: Were these changes added by mistake?

sagudev: I run `cargo fmt` and I guess this was not formatted properly before.

sagudev:
Failures are not my fault: https://github.com/actions/runner-images/issues/8125

mrobinson (approved):

PR #369, opened 2023-08-17
Run CI only on master in upstream but everywhere in forks

sagudev:
Makes my life easier when testing changes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,7 +2,6 @@ name: Rust
 
 on: 
   push:
-    branches: [master]
   pull_request:
   merge_group:
     types: [checks_requested]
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -18,6 +17,8 @@ env:
 
 jobs:
   mac:
+    # Run CI only on master in upstream
+    if: ${{ !(github.repository == 'servo/servo' && github.event_name == 'push') || github.ref_name == 'master' }}
```
mrobinson, line 21: I'm guessing `servo/servo` here should be `servo/mozjs`?

sagudev: Yes

mrobinson (approved):

PR #370, opened 2023-08-17
Clean up the Cargo makefile

mrobinson:
1. Consistently use tabs.
2. Move all configure options next to each other
3. Use else if where possible.
4. Remove unused arguments to configure


mrobinson:
Thanks for the review.

PR #373, opened 2023-08-17
Properly mark failed CI as failing

mrobinson:
There was a syntax error in `if` condition for the `Result` CI job.
This meant that when a dependent job failed, the CI pipeline did not.
This should fix branch protection for the merge queue.


sagudev:
I had this marked in my vscode using https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions extension but I thought that it was extension mistake. I guess it was not.

jdm (approved):

mrobinson:
Hrm. I think might be a bit more going on here:

1. The job was being skipped even with the syntax fix, but I'm attempting to see if adding `if: ${{ always() }}` fixes this.
2. Running the workflow for pushes to the non-master branch means that the `Result` job runs and passes always. `Result` is the gate for the merge queue. I'm afraid that this might allow accidentally merging changes that fail the CI. I've tried to skip it in that case, I'm still a bit worried.

@sagudev I think it might be good to modify this workflow a bit so that it only runs when pushing to a special branch like "try."

sagudev:
For second point moving mark successful down and adding additional if logic to unsuccessful should be enough.

Also some old servo workflows have used explicit checking that needs jobs where successful not skipped or cancelled this might come in handy.

I've also seen something similar in one of the rust-lang repos that was using mq, but I cannot remember which one.

mrobinson:
Okay. I believe that the key here was to not just use `success()` and `failure()` which I think only check steps before the current one, but also the result of all dependent jobs for the `Result` job. It's quite likely that workflows in other repositories need this fix as well.

sagudev:
Something is still wrong. Merges are still not waiting.

mrobinson:
Yep! I've posted a revert of #369 as I think that's the culprit here. Let's revert for now and maybe we can land a similar change later that doesn't bust the merge queue.

sagudev:
It's wired as mq simply does not wait for result job: https://github.com/servo/mozjs/actions/runs/5893040023/job/15984166916

PR #380, opened 2023-09-04
use windows-2019 until build with windows-2022 is fixed

mukilan:
Windows builds are failing recently, potentially caused by the windows-2022 images being [updated to use LLVM 16](https://github.com/servo/mozjs/pull/368#issuecomment-1700838483). 

I haven't analyzed the failures in detail yet. To unblock open PRs, this change makes CI use windows-2019 which is still on LLVM 15. I have validated that the builds pass on my [fork](https://github.com/mukilan/mozjs/actions/runs/6072763112).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -62,7 +62,7 @@ jobs:
         cargo build --verbose ${{ matrix.features }}
         cargo test --verbose ${{ matrix.features }}
   windows:
-    runs-on: windows-latest
+    runs-on: windows-2019
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


sagudev:
Actually it is just other way around. windows-2022 does not ship llvm 16 but llvm 15 which is not compatible with latest msvc stdlib. Example error: https://github.com/servo/mozjs/actions/runs/6058661158/job/16440993596?pr=379#step:7:616

mukilan:
Ah, thanks for the context! I'll keep this PR open in case GitHub doesn't deploy the new Window 2022 images with LLVM 16  soon.

PR #387, opened 2023-09-06
Reformatted JSVal Implementation

redfire75369:
Aligned JSVal Implementation with C++ and reduced 32-bit vs 64-bit specific implementations

### DIFF
--- a/mozjs-sys/src/jsval.rs
+++ b/mozjs-sys/src/jsval.rs
@@ -98,6 +98,18 @@ fn AssertGCPointerAlignment(bits: u64) {
 #[inline(always)]
 fn AssertGCPointerAlignment(bits: u64) {}
 
+#[cfg(target_pointer_width = "64")]
+#[inline(always)]
+fn IsValidUserModePointer(bits: u64) -> bool {
+    bits & 0xFFFF_0000_0000_0000 == 0
+}
+
+#[cfg(target_pointer_width = "32")]
+#[inline(always)]
+fn IsValidUserModePointer(_: u64) -> bool {
+    true
+}
+
 #[inline(always)]
 pub fn Int32Value(i: i32) -> JSVal {
     BuildJSVal(ValueTag::INT32, i as u32 as u64)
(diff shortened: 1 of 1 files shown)


mozjs/src/jsval.rs
```diff
@@ -14,117 +14,126 @@ use crate::jsapi::JS::Symbol;
...
+
+        #[cfg(target_pointer_width = "32")]
+        #[repr(u32)]
+        #[allow(dead_code)]
+        enum $name {
+            $(
+                $variant = JSVAL_TAG_CLEAR | ($value_type as u32),
+            )*
+        }
+    };
```
mrobinson, line 54: Sorry, drive by review:

I feel like the addition of a macro here obfuscates this code a bit and the definition of `ValueTag` seems identical between the two pointer sizes apart from the clear value.. Would it be possible to instead conditionally define a value to bitwise or with the ValueTag members?

```rust
#[cfg(target_pointer_width = "32")]
const JSVAL_TAG_CLEAR: u32 = 0xFFFFFF80;
#[cfg(target_pointer_width = "64")]
const JSVAL_TAG_CLEAR: u32 = 0x1FFF0;

#[repr(u32)]
#[allow(dead_code)]
enum ValueTag {
   INT32 = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_INT32 as u32),
    UNDEFINED = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_UNDEFINED as u32),
    STRING = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_STRING as u32),
    SYMBOL = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_SYMBOL as u32),
    BIGINT = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_BIGINT as u32),
    BOOLEAN = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_BOOLEAN as u32),
    MAGIC = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_MAGIC as u32),
    NULL = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_NULL as u32),
    OBJECT = JSVAL_TAG_CLEAR | (JSValueType::JSVAL_TYPE_OBJECT as u32),
}
```

Redfire75369: I see what you mean. I don't think the 64-bit one would count as a "clear tag" though.
Looking at [Value.h#207](https://searchfox.org/mozilla-central/source/js/public/Value.h#207) and [Value.h#230](https://searchfox.org/mozilla-central/source/js/public/Value.h#230), they are specifically separate.
I was trying to make the implementation align with the C++ one, so I left them separate. They also seem to serve different purposes in terms of the bit representation.

mrobinson: Hrm. If they should be represented separately, I don't think sharing the macro is the best way. Maybe instead come up with a good name for the thing you bitwise or with and add a comment explaining what is going on here.

PR #394, opened 2023-09-10
More const in jsglue

sagudev:
Previously manual rust bindings have const ptr all over the place, now that bindings are auto generated those const need to be present in C++ code.


mrobinson:
Do these changes fix compilation warnings or is something else going on here?

sagudev:
It makes servo switching easier as the final product are more similar rust bindings as before.

The only changes in servo needed were https://github.com/servo/servo/pull/30326/commits/bef06c2670a3e2acc4bc94bdc33551981278509f, but if weren't for this PR there would also be some const -> mut changing in servo needed. See errors from this job run: https://github.com/servo/servo/actions/runs/6131256294/job/16641083229

jdm (approved):

PR #399, opened 2023-09-11
Fixed build.rs Warnings

Redfire75369:
The warning was caused by `make` not being accessed after the initial assignment on Windows and immediately being reassigned.

### DIFF
--- a/mozjs/build.rs
+++ b/mozjs/build.rs
@@ -76,6 +76,7 @@ fn main() {
     }
 }
 
+#[cfg(unix)]
 fn find_make() -> OsString {
     if let Some(make) = env::var_os("MAKE") {
         make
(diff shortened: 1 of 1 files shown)


sagudev:
could you also fix warnings on linux: https://github.com/servo/mozjs/actions/runs/6148189813/job/16681363526#step:6:322

mozjs/build.rs
```diff
@@ -183,6 +184,11 @@ fn build_jsapi(build_dir: &Path) {
         make = OsStr::new("mozmake").to_os_string();
     }
 
+    #[cfg(unix)]
```
sagudev, line 187: `#[cfg(not(windows))]` might make more sense

sagudev (approved):

PR #400, opened 2023-09-11
Make generate_wrappers work on dirty builds

### LINKED ISSUE #398: generate_wrappers.sh only works on Clean Builds
Due to various reasons, it is possible for the `target` directory to contain multiple versions of the build directories for `mozjs_sys` and `mozjs`. This causes the following lines in [`generate_wrappers.sh`](https://github.com/servo/mozjs/blob/master/rust-mozjs/src/generate_wrappers.sh) to fail as multiple versions of  `jsapi.rs`/`gluebindings.rs` are found.

<https://github.com/servo/mozjs/blob/6b299dec6b9421a903ef3ce752b957b3473c0b1c/rust-mozjs/src/generate_wrappers.sh#L31-L34>

Although running `cargo clean` every time is not the worst, it would probably be better for `generate_wrappers.sh` to detect the most recently modified one. I'm trying to work on a solution but my shell skills are insufficient.

sagudev:
Fixes #398

### DIFF
--- a/mozjs/src/generate_wrappers.sh
+++ b/mozjs/src/generate_wrappers.sh
@@ -27,12 +27,15 @@ grep_heur() {
   grep -v 'MutableHandleObjectVector' # GetDebuggeeGlobals has it
 }
 
-# clone file and reformat
-cp target/debug/build/mozjs_sys-*/out/build/jsapi.rs target/jsapi.rs
-rustfmt target/jsapi.rs --config max_width=1000
-cp target/debug/build/mozjs-*/out/gluebindings.rs target/glue.rs
-rustfmt target/glue.rs --config max_width=1000
+find_fmt_parse() {
+  # clone file and reformat
+  # this $(find) only gets last modified file
+  cp $(find target -name "$1" -printf "%T@ %p\n" | sort -n | tail -n 1 | tr ' ' '\n' | tail -n 1) "target/wrap_$1"
+  rustfmt "target/wrap_$1" --config max_width=1000
+  
+  # parse file
+  grep_heur "target/wrap_$1" | $gsed 's/\(.*\)/wrap!('"$2"': \1);/g'  > "mozjs/src/$2_wrappers.in"
+}
 
-# parse file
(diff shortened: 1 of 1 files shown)


sagudev:
@Redfire75369 can you check if it works for you.

Redfire75369:
Yes, it works. Seems fine to me

mrobinson:
Is it possible to refactor the call to find and the parsing of its output into a sh function in order to avoid so much duplication?

sagudev:
I wish I had this idea before.

mozjs/src/generate_wrappers.sh
```diff
@@ -27,12 +27,15 @@ grep_heur() {
   grep -v 'MutableHandleObjectVector' # GetDebuggeeGlobals has it
 }
 
-# clone file and reformat
-cp target/debug/build/mozjs_sys-*/out/build/jsapi.rs target/jsapi.rs
-rustfmt target/jsapi.rs --config max_width=1000
-cp target/debug/build/mozjs-*/out/gluebindings.rs target/glue.rs
-rustfmt target/glue.rs --config max_width=1000
+find_fmt_parse() {
```
mrobinson, line 30: How about `find_latest_version_of_file_parse_and_format`?

mrobinson (approved):
Looks good with one naming nit.

mozjs/src/generate_wrappers.sh
sagudev: That is so loooooong. Maybe just `last_file_parse`?

mrobinson: Bits are very cheap. It is better to be verbose and obvious, rather than short and obscure. Many people will read this code in the future and it took me a bit to understand it.

sagudev: used long name and added usage comment and some more comments inside function

mrobinson (approved):

mrobinson:
Thank you!

PR #402, opened 2023-09-12
Add cross testing for linux arm64 & armhf

### LINKED ISSUE #401: Add cross compilation testing
The goal would be to test most interesting platforms that are also supported by SM:
- aarch64-linux
- armhf-linux
- arm64-windows (might need SM bump)

sagudev:
Partially resolves #401

This PR uses our [public docker images](https://github.com/orgs/servo/packages) for cross compiling. Minimal cross configuration is provided for easier usage of those docker images.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -110,8 +110,8 @@ jobs:
         run: |
           ./android-build cargo build --target="armv7-linux-androideabi"
 
-  cross:
-    name: Cross ${{ matrix.target }}
+  linux-cross-compile:
+    name: linux (${{ matrix.target }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -153,7 +153,7 @@ jobs:
   build_result:
     name: Result
     runs-on: ubuntu-latest
-    needs: ["mac", "linux", "windows", "android", "cross"]
+    needs: ["android", "linux", "linux-cross-compile", "mac", "windows"]
     if: ${{ always() }}
     steps:
       - name: Mark the job as successful
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -110,6 +110,27 @@ jobs:
         run: |
           ./android-build cargo build --target="armv7-linux-androideabi"
 
+  cross:
+    name: Cross ${{ matrix.target }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - aarch64-unknown-linux-gnu
+          - armv7-unknown-linux-gnueabihf
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+      - name: Install cross
+        uses: baptiste0928/cargo-install@v2
+        with:
+          crate: cross
+      - run: cross test --target ${{ matrix.target }}
+
   # The integrity check is currently broken. See issue #345.
   #
   # Integrity:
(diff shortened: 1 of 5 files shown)


Cross.toml
```diff
@@ -0,0 +1,27 @@
+[target.aarch64-unknown-linux-gnu]
+image = "ghcr.io/sagudev/cross/aarch64-unknown-linux-gnu:23"
```
mrobinson, line 2: How are these images created? Is there some way this can be managed via the servo organization?

mozjs/makefile.cargo
```diff
@@ -54,13 +54,21 @@ ifneq ($(HOST),$(TARGET))
 			$(NULL)
 	endif
 
+	ifeq (armv7-unknown-linux-gnueabihf,$(TARGET))
+		# Reset TARGET variable because armv7 target name used by Rust is not
+		# the same as the target name needed for the CXX toolchain.
+		TARGET = arm-linux-gnueabihf
+		# fpu according to cross-rs
```
mrobinson, line 61: Is there a link to the documentation or code that means that we need this fpu setting?

Cross.toml
sagudev: Well I needed to to few updates to cross images most notably we need ubuntu 22.04 for libstdc++12 default cross images have only ubuntu 20.04.

Here are my sources: https://github.com/sagudev/cross/tree/patch-1-ci. This should only be temporary solution until cross get newer images. I will try to get my changes in upstream, but this might take a long time.

mozjs/makefile.cargo
sagudev: There is [this ](https://github.com/cross-rs/cross/blob/main/docker/Dockerfile.armv7-unknown-linux-gnueabihf#L50)but I am not really sure it's needed. I will do try commit and see.

Cross.toml
mrobinson: Thanks for the information. Isn't it possible to cross-compile without `cross` until the time that it has what we need?

mozjs/makefile.cargo
sagudev: Looks like it's not needed really.

Cross.toml
sagudev: Cross compiling is hard, especially on linux. One nice thing that cross offers is also running test in qemu, that is something that is even harder to set up manually.

sagudev: > Is there some way this can be managed via the servo organization?

Yes, they can be managed via servo org. When we liberate servo from mach, we could also generate docker images for cross compiling servo with clang and other deps preinstalled, then we could use same images here too.

Cross.toml
```diff
@@ -0,0 +1,5 @@
+[target.aarch64-unknown-linux-gnu]
+image = "ghcr.io/sagudev/cross-aarch64-unknown-linux-gnu:docker-images"
```
mukilan, line 2: Can these be switched to servo images as well?

sagudev: Good catch!

.github/workflows/rust.yml
```diff
@@ -110,6 +110,24 @@ jobs:
         run: |
           ./android-build cargo build --target="armv7-linux-androideabi"
 
+  cross:
+    name: Cross ${{ matrix.target }}
```
mrobinson, line 114: Just for the sake of clarity and consistency with the other job names:

```suggestion
  linux-cross-compile:
    name: linux (${{ matrix.target }})
```

.github/workflows/rust.yml
```diff
@@ -135,7 +153,7 @@ jobs:
   build_result:
     name: Result
     runs-on: ubuntu-latest
-    needs: ["mac", "linux", "windows", "android"]
+    needs: ["mac", "linux", "windows", "android", "cross"]
```
mrobinson, line 156:
```suggestion
    needs: ["android", "linux", "linux-cross-compile", "mac", "windows"]
```

mrobinson (approved):
Nice.

PR #407, opened 2023-09-15
Added Windows on Arm Cross Compile to CI

Redfire75369:


sagudev (changes requested):
Given that it's so similar to windows-x64 could you merge them.

sagudev (approved):

mukilan:
Thank you!

PR #409, opened 2023-09-18
Move `rust-mozjs`'s build script to `mozjs-sys`

wusyong:
Move all build scripts into `mozjs-sys` crate which is more conventional when binding FFI.
This can also make possible future changes like linking binary build instead of building everything from source easier.

I added another commit to rename original mozjs-sys's `jsglue.cpp` to `jsapi.cpp`, and kept `rust-mozjs`'s `jsglue`.
It's just clearer IMHO. Let me know if there are better names.


wusyong:
If it's higher level wrappers, it should be implemented by Rust using sys crate directly.
Having another build script to compile and bindgen C++ code makes the structure a bit more complicated.
I'm okay if this module should stay in rust-mozjs, but I think it's better to rewrite in Rust.

wusyong:
@sagudev Is it possible to push this PR again? I've rebased the PR to latest main branch.
It will still be the same files but they will be inside the same `OUT_DIR` at least.
This should make #439 easier.

sagudev:
@jdm what do you think about this?

Redfire75369:
I'm not in favour of either option, but I will note that the current split between `mozjs-sys`'s `jsglue.cpp` and `mozjs`'s `jsglue.cpp` has no consistent rhyme or reason, some cleanup would be required there regardless of the outcome.

jdm (approved):
This seems like a reasonable choice!

sagudev (approved):

wusyong:
Thanks for merging!

PR #413, opened 2023-09-20
Match folder names to crate names

### LINKED ISSUE #411: Match folder names to crate names.
Currently we have `rust-mozjs` with `mozjs` crate and `mozjs` folder with `mozjs-sys` crate.

sagudev:
`mozjs` -> `mozjs-sys`
`rust-mozjs` -> `mozjs`

Fixes #411


mrobinson:
I'm really happy to see this change. Looks like CI is failing.

sagudev:
Ah, I will investigate tomorrow.

sagudev:
Apparently `git mv` exists for a reason. Anyway it would be nice to merge this without squash again, as there is one big commit that does moving and one commit that fixes all scripts/docs about folder layout assumptions.

mrobinson:
> Apparently `git mv` exists for a reason. Anyway it would be nice to merge this without squash again, as there is one big commit that does moving and one commit that fixes all scripts/docs about folder layout assumptions.

One benefit of squash here is that it allows git bisect to work.

sagudev:
> > Apparently `git mv` exists for a reason. Anyway it would be nice to merge this without squash again, as there is one big commit that does moving and one commit that fixes all scripts/docs about folder layout assumptions.
> 
> One benefit of squash here is that it allows git bisect to work.

That's why rustc is doing bisects on merge commits. I wonder if `git bisect` can do it too.

mrobinson:
> That's why rustc is doing bisects on merge commits. I wonder if `git bisect` can do it too.

Since the commit history of mozjs is now linear, there are no more merge commits even without squashing.

mrobinson:
Perhaps a clearer way of stating this is mozjs-sys is "SpiderMonkey and low-level Rust bindings to its C++ API" Does that read a bit more clearly?

mrobinson (approved):

PR #414, opened 2023-09-21
Fixed Windows Compilation with LLVM 17

### LINKED ISSUE #410: Build Failure on Windows with LLVM 17.0.1
```bat
  error: unsupported option '-G' for target 'x86_64-pc-windows-msvc'
  # ...
  thread 'main' panicked at 'Unable to generate bindings to jsglue: ClangDiagnostic("error: unsupported option '-G' for target 'x86_64-pc-windows-msvc' ...)
```
Not sure why this happens, but updating from 16.0.6 caused this issue.

Full Log with `-v`: <https://paste.gg/p/redfire/0ec149108e884db08c0bd9ca052645a7>

Redfire75369:
Fixes #410

### DIFF
--- a/mozjs/build.rs
+++ b/mozjs/build.rs
@@ -84,7 +84,6 @@ fn main() {
         builder = builder.clang_args([
             &format!("-FI{}", confdefs_path.to_string_lossy()),
             "-DWIN32",
-            "-GR-",
             "-std=c++17",
         ])
     } else {
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -356,7 +356,7 @@ fn build_jsapi_bindings(build_dir: &Path) {
     );
 
     println!(
-        "Generting bindings {:?} {}.",
+        "Generating bindings {:?} {}.",
         builder.command_line_flags(),
         bindgen::clang_version().full
     );


mozjs/build.rs
```diff
@@ -82,9 +82,9 @@ fn main() {
 
     if msvc {
         builder = builder.clang_args([
+            "-fms-compatibility",
             &format!("-FI{}", confdefs_path.to_string_lossy()),
             "-DWIN32",
-            "-GR-",
```
sagudev, line 87: Is removing this still needed?

Redfire75369: I believe so. Either way, `-fms-compatibility` effectively includes that by default.

sagudev: I would keep it to make things more explicit.

Redfire75369: Since we're compiling for Windows with MSVC, I'd say just having `-fms-compatibility` is fine. If anything further breaks, that's a clang thing.

Moreover, this lines up with `mozjs-sys/build.rs`, which also uses this flag.

sagudev: I think (not sure) that those options were historically sourced from SM, and there is -GR- specified explicitly, that's why I would keep it. @jdm might know this more specifically.

sagudev (approved):

sagudev:
Last commit wasn't merged but I think `-fms-compatibility` is also good so I wouldn't revert it.

PR #416, opened 2023-09-23
Added BigInt64 and BigUint64 Typed Array Support

Redfire75369:


sagudev (changes requested):
Could you also add them to this test: https://github.com/servo/mozjs/blob/master/mozjs/tests/typedarray.rs

Redfire75369:
The current version of that test already does not test for all the different variants of TypedArrays. For example, none of the `Int*Array` and `Float*Array` are tested. As such, if tests are added for `Big{Int,Uint}64Array`, they should probably be added for all of them.

jdm (approved):

PR #418, opened 2023-09-23
Update SM filter

### LINKED ISSUE #419: Windows Path Too Long
In the CI for spiderfire, it is said that the path is too long (260 chars).
```
path too long: 'C:/Users/runneradmin/.cargo/git/checkouts/mozjs-fa11ffc7d4f1cc2d/d4a3ea8/mozjs-sys/mozjs/js/src/tests/test262/language/eval-code/direct/arrow-fn-no-pre-existing-arguments-bindings-are-present-arrow-func-declare-arguments-assign-incl-def-param-arrow-arguments.js'; class=Filesystem (30)
  ```
https://github.com/Redfire75369/spiderfire/actions/runs/6284551616/job/17066027695#step:11:32

### PR DESCRIPTION
**Pull request by sagudev**
Updated filter files adds some new filters (for stuff we do not need/use; some new rust stuff that is also not really needed) and some rules are updated. We could probably remove even more, but this ones are the most easiest.

Reduces size of `mozjs-sys/mozjs` from 350MB to 210MB.

Fixes #419 

There should be no more Cargo.toml so we should be able to publish on crates.io I think.

NOTE TO REVIEWS: Last commit is automated, so you do not need to review it.

### CODE REVIEW

File: mozjs-sys/etc/filters.txt [hunk 1]
```diff
@@ -1,3 +1,7 @@
+# we do not need testing infra
+- /js/src/fuzz-tests
+- /js/src/jit-test
```
sagudev, line 3: There is actually more testing stuff that we could remove, but this have some rust in them, so they are more problematic and not really needed. There are also jsapi-test that are kept for now.

jdm (approved):
If it still builds then I guess we're good!

PR #423, opened 2023-10-21
Added Glue for GetRegExpFlags

Redfire75369:
- Updated Comment on MSVC Non-POD Return Values

### DIFF
--- a/mozjs/src/jsglue.cpp
+++ b/mozjs/src/jsglue.cpp
@@ -23,6 +23,7 @@
 #include "js/Principals.h"
 #include "js/Promise.h"
 #include "js/Proxy.h"
+#include "js/RegExp.h"
 #include "js/ScalarType.h"
 #include "js/Stream.h"
 #include "js/StructuredClone.h"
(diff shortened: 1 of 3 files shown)


mozjs-sys/src/jsimpls.rs
```diff
@@ -194,7 +194,7 @@ impl JS::AutoGCRooter {
         // mutable before setting it.
         // https://github.com/rust-lang-nursery/rust-bindgen/issues/511
         #[allow(non_snake_case)]
-        let autoGCRooters = &*self.stackTop as *const _ as *mut _;
```
sagudev, line 197: I wonder if we still need hoops?

sagudev:
And wrappers need regeneration.

mozjs-sys/src/jsimpls.rs
sagudev: We do not. This is now in #425

mozjs/src/jsglue.cpp
```diff
@@ -985,13 +986,12 @@ bool WriteBytesToJSStructuredCloneData(const uint8_t* src, size_t len,
...
 
-// MSVC uses a different calling conventions for functions
+// MSVC uses a different calling convention for functions
 // that return non-POD values. Unfortunately, this includes anything
-// with a constructor, such as JS::Value, so we can't call these
-// from Rust. These wrapper functions are only here to
-// ensure the calling convention is right.
-// https://docs.microsoft.com/en-us/cpp/build/return-values-cpp
-// https://mozilla.logbot.info/jsapi/20180622#c14918658
+// with a constructor, such as JS::Value as JS::RegExpFlags, so we
```
sagudev, line 991: Did you meant `as JS::Value or JS::RegExpFlags`?

sagudev (approved):

PR #424, opened 2023-10-29
Add simple traceable impl for std::time::Duration

sagudev:


sagudev:
Ah we need to land fix for rust 1.73:
```
error: assigning to `&T` is undefined behavior, consider using an `UnsafeCell`
   --> mozjs-sys/src/jsimpls.rs:198:9
    |
197 |         let autoGCRooters = &*self.stackTop as *const _ as *mut _;
    |                             ------------------------------------- casting happend here
198 |         *autoGCRooters = self.down;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(invalid_reference_casting)]` on by default
```

mrobinson (approved):

PR #425, opened 2023-10-31
Remove casting hoops (fixes compilation on rust 1.73)

sagudev:
This hoops are not needed anymore because `AutoGCRooter` is now properly handled by bindgen:
```rust
pub struct AutoGCRooter {
    pub down: *mut AutoGCRooter,
    pub stackTop: *mut *mut AutoGCRooter,
    pub kind_: AutoGCRooterKind,
}
```

and code is now more similar to upstream: https://searchfox.org/mozilla-central/source/js/public/RootingAPI.h#1082 and to JS::Rooted method:
https://github.com/servo/mozjs/blob/ec63ca39efb346c414e0f55ef26e63c0b17ee69a/mozjs-sys/src/jsimpls.rs#L418-L421

### DIFF
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -189,13 +189,7 @@ impl JS::AutoGCRooter {
 
     pub unsafe fn remove_from_root_stack(&mut self) {
         assert!(*self.stackTop == self);
-        // This hoop-jumping is needed because bindgen gives stackTop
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

mrobinson:
Looks like this has failed, so I propose we land your two fixes together.

sagudev:
Landed within #426

PR #426, opened 2023-11-01
Two changes which fix the build

sagudev:
This change includes two fixes:

1. [Remove casting hoops (fixes compilation on rust 1.73)](https://github.com/servo/mozjs/pull/425)
2. Port Bug1857492 (Build fails on Python 3.12: No module named 'six.moves') patch

These two changes have been merged together as they are both necessary for the CI to pass.

Fixes https://github.com/servo/servo/pull/30658#issuecomment-1787849458


mrobinson (approved):
Thanks for the fix!

mrobinson (approved):
Thanks for the fix!

mrobinson (approved):
Thanks for the fix!

mozjs-sys/src/jsimpls.rs
```diff
@@ -189,13 +189,7 @@ impl JS::AutoGCRooter {
 
     pub unsafe fn remove_from_root_stack(&mut self) {
         assert!(*self.stackTop == self);
-        // This hoop-jumping is needed because bindgen gives stackTop
-        // the type *const *mut AutoGCRooter, so we need to make it
-        // mutable before setting it.
-        // https://github.com/rust-lang-nursery/rust-bindgen/issues/511
-        #[allow(non_snake_case)]
-        let autoGCRooters = &*self.stackTop as *const _ as *mut _;
-        *autoGCRooters = self.down;
```
mrobinson, line 198: Looks like this commit was inadvertently included though.

mrobinson (changes requested):

mozjs-sys/src/jsimpls.rs
sagudev: Without it CI fails, but I would like to try to send #425 to mq first (and hope we get old mac runners)

mrobinson (approved):

mrobinson:
I am going to go ahead and try to land this. Sorry for the rush, but I'd like to unstick the Servo merge queue.

mrobinson:
I've updated the PR message to explain that this has two changes and updated the PR title.

sagudev:
> I am going to go ahead and try to land this. Sorry for the rush, but I'd like to unstick the Servo merge queue.

I understand, we should probably make scheduled weekly builds on less active repos to prevent such errors in future (rustc error).

I will make companion PR to servo as soon as this lands.

PR #427, opened 2023-11-09
The options to the linker need to be separate elements in the list, n…

JGailor:
The options to the linker were a single element in a list instead of two elements. This was breaking build on MacOS

### DIFF
--- a/mozjs-sys/mozjs/build/moz.configure/toolchain.configure
+++ b/mozjs-sys/mozjs/build/moz.configure/toolchain.configure
@@ -1752,7 +1752,7 @@ def select_linker_tmpl(host_or_target):
             die("Unsupported linker " + linker)
 
         # Check the kind of linker
-        version_check = ["-Wl,--version"]
+        version_check = ["-Wl", "--version"]
         cmd_base = c_compiler.wrapper + [c_compiler.compiler] + c_compiler.flags
 
         def try_linker(linker):


sagudev:
Upstream does not have this: https://searchfox.org/mozilla-central/source/build/moz.configure/toolchain.configure#1736

JGailor:
Can you give some more context on this?  The parameters look wrong in the original code and making the list 2 separate elements was the way I was able to get mozjs and then Servo to compile on a Mac OS Ventura.

sagudev:
mozjs-sys/mozjs folder is from mozilla, so it's  always good to check if upstream has similar patch. But from what I know `-Wl,--version` is correct and should be passed as one argument (see: https://stackoverflow.com/a/6562437) also android CI fails if it's not.

We do run tests on Mac OS Venturain CI, so something might be wrong on your end. Which compiler/linker do you use? Clang with lld linker should work.

sagudev:
Hm, I did some more digging and apparently apple changed their default linker: https://github.com/mesonbuild/meson/issues/12282#issuecomment-1754872730 and we should be able to get old one using `-Wl,-ld_classic` and get version information more regularly using -v option. There is also some interesting info here: https://github.com/mesonbuild/meson/issues/11958 how compiler outputs now differ and https://github.com/mesonbuild/meson/issues/12282#issuecomment-1731467762 with fix for mason.

Good old clang should still work and maybe it's only supported configuration: https://firefox-source-docs.mozilla.org/contributing/build/supported.html#tier-1-targets. Anyhow [this code](https://github.com/servo/mozjs/pull/427/files#diff-6295b695b1c82a3d390b65c8ac462816e4c6b2ff30649b45bf8696c905e09612L1814-L1823) should select lld on too latest xcode; so how exactly does error looks?

mrobinson:
This is the failure message:

```
  DEBUG: Executing: `/usr/bin/cc -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.sdk -mmacosx-version-min=11.0 -std=gnu99 --target=arm64-apple-darwin -Wl,--version`
  ERROR: Failed to find an adequate linker
```

sagudev:
Closed in favor of https://github.com/servo/mozjs/pull/436

Naist4869:
> Closed in favor of #436以#436 结束

["-Wl", "--version"] is work for me ,

return try_linker(linker, "-Wl,-v") not work.

PR #428, opened 2023-11-10
Fix Mac CI and schedule a workflow run weekly

sagudev:
Runs CI every week to catch runner/rust changes.

Includes a fix for mac CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,9 @@ on:
   workflow_dispatch:
   merge_group:
     types: [checks_requested]
+  schedule:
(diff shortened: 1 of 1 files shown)


mrobinson (approved):

Redfire75369:
The merge queue is using the old workflow file instead of the new one, unlike the PR checks.

sagudev:
> The merge queue is using the old workflow file instead of the new one, unlike the PR checks.

No, I think only runners image changed again; I am slowly working on it.

sagudev:
I think this is caused by cc v1.0.84.

Fontsan also fails: https://github.com/servo/servo/pull/30735

PR #430, opened 2023-11-28
Added Traceable for PropertyDescriptor

Redfire75369:


sagudev:
Relevant code in mozilla: https://searchfox.org/mozilla-central/source/js/src/gc/RootMarking.cpp#212

jdm (approved):

Redfire75369:
The error on debug windows arm64 seems spurious? There's no error and seemingly a segfault in rustc?

sagudev:
> The error on debug windows arm64 seems spurious? There's no error and seemingly a segfault in rustc? Could you perhaps try rerunning the check?

I also managed to get this on main: https://github.com/servo/mozjs/actions/runs/7068898831

Redfire75369:
The most likely cause seems to be rust 1.74, although there's a chance its from the upgrade to VS 17.8.34316.

sagudev:
I was diffing from [last know working build](https://github.com/servo/mozjs/actions/runs/7068047786/job/19242172720) and OS image and rust version is the same.

Redfire75369:
Weird, the debug aarch64 windows build didn't fail this time

Redfire75369:
In that case, this certainly seems spurious.

sagudev:
50% of manually dispatched runs failed

sagudev:
Or maybe it is rust problem after all: https://github.com/rust-lang/rust/issues/118462

PR #434, opened 2023-12-02
Modified `Range::new` to take Raw Pointers

Redfire75369:
Small change to #433. `slice::as_mut_ptr_range` gives a `Range<*mut T>`, it would need to be converted to a reference to be used.


sagudev:
Would it be possible to also add test for `Range`?

Redfire75369:
The errors seem to be caused by a C++ ABI issue, which is annoying, as per usual. The moment the `Range<T>` passes to C++, all hell breaks loose and the pointer fields point to nonsense.

mozjs/src/jsglue.cpp
```diff
@@ -555,6 +556,12 @@ bool ShouldMeasureObject(JSObject* obj, nsISupports** iface) {
   return false;
 }
 
+typedef mozilla::Range<const JS::Latin1Char> RangeLatin1;
+
+typedef mozilla::Range<const char16_t> RangeConstUtf16;
+
+typedef mozilla::Range<char16_t> RangeUtf16;
+
```
sagudev, line 564: These aliases are not used.

PR #436, opened 2023-12-06
Fix ld64 detection

sagudev:
This should solve https://github.com/servo/mozjs/pull/427

### DIFF
--- a/mozjs-sys/mozjs/build/moz.configure/toolchain.configure
+++ b/mozjs-sys/mozjs/build/moz.configure/toolchain.configure
@@ -1752,10 +1752,9 @@ def select_linker_tmpl(host_or_target):
             die("Unsupported linker " + linker)
 
         # Check the kind of linker
-        version_check = ["-Wl,--version"]
         cmd_base = c_compiler.wrapper + [c_compiler.compiler] + c_compiler.flags
 
-        def try_linker(linker):
+        def try_linker(linker, version_check="-Wl,--version"):
             # Generate the compiler flag
             if linker == "ld64":
                 linker_flag = ["-fuse-ld=ld"]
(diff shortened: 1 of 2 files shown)


mozjs-sys/mozjs/build/moz.configure/toolchain.configure
```diff
@@ -1779,11 +1778,11 @@ def select_linker_tmpl(host_or_target):
...
             env["LC_ALL"] = "C"
             retcode, stdout, stderr = get_cmd_output(*cmd, env=env)
-            if retcode == 1 and "Logging ld64 options" in stderr:
+            if retcode == 1 and ("Logging ld64 options" in stderr or "PROGRAM:ld PROJECT:dyld" in stderr):
                 kind = "ld64"
 
-            elif retcode != 0:
-                return None
+            elif retcode != 0 and version_check == "-Wl,--version":
+                return try_linker(linker, "-Wl,-v")
```
mrobinson, line 1785: Any reason you aren't exiting early here like in the old version ala:

```
elif retcode != 0
    if version_check == "-Wl,--version":
        try_linker(linker, "-Wl,-v")
   else:
        return None
```

sagudev: Hm, we could, but this way we fallback to -v for all linkers.

mrobinson (approved):

mozjs-sys/mozjs/build/moz.configure/toolchain.configure
sagudev: I did revert this to keep the change simpler and we could reintroduce this if needed.

PR #437, opened 2023-12-06
Switched Bindgen to Use `BINDGEN_CXXFLAGS` Environment Variable

### LINKED ISSUE #435: Windows Builds Fail when `clang-cl` Flags are added to `CXXFLAGS`
The root issue here is that bindgen uses `clang` (not `clang-cl`) on windows, and on unrecognised flags, it fails. Upon adding `CXXFLAGS="/clang:-flto=thin /clang:-fuse-ld=lld-link"`, the below code automatically adds it to bindgen's flags as well. As `/clang:` flags are only recognised by `clang-cl`, there is no way for `cc` and the `makefile.cargo` to receive the `CXXFLAGS` but not bindgen.
https://github.com/servo/mozjs/blob/c7fb1b8d029ba80443da92fbdadb650fcdb493a7/mozjs-sys/build.rs#L334-L338

There are a few possible approaches:
1. If a flag starts with `/clang:` on windows, do not add it to bindgen.
2. Add a separate `BINDGEN_CXXFLAGS` variable for bindgen's flags.

As said earlier, the root issue is the usage of `clang`, which means this error may not be just from `/clang:` flags but almost any `cl` flag. Hence, I would suggest the latter approach.

For a concrete example of the failure, see <https://github.com/Redfire75369/spiderfire/actions/runs/7112793006/job/19363443030#step:6:18053>, where a `libclang error` is caused.

In the event it matters for updating whatever depends on it, #136 was the original PR where this behaviour was introduced.

Redfire75369:
Resolves #435, using the second approach mentioned.
Will require a change in servo at the following location to `env["CXXFLAGS"] = env["BINDGEN_CXXFLAGS"] = ' '.join([`.
https://github.com/servo/servo/blob/bbc35b682f0fb926364e5800d20f77bba944a020/python/servo/command_base.py#L690


sagudev:
There is `BINDGEN_EXTRA_CLANG_ARGS`, could we use it instead of introducing new variable?

Redfire75369:
The issue is `CXXFLAGS` being passed to both bindgen and cc/makefile.cargo, not bindgen itself.

Redfire75369:
Closed in favour of #553.

Issue #439, opened 2023-12-13
feat: add a feature flag to build from source


wusyong:
This is an feature request starting from zulip: https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Spidermonkey.20dylib.20in.20mozjs/near/406266292
Basically, I hope we can offer a feature flag to toggle building mozjs's spidermonkey from source.
Because 99% of people always stuck on building mozjs when they want to try servo.
It would be  awesome if beginners can turn the flag off and download mozjs's spidermonkey binary without compiling and bindgen.
[rusty_v8](https://github.com/denoland/rusty_v8/tree/713758ba135eebe3c63515ab469288ddcbeef3c6#build-v8-from-source) already has mechanism to do so.

wusyong:
Currently I'm testing linking existing static libraries and including generated rust files. It seems working so far.
The static libs need to be linked are: `libjs_static.a`, mozjs-sys's `libjsglue.a`, and mozjs's `libjsglue.a`.
And the generated rust files are: `jsapi.rs` and `gluebindings.rs`.
I'm not sure where to put the files yet. But if there's no suggestion, I'll start by putting them in `OUT_DIR` without any additional directory path.
And I still hope we can push #409, so there will be less files to worry about.

sagudev:
Great stuff, we would probably need two variants one with debug-mozjs feature and one without. Although we could just make release version and if debug-mozjs feature is activated we build it from source.

sagudev:
We would need to extend CI, to automagically produce prebuilds and then make tags for every commit (given that mozjs is not versioned), although this could result a lot of tags.

wusyong:
> Great stuff, we would probably need two variants one with debug-mozjs feature and one without. Although we could just make release version and if debug-mozjs feature is activated we build it from source.

I guess one release version (with `streams` feature) is good enough for now. We don't have enough resources to support multiple version at the moment.



> We would need to extend CI, to automagically produce prebuilds and then make tags for every commit (given that mozjs is not versioned), although this could result a lot of tags.

Maybe we can produce prebuilds on version release. It's also a good improvement for servo since it's still using git dependency I believe. prebuilds per version release seems pretty reasonable and should be more stable for users.

wusyong:
Closed by #450
@sagudev Thanks for reviewing!

sagudev:
Thank you for all hard work.

I will update companion PR later today.

PR #443, opened 2023-12-27
feat: add `MOZJS_MIRROR` and `MOZJS_CREATE_MIRROR` env variables

wusyong:
Blocked by #409 

This PR adds two env variables for mozjs-sys build script. `MOZJS_CREATE_MIRROR` can create a spidermonkey tarball for release usage. While `MOZJS_MIRROR` can use this tarball to extract and link the static libraries without compiling spidermonkey and bindgen wrappers.


wusyong:
This is the next step to push #439. This is still a draft because #409 hasn't been merged yet. And I haven't decided where to put the tarball yet either.
Right now it can already create the compressed file by setting `MOZJS_CREATE_MIRROR=1`, and then, reusing it by `MOZJS_MIRROR=/path/to/tar.gz`.
I'm not sure what github action should be used for releasing it. And I don't have the permission to release either. I'll probably try it with my local fork first to see how it works.

sagudev:
Why don't we just use uncompressed files instead of taring them, like rusty_v8 is doing?

wusyong:
> Why don't we just use uncompressed files instead of taring them, like rusty_v8 is doing?

rusty_v8 only has one `.a` static binary AFAIK. They bind the functions manually so they don't have bindgen files.
Do you mean we just upload every single files and download them one by one?
I use `tar` crate just because it's the first option when I think about grouping the files and extracting them.

wusyong:
Please let me close this draft since my fork isn't actually forked from this repo which cause a bit inconvenience.
And I would like to test a little bit on my own CI pipeline.
I'll open another one pretty soon.

PR #444, opened 2024-01-02
readme: Improve Windows build instructions

LemonHX:


mrobinson (changes requested):
A few comments:

1. Please don't intersperse build instructions with tips and solutions to build problems. Let's separate those out into separate sections.
2. Please use full sentences with a capital letter and a period when writing documentation.
3. Instead of simply writing" if error shows command in `maybe-configure` failed, try" please describe the error that shows up so that the reader knows if the change applies to them. In this case it would be better to know what is going wrong here exactly so we can fix the build script instead of adding a workaround to the build instructions.
4. It would be good to get the exact command necessary to execute with vs_BuildTool.exe rather than listing generally what components are necessary. See how it's done for Servo: https://github.com/servo/servo/wiki/Building
5. Let's move the editor setup to a different file and out of the README which should have the most basic information.

LemonHX:
thank you very much, I will do it soon

Redfire75369:
I'm not sure the instructions should use powershell syntax for setting environment variables.

LemonHX:
> I'm not sure the instructions should use powershell syntax for setting environment variables.

nowadays win11 devs use ps for default also in most recent vscode by default.

LemonHX:
have a look, if that is suitable I will squash commits

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
-3. Open up a shell configured to use Visual Studio. This could be the
-   one included with Visual Studio (e.g. Visual Studio 2017 / X64 Native
-   Tools Command Prompt for VS 2017) or a shell in which you have run:
+2. Download and install Clang for Windows (64 bit) from <https://releases.llvm.org/download.html>. 
+   > Note: LLVM under 14 is not able to build mozjs-sys.
 
-   ```shell
-   "c:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
+3. Download and install `Visual Studio 2022` with `C++ desktop development` with following features:
+   - Windows 11 SDK
```
mrobinson, line 36: Let's use the Windows 10 SDK which is what we are using for Servo as well.

```suggestion
   - Windows 10 SDK
```

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
+      --add Microsoft.VisualStudio.Component.VC.ATL^
+      --add Microsoft.VisualStudio.Component.VC.ATLMFC^
+      --add Microsoft.VisualStudio.Component.VC.CoreIde^
+      --add Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Core^
+      --add Microsoft.VisualStudio.Workload.VCTools
    ```
 
-4. Set the following environment variables according to where you installed
-   the dependencies above:
+1. Install Python3 and add into PATH.
```
mrobinson, line 57:
```suggestion
3. Install [Python 3.11](https://www.python.org/downloads/windows/) or greater.
  - Ensure that Python is added to the system `PATH`
  - Ensure that a `PYTHON` and `PYTHON3` environment variable point to the Python binary (ie `C:\Python311\python.exe`
```

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
 
+2. Set the following environment variables according to where you installed
```
mrobinson, line 59:
```suggestion
4. Set the following environment variables according to where you installed
```

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
+	  CC="$(CC)" CFLAGS="$(CFLAGS)" \
+	  CPP="$(CPP)" CPPFLAGS="$(CPPFLAGS)" \
+	  CXX="$(CXX)" CXXFLAGS="$(CXXFLAGS)" \
+	  AS="$(AS)" AR="$(AR)" \
+	  $(JSSRC)/configure $(strip $(CONFIGURE_FLAGS)) || (cat config.log && exit 1) ; \
+	fi
+
+   exit code: not 0
    ```
+   try to modify `mozjs/mozjs/js/src` replace line  `PYTHON3="${PYTHON3:-python3}"` with `PYTHON3="${PYTHON3:-python}"`
```
mrobinson, line 89: I think by setting the `PYTHON3` environment variable above you can remove this workaround.

mrobinson (changes requested):
This is looking a lot better. A few suggestions:

README.md
sagudev: I think mozilla does not always support the latest IIRC, but they definitely support older (python 3.8): https://firefox-source-docs.mozilla.org/build/buildsystem/python.html

mrobinson: I've modified this to say 3.11.

sagudev: All I am saying is that I would rather lower python version recommendation, preferably to 3.8, as it is in mozilla.

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
 ### Windows
-
 1. Download and unzip [MozTools 4.0](https://github.com/servo/servo-build-deps/releases/download/msvc-deps/moztools-4.0.zip).
 
-2. Download and install Clang for Windows (64 bit) from <https://releases.llvm.org/download.html>.
-
-3. Open up a shell configured to use Visual Studio. This could be the
-   one included with Visual Studio (e.g. Visual Studio 2017 / X64 Native
-   Tools Command Prompt for VS 2017) or a shell in which you have run:
+2. Download and install Clang for Windows (64 bit) from <https://releases.llvm.org/download.html>.
```
Redfire75369, line 32: Do we want to suggest downloading LLVM through Visual Studio? It's what we use in CI after all.

README.md
```diff
@@ -27,26 +27,66 @@ export LIBCLANG_PATH=/usr/lib/clang/4.0/lib
...
+      --add Microsoft.VisualStudio.Workload.VCTools
    ```
 
-4. Set the following environment variables according to where you installed
-   the dependencies above:
+1. Install Python3 and add into PATH.
 
+2. Set the following environment variables according to where you installed
+   the dependencies above:
    ```shell
```
Redfire75369, line 61:
```suggestion
   ```powershell
```
Since it's powershell syntax, the highlighting should probably be changed.

Redfire75369 (changes requested):

README.md
sagudev: I mean "Python 3.8 or greater" that includes 3.11; it's just to make sure that others do not get wrong impression of minimal supported python version.

README.md
LemonHX: sure! shall I remove this?
but btw I am not sure env is effected in this build file

README.md
LemonHX: I have no idea, I and my colleagues are downloading the most recent manually.
I uses theird party bin
he is using scoop

README.md
mrobinson: Does properly setting the `PYTHON3` environment variable remove the need for manually editing the build files? If that's the case, I think we should definitely remove this.

mrobinson (approved):
I have made a few edits, but otherwise this looks good.

PR #445, opened 2024-01-08
feat: add `MOZJS_ARCHIVE` and `MOZJS_CREATE_ARCHIVE` env variables

wusyong:
This PR adds two env variables for mozjs-sys build script:
- `MOZJS_CREATE_ARCHIVE=1` can create a spidermonkey tarball for release usage.
- `MOZJS_ARCHIVE=path/to/libmozjs.tar.gz` can use this tarball to extract and link the static libraries without compiling spidermonkey and bindgen wrappers.

I'm also testing [github release](https://github.com/wusyong/mozjs/releases/tag/tag-8410b587d66a36f1660cc3b828359e199eb0760a) to publish these archives. If you are interested in testing, feel free to download it and try it.
After downloading it, run `MOZJS_MIRROR=path/to/tarball.tar.gz cargo test` to see if that works.


mozjs-sys/build.rs
```diff
@@ -52,12 +56,21 @@ fn main() {
...
 
-    fs::create_dir_all(&build_dir).expect("could not create build dir");
-
-    build_spidermonkey(&build_dir);
-    build_jsapi(&build_dir);
-    build_jsapi_bindings(&build_dir);
-    jsglue::build(&build_dir);
+    let mirror = env::var_os("MOZJS_MIRROR");
+    if mirror.is_some() {
+        download_static_lib_binaries(&PathBuf::from(mirror.unwrap()), &build_dir);
```
Redfire75369, line 61:
```suggestion
    if let Some(mirror) = env::var_os("MOZJS_MIRROR") {
        download_static_lib_binaries(&PathBuf::from(mirror), &build_dir);
```

mozjs-sys/build.rs
```diff
@@ -3,12 +3,16 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use bindgen::Formatter;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use flate2::Compression;
 use std::env;
 use std::ffi::{OsStr, OsString};
-use std::fs;
+use std::fs::{self, File};
```
Redfire75369, line 11: Although there isn't a consistent import style yet, this should probably be changed.
```suggestion
use std::fs;
use std::fs::File;
```

mozjs-sys/build.rs
```diff
@@ -702,3 +716,97 @@ mod jsglue {
         ("root::JS", "pub(crate) use crate::jsapi::JS::*;"),
     ];
 }
+
+/// Compress spidermonkey build into a tarball with necessary static binaries and bindgen wrappers.
+fn compress_static_lib(build_dir: &Path) -> Result<(), std::io::Error> {
+    let target = env::var("TARGET").unwrap();
+    let tar_gz = File::create(String::from("libmozjs-") + &target + ".tar.gz")?;
```
Redfire75369, line 723:
```suggestion
    let tar_gz = File::create(format!("libmozjs-{}.tar.gz", target))?;
```

Redfire75369 (changes requested):

Redfire75369:
If we get this to a point where we have published tars for all the major platforms, we could technically consider publishing on crates.io again, although it wouldn't support build from source.

wusyong:
One of missing ones is probably macOS arm target since github action hasn't offered this runner for free yet.

wusyong:
@jdm Sorry to ping again, but I would like to know if Servo is still interested in this. More specifically, #439.
I already have another branch in my fork to publish my releases: https://github.com/wusyong/mozjs/pull/5
And I have the draft in #450 to finish the feature request.

I hope this PR can be merged first because this one won't affect any existing behavior yet.
And then we can discuss how should the release CI look like.
I also want to push more regular versioning to mozjs, so the release can be more stable.

sagudev (commented):
`MOZJS_MIRROR` does not work with relative paths (or maybe has wrong cwd for them)

sagudev:
Variable names are weird to me (what does mirror mean in this context?).

sagudev:
> I also want to push more regular versioning to mozjs, so the release can be more stable.

wusyong:
> `MOZJS_MIRROR` does not work with relative paths (or maybe has wrong cwd for them)

Relative paths have to be relative to the build script of `mozjs-sys` instead of current directory. Can we state it should be absolute paths? I fear handling relative paths may fall into many pitfalls from each OS. I added this note to the README when explaining env variables.

> Variable names are weird to me (what does mirror mean in this context?).

I took [`rusty_v8`'s env variables](https://github.com/denoland/rusty_v8?tab=readme-ov-file#the-rusty_v8_mirror-environment-variable) as reference. I don't have strong opinion on this. Do you have prefer conventions?

> I am just curious what happens if you use wrong precompiled stuff. Do we need to have any guards for such cases?

For servo, it needs `streams` feature. So if we link a prebuilt binary without this feature, it will crash and state it can't find the extern functions.

> Actually I am more in favor of what we have now, as currently mozjs is still pretty rough, but I think we should discuss this somewhere else.

I see, let's discuss this in future issue. Doing release per commit is still fine for now since mozjs doesn't update that often, and I also managed to reduce the [binary size](https://github.com/wusyong/mozjs/releases/tag/tag-b04fa9bdb7013fe31065c92b0264d1508577200a) as small as possible.

sagudev:
For the naming rusty_v8 has `RUSTY_V8_ARCHIVE` for offline functionality so replacing `MIRROR` with `ARCHIVE` would make more sense to me.

> So if we link a prebuilt binary without this feature, it will crash and state it can't find the extern functions.

What happens if version mismatches? I am asking if we need to do something like version/commit guard.

> Relative paths have to be relative to the build script of mozjs-sys instead of current directory. Can we state it should be absolute paths?

Maybe making it relative to parent of target folder would be better. I am mostly thinking of this to make build archive -> use own archive simpler (`MOZJS_CREATE_MIRROR` by default puts tar into mozjs-sys IIRC, maybe this should be in target so git does not track it?).

wusyong:
@sagudev 
> For the naming rusty_v8 has `RUSTY_V8_ARCHIVE` for offline functionality so replacing `MIRROR` with `ARCHIVE` would make more sense to me.

Okay, I renamed the variables to `MOZJS_CREATE_ARCHIVE` and `MOZJS_ARCHIVE`.

> What happens if version mismatches? I am asking if we need to do something like version/commit guard.

It will fail to run and crash due to unable to find the foreign function. v8_rusty has version check for the URL.
mozjs doesn't have version release yet but I think commit guard is also fine since mozjs doesn't have many signature changes often.
I can add the guard later once we add the CI to release and another variable to pull from URL.

> Maybe making it relative to parent of target folder would be better. I am mostly thinking of this to make build archive -> use own archive simpler (`MOZJS_CREATE_MIRROR` by default puts tar into mozjs-sys IIRC, maybe this should be in target so git does not track it?).

I updated the behavior a bit, and now the archive is created in the target directory. `MOZJS_ARCHIVE` will also try to get the relative path from workspace if it fails to open first time.

sagudev:
> I can add the guard later once we add the CI to release and another variable to pull from URL.

So this would be part of follow up or also in this PR?

If you want to land just request my review via GitHub UI so I know when it's ready.

wusyong:
It will be in the following PR.

sagudev (approved):
LGTM

Thank you for all hard work!

PR #447, opened 2024-01-13
dependencies: Update bindgen

mrobinson:


sagudev:
This update still does not include float16 fix unfortunately.

sagudev (approved):

mrobinson:
It looks like 0.69.2 includes the fix based on the diff:

https://github.com/rust-lang/rust-bindgen/compare/v0.69.2...main

sagudev:
> It looks like 0.69.2 includes the fix based on the diff:
> 
> [rust-lang/rust-bindgen@v0.69.2...main](https://github.com/rust-lang/rust-bindgen/compare/v0.69.2...main)

I think you need to compare between tags: https://github.com/rust-lang/rust-bindgen/compare/v0.69.1...v0.69.2

Comparing tag from main just shows what is on main but not in tag I think.

mrobinson:
Huh. Either way, as long as the released version is 0.69.3, this should be useful.

PR #448, opened 2024-01-15
Use ANDROID_NDK_ROOT instead of ANDROID_NDK_HOME

mrobinson:
ANDROID_NDK_ROOT seems to be the more common environment variable
nowadays.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -120,7 +120,7 @@ jobs:
           targets: armv7-linux-androideabi
       - name: Build
         env:
-          ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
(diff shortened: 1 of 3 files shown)


sagudev:
There is still one use in mc: https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/android.py#270, but there are also (less important) uses of _ROOT https://searchfox.org/mozilla-central/search?q=ANDROID_NDK_ROOT&path=&case=false&regexp=false

mrobinson:
> There is still one use in mc: https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/android.py#270, but there are also (less important) uses of _ROOT https://searchfox.org/mozilla-central/search?q=ANDROID_NDK_ROOT&path=&case=false&regexp=false

I did see that usage, but it looks like it only used when executing `android.py` as a Python script?

mukilan (approved):

PR #450, opened 2024-01-24
feat: add MOZJS_FROM_SOURCE 

wusyong:
Blocked by #445 and probably need a CI from servo org to publish libmozjs.

This is the last step of #439. Everything should work now but it's based on my fork at the moment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,10 +21,10 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - { target: libmozjs-aarch64-apple-darwin.tar.gz, os: macos-14 }
-          - { target: libmozjs-x86_64-apple-darwin.tar.gz, os: macos-13 }
-          - { target: libmozjs-x86_64-unknown-linux-gnu.tar.gz, os: ubuntu-latest }
-          - { target: libmozjs-x86_64-pc-windows-msvc.tar.gz, os: windows-latest }
+          - { target: aarch64-apple-darwin, os: macos-14 }
+          - { target: x86_64-apple-darwin, os: macos-13 }
+          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
+          - { target: x86_64-pc-windows-msvc, os: windows-latest }
     runs-on: ${{ matrix.platform.os }}
     env:
       RELEASE_TAG: ${{ github.event_name == 'release' && github.ref_name ||  inputs.release-tag }}
@@ -34,11 +34,11 @@ jobs:
         if: ${{ env.RELEASE_TAG == '' }}
         uses: actions/download-artifact@v4
         with:
-          name: ${{ matrix.platform.target }}
+          name: libmozjs-${{ matrix.platform.target }}.tar.gz
       - name: Build from archive
         if: ${{ env.RELEASE_TAG == '' }}
         env:
-          MOZJS_ARCHIVE: ${{ matrix.platform.target }}
+          MOZJS_ARCHIVE: libmozjs-${{ matrix.platform.target }}.tar.gz
         run: |
           cargo build --verbose --features streams
           cargo test --tests --examples --verbose --features streams
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,26 +21,29 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - { target: aarch64-apple-darwin, os: macos-14 }
-          - { target: x86_64-apple-darwin, os: macos-13 }
-          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
-          - { target: x86_64-pc-windows-msvc, os: windows-latest }
+          - { target: libmozjs-aarch64-apple-darwin.tar.gz, os: macos-14 }
+          - { target: libmozjs-x86_64-apple-darwin.tar.gz, os: macos-13 }
+          - { target: libmozjs-x86_64-unknown-linux-gnu.tar.gz, os: ubuntu-latest }
+          - { target: libmozjs-x86_64-pc-windows-msvc.tar.gz, os: windows-latest }
     runs-on: ${{ matrix.platform.os }}
     env:
       RELEASE_TAG: ${{ github.event_name == 'release' && github.ref_name ||  inputs.release-tag }}
-      MOZJS_ARCHIVE: libmozjs-${{ matrix.platform.target }}.tar.gz
     steps:
       - uses: actions/checkout@v4
       - name: Download prebuilt mozjs from artifact
         if: ${{ env.RELEASE_TAG == '' }}
         uses: actions/download-artifact@v4
         with:
-          name: libmozjs-${{ matrix.platform.target }}.tar.gz
-      - name: Download prebuilt mozjs from release
-        if: ${{ env.RELEASE_TAG != '' }}
+          name: ${{ matrix.platform.target }}
+      - name: Build from archive
+        if: ${{ env.RELEASE_TAG == '' }}
+        env:
+          MOZJS_ARCHIVE: ${{ matrix.platform.target }}
         run: |
-          curl -L https://github.com/${{ github.repository_owner }}/mozjs/releases/download/${{ env.RELEASE_TAG }}/libmozjs-${{ matrix.platform.target }}.tar.gz -o libmozjs-${{ matrix.platform.target }}.tar.gz
-      - name: Build
+          cargo build --verbose --features streams
+          cargo test --tests --examples --verbose --features streams
+      - name: Build from auto-download
+        if: ${{ env.RELEASE_TAG != '' }}
         run: |
           cargo build --verbose --features streams
           cargo test --tests --examples --verbose --features streams
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,6 +17,7 @@ env:
   CARGO_INCREMENTAL: 0
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MOZJS_CREATE_ARCHIVE: 1
+  MOZJS_FROM_SOURCE: 1
 
 jobs:
   mac:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,6 @@ env:
   CARGO_INCREMENTAL: 0
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MOZJS_CREATE_ARCHIVE: 1
-  MOZJS_FROM_SOURCE: 1
 
 jobs:
   mac:
--- a/README.md
+++ b/README.md
@@ -9,27 +9,36 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
 Mozjs is currently tracking SpiderMonkey on [ESR-115](https://searchfox.org/mozilla-esr115/source/) branch
 (currently version 115.3).
 
-## Binary Build
+## Building from Pre-built Archive
 
-Spidermonkey is very large which could take a long time to compile. If `streams` feature is enabled, mozjs provides prebuilt archive to download
-and link during build script by default. You can also create your own archive and link to it. Mozjs currently offer two environment variables to enable such work:
+SpiderMonkey is very large which could take a long time to compile. If building with default
+feature, `mozjs` provides a pre-built archive which can be linked against. You can also create
+your own archive and link to it. `mozjs` currently offers two environment variables to enable
+this feature:
 
-- `MOZJS_CREATE_ARCHIVE=1` can create a spidermonkey tarball for release usage. It will be created in `target` directory.
-- `MOZJS_ARCHIVE` can use this tarball without compiling spidermonkey and bindgen wrappers. It has 2 usages to specify the archive:
-   - `MOZJS_ARCHIVE=path/to/libmozjs.tar.gz` will find the tarball in local path, extract, and then link to the static libraries.
-   - `MOZJS_ARCHIVE=https://url/to/release/page` will download the archive with provided url as base, extract, and then link to it. The base url should
-   be similar to `https://github.com/servo/mozjs/releases/download`. The build script will append the version and target accordingly.
+- `MOZJS_CREATE_ARCHIVE=1` will create a SpiderMonkey binary tarball for release usage. It will
+   be created in the `target` directory.
+- `MOZJS_ARCHIVE` can be used to build against a pre-built archive. Using this flag compiling
+   SpiderMonkey and the bindgen wrappers is unnecessary. There are two ways to use it:
+   - `MOZJS_ARCHIVE=path/to/libmozjs.tar.gz`: This option will look for the archive at the local
+      path, extract it, and then link against the static libraries included in the archive.
+   - `MOZJS_ARCHIVE=https://url/to/release/page`: This option will download the archive from
+      the provided base URL, extract it, and then link against the static libraries included in the
+      archive. The base URL should  be similar to `https://github.com/servo/mozjs/releases/`.
+      The build script will append the version and target accordingly. See the files at the example
+      URL for more details.
 
 ## Building from Source
 
-If `MOZJS_FROM_SOURCE=1` or `MOZJS_CREATE_ARCHIVE` are enabled or steps to link prebuilt archive fails, mozjs will try to build the spidermonky from source. 
+If `MOZJS_FROM_SOURCE=1` or `MOZJS_CREATE_ARCHIVE` are enabled or linking against a
+pre-built archive fails, `mozjs` will build SpiderMonkey from source.
 
 ### Linux
 
 Install Python, Clang and `build-essential`, for example on a Debian-based Linux:
 
 ```sh
-sudo apt-get install build-essential python3 python3-distutils llvm libclang-dev clang
+sudo apt-get install build-essential python3 python3-distutils llvm libclang-dev clang curl
 ```
 
 If you have more than one version of Clang installed, you can set the `LIBCLANG_PATH`
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@ Mozjs is currently tracking SpiderMonkey on [ESR-115](https://searchfox.org/mozi
 
 ## Binary Build
 
-Spidermonkey is very large which could take a long time to compile. Mozjs provides prebuilt archive to download and link during build script by default.
-You can also create your own archive and link to it. Mozjs currently offer two environment variables to enable such work:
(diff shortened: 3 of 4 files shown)


sagudev:
Maybe we need to automatically enable `MOZJS_FROM_SOURCE` if features are not right but not if `MOZJS_ARCHIVE` is provided (in this case we trust user that it knows what it is doing).

No rush, because we should wait for #445.

mozjs-sys/build.rs
```diff
@@ -57,9 +57,20 @@ fn main() {
...
+    let build_from_source = match env::var_os("MOZJS_FROM_SOURCE") {
+        Some(_) => true,
+        None => match link_static_lib_binaries(&build_dir) {
+            Ok(()) => false,
+            Err(e) => {
+                println!("cargo:warning=Failed to link pre-built archive by {e}. Building from source instead.");
+                true
+            }
+        },
+    };
```
wusyong, line 70: I updated the step of linking prebuilt archive to return Result. If it fails, it will try to build it from source instead.
This means if the target isn't correct or the path `MOZJS_ARCHIVE` doesn't exist, it will continue to build from source.

wusyong:
The action seems to work on [PR](https://github.com/wusyong/mozjs/actions/runs/8262811284) and [main branch](https://github.com/wusyong/mozjs/actions/runs/8263141641). But I haven't added the job to test auto-downloaded prebuilt. What's preferred job or step to test this?

sagudev:
> But I haven't added the job to test auto-downloaded prebuilt. What's preferred job or step to test this?

I think adjusting release.yml to support auto-downloading (instead of manually specifying release/download) would be the way to go. Actually we could always test with auto-download, except when this would require new release (mozjs-sys version was bumped), so in that case we need to fallback to testing artifacts on PR and MQ (but we can auto-download when pushed on main as the release should be made by the same run).

mozjs-sys/build.rs
sagudev: I would just make that we fallback if automatic fails, so if one used `MOZJS_ARCHIVE`, they wanted something explicitly so we should hard fail to let them know something is wrong.

sagudev: Also I think we should only fallback if download or extraction fail, if there is any other problem we should hard fail.

.github/workflows/rust.yml
```diff
@@ -17,6 +17,7 @@ env:
   CARGO_INCREMENTAL: 0
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MOZJS_CREATE_ARCHIVE: 1
+  MOZJS_FROM_SOURCE: 1
```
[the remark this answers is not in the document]
wusyong: Do you mean only extend `MOZJS_CREATE_ARCHIVE`? Or adding `MOZJS_FROM_SOURCE` but also make `MOZJS_CREATE_ARCHIVE` cover what it does?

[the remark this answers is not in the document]
sagudev: Second one, so that when you want to create archive you do not need to set `MOZJS_FROM_SOURCE` seperatly.

wusyong:
Behaviors of `MOZJS_CREATE_ARCHIVE` and `MOZJS_ARCHIVE` have been updated based on review.
And I added another step in `release.yml` to auto-download archive and build when it's pushed on main branch.

.github/workflows/release.yml
```diff
@@ -21,26 +21,29 @@ jobs:
...
       matrix:
         platform:
-          - { target: aarch64-apple-darwin, os: macos-14 }
-          - { target: x86_64-apple-darwin, os: macos-13 }
-          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest }
-          - { target: x86_64-pc-windows-msvc, os: windows-latest }
+          - { target: libmozjs-aarch64-apple-darwin.tar.gz, os: macos-14 }
+          - { target: libmozjs-x86_64-apple-darwin.tar.gz, os: macos-13 }
+          - { target: libmozjs-x86_64-unknown-linux-gnu.tar.gz, os: ubuntu-latest }
+          - { target: libmozjs-x86_64-pc-windows-msvc.tar.gz, os: windows-latest }
```
sagudev, line 27: I would just keep it like it was and then just use `MOZJS_ARCHIVE: libmozjs-${{ matrix.platform.target }}.tar.gz` instead of `MOZJS_ARCHIVE: ${{ matrix.platform.target }}`.

mozjs-sys/build.rs
```diff
@@ -57,17 +57,30 @@ fn main() {
     // Used by mozjs downstream, don't remove.
     println!("cargo:outdir={}", build_dir.display());
 
-    if let Some(archive) = env::var_os("MOZJS_ARCHIVE") {
-        download_static_lib_binaries(&PathBuf::from(archive), &build_dir);
+    // Link to pre-built archive first if it exists.
+    let create_archive = env::var_os("MOZJS_CREATE_ARCHIVE").is_some();
+    let build_from_source = if env::var_os("MOZJS_FROM_SOURCE").is_some() || create_archive {
+        true
```
sagudev, line 63: If I am reading things correctly, if features mismatch with what artifacts provide we run compilation with artifacts anyway and then on failure (if it even fails), we do fallback. I think it would be nicer to fallback to build from source directly. So if `streams` feature is not selected or `debug-mozjs` feature is enabled. And when building from source is enabled we should print reason for this using `cargo:info` for easier debugging.

wusyong: I added two conditions to check streams and debug-mozjs features. Do they match what you described?

sagudev: It's good. Is there any specific reason for mixing `env::var` vs `env::var_os`, or can we only use one variant everywhere?

Separating `MOZJS_FROM_SOURCE` and `MOZJS_CREATE_ARCHIVE` into separate if's with reasons info would also make sense, right?

Otherwise this looks really good, I will do more test after #459 to make sure everything works as expected and then we can land this.

wusyong:
@sagudev Sine #459 has been merged, could this PR be reviewed again? Or should it need to wait for any other PR?

sagudev:
I will do it later today (and if I forgot do not hesitate to ping me tomorrow).

mozjs-sys/build.rs
```diff
@@ -57,17 +57,40 @@ fn main() {
     // Used by mozjs downstream, don't remove.
     println!("cargo:outdir={}", build_dir.display());
 
-    if let Some(archive) = env::var_os("MOZJS_ARCHIVE") {
-        download_static_lib_binaries(&PathBuf::from(archive), &build_dir);
+    // Link to pre-built archive first if it exists.
+    let create_archive = env::var_os("MOZJS_CREATE_ARCHIVE").is_some();
+    let build_from_source = if env::var_os("MOZJS_FROM_SOURCE").is_some() {
+        println!("cargo:info=Environment variable MOZJS_FROM_SOURCE is set. Building from source directly.");
```
sagudev, line 63: `cargo:info` does not actually exist 😭. So we should either change this to `cargo:warning` or use normal prints, which will be available only when very verbose output is enabled. I prefer the letter.

wusyong: Okay I'll update to normal prints. It seems build script error will also show std output.

sagudev (approved):
LGTM

But given the nature of this change, maybe this needs another pair of eyes.

README.md
```diff
@@ -9,7 +9,20 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
...
+
+- `MOZJS_CREATE_ARCHIVE=1` can create a spidermonkey tarball for release usage. It will be created in `target` directory.
+- `MOZJS_ARCHIVE` can use this tarball without compiling spidermonkey and bindgen wrappers. It has 2 usages to specify the archive:
+   - `MOZJS_ARCHIVE=path/to/libmozjs.tar.gz` will find the tarball in local path, extract, and then link to the static libraries.
+   - `MOZJS_ARCHIVE=https://url/to/release/page` will download the archive with provided url as base, extract, and then link to it. The base url should
+   be similar to `https://github.com/servo/mozjs/releases/download`. The build script will append the version and target accordingly.
+
+## Building from Source
+
+If `MOZJS_FROM_SOURCE=1` or `MOZJS_CREATE_ARCHIVE` are enabled or steps to link prebuilt archive fails, mozjs will try to build the spidermonky from source.
```
mrobinson, line 25:
```suggestion
SpiderMonkey is very large which could take a long time to compile. If building with default
feature, `mozjs` provides a pre-built archive which can be linked against. You can also create
your own archive and link to it. `mozjs` currently offers two environment variables to enable
this feature:

- `MOZJS_CREATE_ARCHIVE=1` will create a SpiderMonkey binary tarball for release usage. It will
   be created in the `target` directory.
- `MOZJS_ARCHIVE` can be used to build against a pre-built archive. Using this flag compiling
   SpiderMonkey and the bindgen wrappers is unnecessary. There are two ways to use it:
   - `MOZJS_ARCHIVE=path/to/libmozjs.tar.gz`: This option will look for the archive at the local
      path, extract it, and then link against the static libraries included in the archive.
   - `MOZJS_ARCHIVE=https://url/to/release/page`: This option will download the archive from
      the provided base URL, extract it, and then link against the static libraries included in the
      archive. The base URL should  be similar to `https://github.com/servo/mozjs/releases/`.
      The build script will append the version and target accordingly. See the files at the example
      URL for more details.

## Building from Source

If `MOZJS_FROM_SOURCE=1` or `MOZJS_CREATE_ARCHIVE` are enabled or linking against a
pre-built archive fails, `mozjs` will build SpiderMonkey from source.
```

README.md
```diff
@@ -9,7 +9,20 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
 Mozjs is currently tracking SpiderMonkey on [ESR-115](https://searchfox.org/mozilla-esr115/source/) branch
 (currently version 115.3).
 
-## Building
+## Binary Build
```
mrobinson, line 12:
```suggestion
## Building from Pre-built Archive
```

mozjs-sys/build.rs
```diff
@@ -57,17 +57,42 @@ fn main() {
...
+        println!(
+            "Environment variable MOZJS_CREATE_ARCHIVE is set. Building from source directly."
+        );
+        true
+    } else if env::var_os("CARGO_FEATURE_DEBUGMOZJS").is_some() {
+        println!("debug-mozjs feature is enabled. Building from source directly.");
+        true
+    } else if !env::var_os("CARGO_FEATURE_STREAMS").is_some() {
+        println!("streams feature isn't enabled. Building from source directly.");
+        true
```
mrobinson, line 75: Let's extract this into a helper called `should_build_from_source()` that can use early returns.

mozjs-sys/build.rs
```diff
@@ -57,17 +57,42 @@ fn main() {
...
         build_jsapi(&build_dir);
         build_jsapi_bindings(&build_dir);
         jsglue::build(&build_dir);
 
         // If this env variable is set, create the compressed tarball of spidermonkey.
-        if env::var_os("MOZJS_CREATE_ARCHIVE").is_some() {
+        if create_archive {
             compress_static_lib(&build_dir).expect("Failed to compress static lib binaries.");
         }
     }
```
mrobinson, line 98: These two parts of the code are alternatives, so they should be two bodies of an if-else statement.

mozjs-sys/build.rs
```diff
@@ -791,22 +816,53 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
 }
 
-/// Download static library tarball instead of building it from source.
-fn download_static_lib_binaries(archive: &Path, build_dir: &Path) {
-    // Only download the files if build directory doesn't exist.
-    if !build_dir.exists() {
-        // TODO download from https
-        decompress_static_lib(archive, build_dir).expect("Failed to decompress static libs");
+/// Download spidermonkey archive by curl with provided base url. If it's None, it will use
+/// servo/mozjs's release page as base url.
```
mrobinson, line 828:
```suggestion
/// Download the SpiderMonkey archive with cURL using the provided base URL. If it's None,
/// it will use `servo/mozjs`'s release page as the base URL.
```

mozjs-sys/build.rs
```diff
@@ -791,22 +816,53 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
+fn download_archive(base: Option<&str>) -> Result<PathBuf, std::io::Error> {
+    let base = base.unwrap_or("https://github.com/servo/mozjs/releases/download");
+    let version = env::var("CARGO_PKG_VERSION").unwrap();
+    let target = env::var("TARGET").unwrap();
+    let archive_path = PathBuf::from(env::var_os("OUT_DIR").unwrap()).join("libmozjs.tar.gz");
+    if !archive_path.exists() {
+        if !Command::new("curl")
```
mrobinson, line 835: We should document somewhere that cURL is now a dependency.

mozjs-sys/build.rs
```diff
@@ -791,22 +816,53 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
+            .success()
+        {
+            return Err(std::io::Error::from(std::io::ErrorKind::NotFound));
+        }
     }
 
+    Ok(archive_path)
+}
+
+/// Link static library tarball instead of building it from source.
```
mrobinson, line 854:
```suggestion
/// Link against the static library archive instead of building SpiderMonkey from source.
```

mozjs-sys/build.rs
```diff
@@ -791,22 +816,53 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
+fn link_static_lib_binaries(build_dir: &Path) -> Result<(), std::io::Error> {
+    if let Ok(archive) = env::var("MOZJS_ARCHIVE") {
+        // If there's archive variable, assume it's a url base to download first
+        // If not, assign it as a local path
```
mrobinson, line 858:
```suggestion
        // If the archive variable is present, assume it's a URL base to download from.
```

mozjs-sys/build.rs
```diff
@@ -791,22 +816,53 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
+            return Err(std::io::Error::from(std::io::ErrorKind::NotFound));
+        }
     }
 
+    Ok(archive_path)
+}
+
+/// Link static library tarball instead of building it from source.
+fn link_static_lib_binaries(build_dir: &Path) -> Result<(), std::io::Error> {
+    if let Ok(archive) = env::var("MOZJS_ARCHIVE") {
```
mrobinson, line 856: Could this be mapped to an `Option` instead of repeating the `download_archive` and `decompress_static_lib` calls below?

mrobinson (changes requested):
Looks pretty good. I just have a few minor comments.

mozjs-sys/build.rs
wusyong: I added in mozjs's README. I'll also add them to servo's README later.

README.md
```diff
@@ -9,14 +9,36 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
 Mozjs is currently tracking SpiderMonkey on [ESR-115](https://searchfox.org/mozilla-esr115/source/) branch
 (currently version 115.3).
 
-## Building
+## Building from Pre-built Archive
+
+SpiderMonkey is very large which could take a long time to compile. If building with default
+feature, `mozjs` provides a pre-built archive which can be linked against. You can also create
+your own archive and link to it. `mozjs` currently offers two environment variables to enable
+this feature:
```
mrobinson, line 17: Sorry noticed some mistakes in my version:

```suggestion
SpiderMonkey is very large and can take a long time to compile. If building with default
features, `mozjs` provides a pre-built archive which can be linked against. You can also create
your own archive and link to it. `mozjs` currently offers two environment variables to enable
this feature:
```

README.md
```diff
@@ -9,14 +9,36 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
...
+
+- `MOZJS_CREATE_ARCHIVE=1` will create a SpiderMonkey binary tarball for release usage. It will
+   be created in the `target` directory.
```
mrobinson, line 20:
```suggestion
- `MOZJS_CREATE_ARCHIVE=1` will create a SpiderMonkey binary archive for release usage. It will
   be created in the `target` directory.
```

README.md
```diff
@@ -9,14 +9,36 @@ that are battle-tested in [Servo](https://servo.org/), split in two crates:
...
+- `MOZJS_ARCHIVE` can be used to build against a pre-built archive. Using this flag compiling
```
mrobinson, line 21:
```suggestion
- `MOZJS_ARCHIVE` can be used to build against a pre-built archive. Using this flag, compiling
```

mozjs-sys/build.rs
```diff
@@ -91,6 +90,34 @@ fn main() {
...
+        println!("streams feature isn't enabled. Building from source directly.");
+        true
+    } else {
+        match link_static_lib_binaries(&build_dir) {
+            Ok(()) => false,
+            Err(e) => {
+                println!("cargo:warning=Failed to link pre-built archive by {e}. Building from source instead.");
+                true
+            }
+        }
```
mrobinson, line 117: Please move this of the helper and call it based on the return value. For instance:

```rust
let mut build_from_source = should_build_from_source(&build_dir);
if !build_from_source && link_static_lib_binaries(&build_dir).is_err() {
    println!("cargo:warning=Failed to link pre-built archive by {e}. Building from source instead.");
    build_from_source = true;
}
```

mozjs-sys/build.rs
mrobinson: It would be nice to add this for Windows as well.

mozjs-sys/build.rs
mrobinson: How about something like this:

```rust
   let archive = let Ok(archive) = env::var("MOZJS_ARCHIVE") {
        download_archive(Some(&archive)).unwrap_or(PathBuf::from(archive));
    } else {
        download_archive(None)?;
    };
    decompress_static_lib(&archive, build_dir)?;
```

sagudev:
> It would be nice to add this for Windows as well.

curl is preinstalled on windows 10/11.

mozjs-sys/build.rs
wusyong: I believe @sagudev wants `decompress_static_lib` in the first branch to have a hard fail because it manually specifies `MOZJS_ARCHIVE`, and we want it to panic instead of fall back.

mozjs-sys/build.rs
wusyong: Windows has builtin curl and both cmd and powershell seem working in my local test.

mrobinson (approved):

wusyong:
@sagudev Is it okay to merge this PR?

Issue #446, opened 2024-01-10
Make ‎TypedArray::update safe


gterzian:
I think [`‎TypedArray::update`](https://github.com/servo/mozjs/blob/7184f658b7e2de8a8d288ff39001364513115782/mozjs/src/typedarray.rs#L255) can be made safe, while containing an unsafe block calling `Self::update_raw`, that way we can remove these [kind of unsafe blocks](https://github.com/servo/servo/pull/30990/files#diff-88b03ac9bcc2650883607d41bc374d7f3976567cfbee702cc124f14ce2fbc79bR283) in Servo.

Taym95:
How do you think we can do this @gterzian, from code I see unsafe code call is chained, update -> update_raw -> copy_nonoverlapping

gterzian:
I think `update` should simply be safe externally, but internally have an unsafe block that calls `update_raw`.

Taym95:
this can be closed.

PR #452, opened 2024-02-12
Make TypedArray::update externally safe by adding unsafe block

Taym95:
#446

### DIFF
--- a/mozjs/src/typedarray.rs
+++ b/mozjs/src/typedarray.rs
@@ -252,8 +252,10 @@ impl<T: TypedArrayElementCreator + TypedArrayElement, S: JSObjectStorage> TypedA
     }
 
     ///  Update an existed JS typed array
-    pub unsafe fn update(&mut self, data: &[T::Element]) {
-        Self::update_raw(data, self.object.as_raw());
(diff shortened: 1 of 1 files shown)


mozjs/src/typedarray.rs
```diff
@@ -252,8 +252,10 @@ impl<T: TypedArrayElementCreator + TypedArrayElement, S: JSObjectStorage> TypedA
     }
 
     ///  Update an existed JS typed array
-    pub unsafe fn update(&mut self, data: &[T::Element]) {
-        Self::update_raw(data, self.object.as_raw());
+    pub fn update(&mut self, data: &[T::Element]) {
+        unsafe {
+            Self::update_raw(data, self.object.as_raw());
+        }
```
Taym95, line 258: @gterzian do you think we can merge this one?

gterzian: Yes I think so. @jdm ?

jdm: Yeah, I don't see any way to.misyse this interface.

jdm (approved):

PR #453, opened 2024-02-13
Implement TypedArrayElementCreator for ArrayBufferViewU8

Taym95:


Redfire75369:
An `ArrayBuffer` isn't an `ArrayBufferView`, so this isn't really correct.

Taym95:
> An `ArrayBuffer` isn't an `ArrayBufferView`, so this isn't really correct.

Yes got it, I am just testing it around

PR #454, opened 2024-02-29
Add prebuilt mozjs archive CI

wusyong:
Another part of #439. This PR adds a CI pipeline to publish prebuilt mozjs archive.
I've tried to run it and created a working version release: https://github.com/wusyong/mozjs/releases/tag/2024-02-29
I set the workflow trigger on the main branch update since there are few commits for this repo.
Let me know if there's preferred condition.


sagudev (commented):
I am not really fan of doing manual uploads using curl commands. We should use actions to make this less complex. You can take a look how this is done in servo: https://github.com/servo/servo/blob/main/.github/workflows/nightly.yml. In servo we also mark releases that are not full as draft (and mark them as final if all runs succeed).

I think mac-arm and x86 could be merged into one job with matrix.

sagudev:
Furthermore I think we can simply use main workflow insted of adding new one and add `MOZJS_CRATE_ARCHIVE=1` to main (as it does not change an compilation only adds packaging step), the upload archives as artifects and if run as done on landing (push to main) actually release artifacts as GitHub release. This should avoid us repeating a lot of lines and allow faster testing of ARCHIVES (as they are stored as artifacts also on forks).

sagudev:
Then there is also option to start CI on release, that could check that archives actually work.

wusyong:
Alright! Let me move the workflow into main one. ~~I use curl because I didn't find any action could move action artifact to release asset, and `actions/upload-release-asset` is unmaintained. Do you have any recommended action that can upload artifact first and then move to release asset later?~~ Edit: I think I can create two steps instead. Let me test them a bit later.

sagudev:
I am not sure we want to do releases on forks. Only on servo/servo would be enough (but only for push to main, we would want to make release for every PR). There is `github.event_name` for this IIRC.

Also to simplify things even more, instead of firstly creating draft release  we should just do normal run uploading archives as artifacts (already done). Then at the end do release step that would download all artifacts and upload them all to the release.

Also upload-release-asset@v1 seems to be unmaintained, maybe we can do this by [gh command](https://stackoverflow.com/questions/64936180/how-to-release-on-github-through-the-command-line-and-to-attach-a-large-file).

wusyong:
I need `github.repository_owner` to test on my own fork. I can make it only releases on `servo/mozjs`.

>Also to simplify things even more, instead of firstly creating draft release we should just do normal run uploading archives as artifacts (already done).

I'll simplify the workflow a bit. I also aware of gh command but have some trouble with multiple build steps. Maybe it can be easier if the release workflow is simplified.

wusyong:
@sagudev  Could I create another yaml for publish release? Or is it better to include in the same `rust.yml`?

wusyong:
[Tested ](https://github.com/wusyong/mozjs/actions/runs/8110908841)and [released](https://github.com/wusyong/mozjs/releases/tag/2024-03-01) by simplified job. Let me know if there is something need improved.
I'll work on a CI job to verify release asset can be built and tested.

.github/workflows/rust.yml
```diff
@@ -65,18 +77,24 @@ jobs:
...
 
   windows:
     runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
-        features: ["--features debugmozjs", ""]
-        target: ["x86_64-pc-windows-msvc", "aarch64-pc-windows-msvc"]
+        features: ["--features debugmozjs", "--features streams"]
+        target: ["", "--target aarch64-pc-windows-msvc"]
```
sagudev, line 97: Why we need to add `--target` in here?

wusyong: I hope it can build and test without `--target` if the target is the same as host.
Otherwise, the archive will be created in `target/x86_64-pc-windows-msvc` due to the limitation of [`get_cargo_target_dir`](https://github.com/servo/mozjs/blob/main/mozjs-sys/build.rs#L724)

wusyong:
There are some [restrictions](https://github.com/orgs/community/discussions/25281) on recursive calls on workflow and auto release. So I chose `workflow_call` to create another workflow to check release binaries. Here's the [successful](https://github.com/wusyong/mozjs/actions/runs/8121042312) action that creates a release and verify on my fork.

I think that's all for the workflow that can create releases and then verify it when the release is published. Let me know if I still miss or need to fix something.

.github/workflows/rust.yml
sagudev: I think we could use `OUT_DIR` for this. Here are result for default than with target invocation:
```
/home/user/Desktop/lol/target/debug/build/lol-bb298d56b00857ef/out
/home/user/Desktop/lol/target/x86_64-unknown-linux-gnu/debug/build/lol-a67d370db0f3eb4c/out
```
so we need to go up (to parrent) four times. Then if current folder is named as rust target (`x86_64-unknown-linux-gnu`) we need to go up one more time. There is also `CARGO_TARGET_DIR` that allows users to override location of target folder (this is used in servo's CI due to space constraints on default storage device (D): https://github.com/servo/servo/blob/main/.github/workflows/windows.yml#L41)

sagudev:
> There are some [restrictions](https://github.com/orgs/community/discussions/25281) on recursive calls on workflow and auto release.

I completely forgot about those. Does it fail running release CI job or does it just not get run. In second case we might still want to have it trigger on release (in addition to workflow call) in case we do manual release.

.github/workflows/release.yml
sagudev, line 1: Isn't the whole point of ARCHIVES to reduce amount of dependencies needed for compiling? So all jobs should only clone repo, download proper archive and run `cargo test` without any other stuff.

.github/workflows/rust.yml
wusyong: I guess I can specify the exact target path, so other parts don't need to be adjusted.
I'll revert back to `target: ["x86_64-pc-windows-msvc", "aarch64-pc-windows-msvc"]` and upload the artifact with the according path.

.github/workflows/release.yml
wusyong: You are right. I'll remove all deps steps and running directly with cargo test.

wusyong:
> I completely forgot about those. Does it fail running release CI job or does it just not get run. If it's the later case we might still want to have it trigger on release (in addition to workflow call) in case we do manual release.

It won't be triggered to run. I'll add the condition to trigger on manual release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+    needs:
+      ["linux", "mac", "windows"]
+    if: ${{ always() }}
+    steps:
+      - name: Mark the job as successful
+        if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
+        run: exit 0
+      - name: Mark the job as unsuccessful
+        if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
+        run: exit 1
```
sagudev, line 106: This workflow does not need result job, but result job in rust.yml would need to include creating release jobs too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,107 @@
...
+    inputs:
+      release-tag:
+        required: true
+        type: string
+
+env:
+  RUST_BACKTRACE: 1
+  SHELL: /bin/bash
+  CARGO_INCREMENTAL: 0
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
sagudev, line 17: This is unused.

sagudev, line 17: Actually I think all 3 are not needed here.

.github/workflows/release.yml
sagudev: I think we can now squash all os jobs to one matrixed job.

.github/workflows/release.yml
wusyong: okay, it's removed and result job in rust.yml needs release jobs now.

sagudev (commented):
Testing artifacts currently requires release, maybe we could always run release.yml but if no release exists we could fallback to artifacts. This way artifacts are always being tested (this is mostly relevant on PR checking).

.github/workflows/rust.yml
```diff
@@ -15,14 +15,19 @@ env:
...
+  MOZJS_CREATE_ARCHIVE: 1
 
 jobs:
   mac:
-    runs-on: macos-13
     strategy:
       fail-fast: false
       matrix:
-        features: ["--features debugmozjs", ""]
+        features: ["--features debugmozjs", "--features streams"]
```
sagudev, line 26: We could extract `--features` from matrix as it is common to all matrix fields.

.github/workflows/rust.yml
```diff
@@ -178,3 +202,32 @@ jobs:
       - name: Mark the job as unsuccessful
         if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
         run: exit 1
+
+  publish-release:
+    name: Publish release
+    runs-on: ubuntu-latest
+    needs:
+      ["linux", "mac", "windows"]
+    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
```
sagudev, line 211: This condition is not needed here, as this job is not run if any needed fails. This whole job should have if guard from step bellow.

wusyong: I can remove this if guard. But I think if I move the if guard from steps below, verify job might be [skipped](https://github.com/wusyong/mozjs/actions/runs/8135726635) as well if this release job is skipped.

sagudev: I think skipped is mapped to cancelled result.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - id: tag-name
+        run: |
+          if [[ '${{ inputs.release-tag }}' != '' ]]; then
+            echo "RELEASE_TAG=${{ inputs.release-tag }}" >> ${GITHUB_OUTPUT}
+          elif [[ '${{ github.event_name }}' == 'release' ]]; then
+            echo "RELEASE_TAG=${{ github.ref_name }}" >> ${GITHUB_OUTPUT}
+          fi
+    outputs:
+      tag: ${{ steps.tag-name.outputs.RELEASE_TAG }}
```
sagudev, line 30: I think we could rewrite this as env var: `RELEASE_TAG: ${{ github.event_name == 'release' && github.ref_name ||  inputs.release-tag }}`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  verify-artifact:
```
sagudev, line 32:
```suggestion
  verify-archive:
```

.github/workflows/rust.yml
```diff
@@ -178,3 +202,32 @@ jobs:
...
+      release-tag: ${{ steps.release.outputs.RELEASE_TAG }}
+
+  verify-release:
+    name: Verify release
+    needs: publish-release
+    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
+    uses: ./.github/workflows/release.yml
+    with:
+      release-tag: ${{ needs.publish-release.outputs.release-tag }}
+
```
sagudev, line 233: This should be moved above build_result job (so the jobs are ordered by their invocation).

.github/workflows/release.yml
wusyong: Thanks for the suggestion. I didn't know github action variable can simplify to this.

.github/workflows/rust.yml
```diff
@@ -165,11 +189,39 @@ jobs:
...
+    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+        with:
+          merge-multiple: true
+      - id: release
+        if: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
+        run: |
+          RELEASE_TAG=$(date "+%F-%H-%M")
```
sagudev, line 206: This could be problematic, if we have two updates in a day (it happens sometimes), so using commit hash might be better, but I don't know how we could bake this info in for https://github.com/servo/mozjs/pull/450. Can we get dependency commit hash using cargo without baking it in.

sagudev: Rusty_v8 does this using `CARGO_PKG_VERSION`:
https://github.com/denoland/rusty_v8/blob/9dd629e1c3c50fbbe70435ed50f6f09e9b69d7bb/build.rs#L320

wusyong: `CARGO_PKG_VERSION` seems like a better approach. I also hope this repo can provide version release again.
I'll update `release.yml` to use mozjs_sys's `CARGO_PKG_VERSION`, and it will skip the release if the version is already exist.

sagudev: I think `CARGO_PKG_VERSION` is actually the only way to go and because we moved all .c* files to mozjs-sys, the only thing that needs versions bumps is `mozjs-sys` when there are any C/C++ file changes.

It would be nice to make CI check for this too, but I think this can be done as a followup if (ever) needed. For now I think it would be enough if we test always latest release on push to main (instead of testing release only if it's new).

.github/workflows/rust.yml
```diff
@@ -165,11 +189,44 @@ jobs:
...
+    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v4
+        with:
+          merge-multiple: true
+      - id: check-tag
+        if: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
+        run: |
+          RELEASE_TAG=mozjs-sys-v$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "mozjs_sys") | .version')
```
sagudev, line 206:
```suggestion
          RELEASE_TAG=mozjs-sys-v$(cargo pkgid -p mozjs_sys | cut -d "@" -f2)
```

wusyong: cargo pkgid requires Cargo.lock. I'll see if I can move this step to previous building job.

sagudev: > cargo pkgid requires Cargo.lock

Didn't know that. Then it might just be better what you wrote originally. Sorry for trouble.

wusyong: Okay I'll revert it.

sagudev:
We should not fail here, right?

> a release with the same tag name already exists: mozjs-sys-v0.68.2

https://github.com/sagudev/mozjs/actions/runs/8207125754/job/22448383805#step:5:11

sagudev:
I think we should just skip release creation in this case, but still verify it.

wusyong:
> We should not fail here, right?
> 
> > a release with the same tag name already exists: mozjs-sys-v0.68.2
> 
> https://github.com/sagudev/mozjs/actions/runs/8207125754/job/22448383805#step:5:11

This [SO ](https://stackoverflow.com/questions/19298600/tag-already-exists-in-the-remote-error-after-recreating-the-git-tag) question looks similar to us. I'll add `git fetch --tags` before checking.

sagudev:
> > We should not fail here, right?
> > > a release with the same tag name already exists: mozjs-sys-v0.68.2
> > 
> > 
> > https://github.com/sagudev/mozjs/actions/runs/8207125754/job/22448383805#step:5:11
> 
> This [SO ](https://stackoverflow.com/questions/19298600/tag-already-exists-in-the-remote-error-after-recreating-the-git-tag) question looks similar to us. I'll add `git fetch --tags` before checking.

I am not sure this will solve it. I simply run workflow twice (second time with empty commit), both times on main. I think we just need to not fail in this case. So maybe something like `gh create release ... || exit 0`.

.github/workflows/rust.yml
```diff
@@ -165,11 +189,44 @@ jobs:
...
+          if git show-ref --tags --verify --quiet "refs/tags/${RELEASE_TAG}"; then
+            RELEASE_TAG=''
+          fi
```
sagudev, line 209: Or did you refer to that code? In that case we can simply remove this and make next step to not fail, which will ensure to test all merged commits with latest release available.

wusyong: Oh sorry I didn't read your comments before pushing. `gh create release ... || exit 0` seems simpler. I'll update it.

sagudev: What you did is actually way better!

wusyong: ah okay! Let me revert it. 😅

.github/workflows/rust.yml
```diff
@@ -165,11 +189,43 @@ jobs:
...
+        with:
+          merge-multiple: true
+      - name: Publish release if tag doesn't exist
+        id: check-tag
+        if: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
+        run: |
+          RELEASE_TAG=mozjs-sys-v$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "mozjs_sys") | .version')
+          git fetch --tags --quiet
+          if ! git show-ref --tags --verify --quiet "refs/tags/${RELEASE_TAG}" ; then
+            gh release create ${{ steps.check-tag.outputs.RELEASE_TAG }} ./*.tar.gz
```
sagudev, line 210: `${{ steps.check-tag.outputs.RELEASE_TAG }}` is not available here yet (step is not finished), but `${RELEASE_TAG}` should work. https://github.com/sagudev/mozjs/actions/runs/8213051004/job/22464164401#step:4:5

sagudev (approved):
Thanks for all your great work!

I can't wait to start using this work in servo.

wusyong:
@sagudev  Thanks for reviewing the PR too! I'll have to work on multi webview suppoer PR next.
But once it's done, I'll get back and update #450.

PR #458, opened 2024-03-18
Add `-ohos` target support to autoconf

jschwe:
Adds support for OpenHarmony OS.
A patch has been submitted upstream, pending review: https://lists.gnu.org/archive/html/config-patches/2024-02/msg00008.html
After the patch has been merged upstream, we can update `config.sub` in spidermonkey and drop this patch.

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -138,8 +138,9 @@ fn cc_flags() -> Vec<&'static str> {
 
     let is_apple = target.contains("apple");
     let is_freebsd = target.contains("freebsd");
+    let is_ohos = target.contains("ohos");
 
-    if is_apple || is_freebsd {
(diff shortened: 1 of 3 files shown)


jdm (approved):

jschwe:
Rebased, so hopefully CI will pass now.

sagudev:
> Rebased, so hopefully CI will pass now.

I planed to do that myself but I forgot. Thanks for work.

jschwe:
@sagudev Do I still need to do something here? It seems CI failed because the macOS job was cancelled - Is this PR blocked on something else?

sagudev:
> @sagudev Do I still need to do something here? It seems CI failed because the macOS job was cancelled - Is this PR blocked on something else?

I didn't cancel anything. Let's try again and see if something is really wrong or was it just a flake.

PR #460, opened 2024-03-26
Rely only on `pkg-config` to find zlib include directory

sagudev:
This solves https://github.com/servo/servo/issues/31851 for me. The problem is that if libz is installed on system we add it's path to includes which causes problems. I think we do not need this as SM build system depends on pkg-config to find this.

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -221,10 +221,6 @@ fn build_spidermonkey(build_dir: &Path) {
         encoding_c_mem_include_dir.replace("\\", "/"),
     ));
     cppflags.push(" ");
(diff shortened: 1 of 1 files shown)


sagudev:
Relevant configure code: https://github.com/servo/mozjs/blob/b695c63b6d832eff191f6785dd56539376ae4332/mozjs-sys/mozjs/moz.configure#L895-L936

sagudev:
System zlib is only used on linux for any other platform (android, darwin, windows) it is built from sources, and the only way to detect it on system is via pkg-config (which windows env does not have even in moztools).

mrobinson (approved):

PR #464, opened 2024-05-29
Respect cc-rs style environment variables when building SM

### LINKED ISSUE #463: build script: Respect same environment variables as `cc-rs`
[`cc-rs`](https://docs.rs/cc/latest/cc/) has [a prioritized list of prefixed variants of environment variables](https://docs.rs/cc/latest/cc/#external-configuration-via-environment-variables) that it respects.
To make downstream users life easier (especially those that are cross-compiling), the `mozjs` build-script should ideally also respect the same variants, otherwise `mozjs`s build behavior differs from the majority of crates, that use `cc-rs`.

Implementation proposal: 

Before building spidermonkey, the build-scipt should iterate through the variants that `cc-rs` reads, and set the base environment variable, if any of the higher priority variants are set in the parent environment.

jschwe:
Downstream projects such as servo often also have many other C/C++ dependencies that are built with cc-rs, so having the Spidermonkey build respecting the same variables, makes configuring the downstream build easier and more predictable.

Closes #463

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -926,8 +926,8 @@ fn get_cc_rs_env(var_base: &str) -> Option<String> {
 /// Like `get_cc_rs_env()` but returns the OsString value.
 fn get_cc_rs_env_os(var_base: &str) -> Option<OsString> {
     fn get_env(var: &str) -> Option<OsString> {
-        let value = env::var_os(var)?;
         println!("cargo:rerun-if-env-changed={}", var);
+        let value = env::var_os(var)?;
         Some(value)
     }
     let target = env::var("TARGET").expect("Cargo should set TARGET");
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -33,6 +33,20 @@ const ENV_VARS: &'static [&'static str] = &[
     "STLPORT_LIBS",
 ];
 
+// For `cc-rs` `TARGET_XX` variables override non prefixed variables,
+// so we should mimic this behavior when building spidermonkey to have a consistent experience.
+const SM_TARGET_ENV_VARS: &'static [&'static str] = &[
+    "AR",
+    "AS",
+    "CC",
+    "CFLAGS",
+    "CLANGFLAGS",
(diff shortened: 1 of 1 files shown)


mozjs-sys/build.rs
```diff
@@ -889,3 +911,35 @@ fn link_static_lib_binaries(build_dir: &Path) -> Result<(), std::io::Error> {
...
+    get_cc_rs_env_os(var_base).map(|val| val.to_str().expect("Not a valid string.").to_string())
+}
+
+/// Like `get_cc_rs_env()` but returns the OsString value.
+fn get_cc_rs_env_os(var_base: &str) -> Option<OsString> {
+    fn get_env(var: &str) -> Option<OsString> {
+        let value = env::var_os(var)?;
+        println!("cargo:rerun-if-env-changed={}", var);
+        Some(value)
+    }
```
sagudev, line 932: I am not sure about this, I would run this unconditionally just to be on the safe side.

sagudev:
Couldn't we just call [cc to get us compiler](https://docs.rs/cc/latest/cc/struct.Build.html#method.get_compiler), instead rolling our own?

mozjs-sys/build.rs
jschwe: Thanks for catching this. Indeed, we want to recompile if a previously unset variable is now set.

jschwe:
> Couldn't we just call [cc to get us compiler](https://docs.rs/cc/latest/cc/struct.Build.html#method.get_compiler), instead rolling our own?

It's not just the compiler, but also `AR`,  `CXXFLAGS` and others. `pkg-config-rs` also [maintains its own copy of the code](https://github.com/rust-lang/pkg-config-rs/blob/246ece2c314552f891692fb66fce2027156b38be/src/lib.rs#L621) which does the same as the `cc-rs` version.

sagudev (approved):

sagudev:
For reference, this is what we are doing for https://github.com/servo/servo-build-deps/blob/main/docker/cross-aarch64-unknown-linux-gnu.dockerfile

PR #465, opened 2024-05-30
Also respect readelf and objcopy cc-rs variants

jschwe:
While compiling servo for OpenHarmony on a MacOS host, I noticed that readelf and objcopy apparently aren't chosen from the same directory as the clang c compiler, but needed to be in PATH for the configure script to pick them up. To allow users to set READELF and OBJCOPY explicitly, we should also read the relevant cc-rs environment variables.

An alternative would have been to inspect the configure scripts and find out why they don't look next to clang, but I'm not familiar enough with the configure script.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
           brew list -1 | grep python | while read formula; do brew unlink $formula; brew link --overwrite $formula; done
           brew install llvm yasm
       - name: Run sccache-cache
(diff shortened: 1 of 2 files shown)


jschwe:
Just adding `readelf` and `objcopy` to the list of variables caused CI for macos to fail - this was somehow related to `sccache`. Disabling sccache and updating the sccache github action both fixed CI, so I'll just update the sccache action in this PR.

sagudev (approved):

sagudev:
Rebased!

Issue #462, opened 2024-04-20
Notes on ESR 128


sagudev:
Just some things that I have to be careful about (because I will probably forget them by the summer when ESR lands):
- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/D194086.patch can be dropped due to https://bugzilla.mozilla.org/show_bug.cgi?id=1853564
- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0029-fix-ld64-detection.patch is still in progress: https://bugzilla.mozilla.org/show_bug.cgi?id=1844694
- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0010-update-mozboot-to-ndk-r25.patch can be dropped due to https://bugzilla.mozilla.org/show_bug.cgi?id=1822171
- ~~investigate https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0027-No-ErrorStackFrames-in-clones.patch: https://github.com/servo/servo/issues/32648~~ DONE
- Remove streams: https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0026-js-streams.patch, WIP: https://github.com/servo/servo/issues/29088
- New MozillaBuild "needs" to be packed into new moztools (not sure if we really need it, but it's nice to have)

wusyong:
https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0010-update-mozboot-to-ndk-r25.patch can be obselete I think. Upstream has newer version of ndk already.

sagudev:
> https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0010-update-mozboot-to-ndk-r25.patch can be obselete I think. Upstream has newer version of ndk already.

You are right!

sagudev:
This will be problematic: https://bugzilla.mozilla.org/show_bug.cgi?id=1824671 as it requires usage of cbindgen ...

sagudev:
Moving to 0 based column number to 1 based column number: https://bugzilla.mozilla.org/show_bug.cgi?id=1862692

sagudev:
Also https://bugzilla.mozilla.org/show_bug.cgi?id=1620223

sagudev:
Changes to offthread compilation are big:
https://bugzilla.mozilla.org/show_bug.cgi?id=1846392
https://bugzilla.mozilla.org/show_bug.cgi?id=1846194
[...]

PR #474, opened 2024-07-10
Update SpiderMonkey to ESR128.0

sagudev:
Thankfully we were more prepared now: https://github.com/servo/mozjs/issues/462

It compiles! Streams patch was successfully ported.

Biggest changes:
- For intl support we need to compile icu_capi from our cargo.toml (similar to what we do with encoding_c)
- There were some changes around zlib (when it's allowed to use sys zlib) so I ripped all zlib out to always use zlib provided by libz-sys crate
- Support for offthread compilation is removed, so we will need to roll our own (before it was done as a task on JS Helpers thread poll), this is planned to be done as a follow up according to plan outlined here: https://github.com/servo/mozjs/pull/474#issuecomment-2225751703
- min mac sdk is now on 14.2 (needs CI fixes: xcode 15.2, will become default https://github.com/actions/runner-images/issues/10120)
- min NDK is 26
- we now build proper static lib in windows, reducing artifact size: https://github.com/servo/mozjs/pull/474/commits/1a084f0738914385b89eedb779f0b3cd3c3f7ead

Companion PR in servo: https://github.com/servo/servo/pull/32769


sagudev:
Another problem: https://bugzilla.mozilla.org/show_bug.cgi?id=1887728, we need macos sdk 14.4, but this requires Xcode 15.4, that is not available on macos 13 (so we got no x86 runners). I think it would be possible to patch it for this ESR cycle.

sagudev:
I was wrong, streams files were not even compiled (only header files), because patch was missing:
```python
if CONFIG["MOZ_JS_STREAMS"]:
    UNIFIED_SOURCES += [
        "builtin/Stream.cpp",
...
```
and there are compilation errors (I knew it was to good to be true).

sagudev:
@wusyong Any idea why this is happening: https://github.com/servo/mozjs/actions/runs/9902233667/job/27356251099#step:4:826, I manually checked and there is s_truncf.obj in zip?

wusyong:
> @wusyong Any idea why this is happening: https://github.com/servo/mozjs/actions/runs/9902233667/job/27356251099#step:4:826, I manually checked and there is s_truncf.obj in zip?

Hmm, it's weird because I checked out to your branch and used that artifact and it seems fine. I'll look into it.

wusyong:
Okay I got the error. It seems the hash isn't correct?

`note: lld-link: error: D:\a\mozjs\mozjs\target\debug\build\mozjs_sys-b0f045240069a423\out\build/js/src/build\js_static.lib: could not get the buffer for a child of the archive: 'D:/a/mozjs/mozjs/target/x86_64-pc-windows-msvc/debug/build/mozjs_sys-c2a3d4c65cace61a/out/build/modules/fdlibm/src/s_truncf.obj': no such file or directory`

mozjs-sys/build.rs
```diff
@@ -287,6 +305,9 @@ fn build_spidermonkey(build_dir: &Path) {
         }
         cmd.env("PKG_CONFIG_PATH", pkg_config_path);
     }
+    if let Ok(include) = std::env::var("DEP_Z_INCLUDE") {
+        cppflags.push(format!("-I{include}").replace("\\", "/"));
+    }
```
wusyong, line 310: What's this path used for? I wonder if the Windows error is caused by the absolute path when building Spidermonkey.

wusyong:
I guess we should take this chance to finally build static library on msvc without linking any object files.

sagudev:
> The build is mozjs_sys-b0f045240069a423 but the linker somehow search for mozjs_sys-c2a3d4c65cace61a

something must be causing hash change.

> I guess we should take this chance to finally build static library on msvc without linking any object files.

Ideally yes.

sagudev:
Doing `ar r js/src/build/js_static.lib modules/fdlibm/src/*.obj` should be enough I think.

wusyong:
> Doing `ar r js/src/build/js_static.lib modules/fdlibm/src/*.obj` should be enough I think.

I'm wondering if there's a flag that can enable it. Because other platforms' static libraries just need the `.a` file.

sagudev:
> > Doing `ar r js/src/build/js_static.lib modules/fdlibm/src/*.obj` should be enough I think.
> 
> I'm wondering if there's a flag that can enable it. Because other platforms' static libraries just need the `.a` file.

I think something is wrong with build sys (in SM), that doesn't make static really static on windows.

mozjs-sys/build.rs
```diff
@@ -308,6 +329,17 @@ fn build_spidermonkey(build_dir: &Path) {
...
+    if target.contains("windows") {
+        let mut make_static = cc::Build::new();
+        for obj in std::fs::read_to_string(build_dir.join("js/src/build/js_static_lib.list"))
+            .unwrap()  // panic on possible file-reading errors
+            .lines()  // split the string into an iterator of string slices
+            .map(String::from) {
+            make_static.object(obj);
+        }
+        make_static.static_flag(true);
+        make_static.compile("js_static");
```
wusyong, line 341:
```suggestion
        make_static.static_flag(true);
        make_static.out_dir(build_dir.join("js/src/build"));
        make_static.compile("js_static");
```

It worked with this change! But artifact becomes [super huge](https://github.com/wusyong/mozjs/actions/runs/9905684451). I'll see if I can simplify the compressed tar ball on Windows.

sagudev: We can remove most stuff from artifact anyway and only keep what we keep on linux.

The reason for size inflation is that cc-rs makes two identical-sized libs (same thing also happens with jsglue and jsapi):
slika

sagudev: > I'll see if I can simplify the compressed tar ball on Windows.

Okay, in the mean time I will continue to work on other things, send PR to my fork when you are done (although we could land this separately, but then I would have to rebase ...)

wusyong: Opened https://github.com/sagudev/mozjs/pull/2

sagudev: Tnx, I did squashed all relevant commits into one: https://github.com/servo/mozjs/pull/474/commits/1a084f0738914385b89eedb779f0b3cd3c3f7ead (authorship was preserved).

sagudev:
For offthread compilation I plan to make `compileToStencil` function that will do everything the threads needs to do, then implement `CompileToStencilOffThread` in mozjs (now with rust/safe api that will still resemble ESR115 API), that will create new thread on each invocation (the thread will run `compileToStencil`).

So servo will be migrated to this `CompileToStencilOffThread` for now (in the future we can use `compileToStencil` to implement thread pooled version of `CompileToStencilOffThread`).

Not sure if this will be done as part of SM bump effort or as a follow up (depending on the amount of changes needed).

sagudev:
I will land this myself, after also companion PR get's reviewed.

jdm (approved):
Fantastic work figuring out all these new API changes and build issues!

PR #478, opened 2024-07-24
Replace more backslashes with forward slashes on windows

jschwe:
- All environment variables that contain paths and are used by the cargo makefile, need to have backslashes changed to forward slashes.


jschwe:
The original title of this PR was: 

> Draft: Fix cross-compiling on windows to non-windows targets

However, I've since given up on this. It seems that the SM build-system has the assumption that on windows hosts, cross-compilation will also target a windows OS. I did a bit of investigation, but concluded that its not worth the effort, given that downloading the prebuilt object from CI works and avoids the whole problem of building spidermonkey on windows for a non-windows OS.

PR #481, opened 2024-07-28
Added Support for External Latin1 Strings

Redfire75369:
I've added support for external latin1 strings, added in [bug 1576076](https://bugzilla.mozilla.org/show_bug.cgi?id=1576076). I've also updated the `.clang-format` based on [bug 1851529](https://bugzilla.mozilla.org/show_bug.cgi?id=1851529), and reformatted the C++ files accordingly.

Requires #474

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,7 @@
-# Taken from https://searchfox.org/mozilla-central/rev/aa1bb4f0ca8bfda4b117d1befca47b72d5dd6d5d/.clang-format
+# Adapted from https://hg.mozilla.org/mozilla-central/file/12bd668f66103e727d3198e600cb25aa4ea8b6dc/.clang-format
 
 BasedOnStyle: Google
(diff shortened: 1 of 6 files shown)


sagudev:
This will need a rebase and an increment mozjs-sys version (and probably companion PR in servo, but I can take care of it if you want after #474 is in servo).

sagudev:
I will check this in the upcoming week, but it would be nice to also have tests for both ExternalStrings.

Redfire75369:
Added Some Tests based on the External String code in spiderfire in [ion/string/mod.rs](https://github.com/Redfire75369/spiderfire/blob/master/ion/src/string/mod.rs#L86-L118) and [ion/src/utils.rs](https://github.com/Redfire75369/spiderfire/blob/master/ion/src/utils.rs).

sagudev:
Needs rebase.

sagudev (approved):

PR #482, opened 2024-07-29
Added Wrapper for NewArrayBufferWithContents

Redfire75369:
Removing the two wrapped array buffer functions from the bindings also shortens imports (`mozjs::jsapi::<function>` as compared to `mozjs::jsapi::glue::<function>`).


jdm (approved):

sagudev:
This will need rebase after #483 lands, so let's wait for it to land.

Redfire75369:
Rebased on latest, with version bump accordingly.

PR #484, opened 2024-07-29
Added Wrapped NewArrayBufferWithContents to Header

Redfire75369:
I forgot to add it to the header in #482, so here's a fix for that.


sagudev:
I hate header files (so error prone). What if we just use .cpp files and feed those to bindgen? (we already do this in mozangle).

sagudev:
Actually jsglue.cpp is already without header, so we should make jsapi.cpp too.

jdm (approved):

PR #486, opened 2024-07-29
Remove jsapi header and feed .cpp into bindgen

sagudev:
Header files are pain (they must be kept in sync) and they are only consumed by bindgen anyway.


jdm (approved):

mrobinson:
This looks like a nice cleanup!

PR #488, opened 2024-07-31
Log errors to hilog on OpenHarmony

jschwe:
Similar to android, stderr output is generally not visible on OpenHarmony devices. Error messages should instead be logged via the `hilog` logging service. 
During normal operations, we can also do a workaround where we redirect stderr to hilog via a dedicated logging thread. 
However, for fatal errors such as failed assertions, which lead to the whole process dying, this solution is not sufficient, since the process might die before the logging thread has a chance to print the error message it received on `stderr`.

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.0-5"
+version = "0.128.0-6"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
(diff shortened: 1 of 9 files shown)


sagudev:
About symbol visibility you should read https://github.com/servo/servo/pull/32769#issuecomment-2253857177, so I think you only need to do something like:
```py
system_headers += [
    'hilog/log.h',
]
```

also this is something that is not really servo specific, so you should try to upstream it.

jschwe:
> so I think you only need to do something like:

Thanks! That looks exactly like what I want. 

> also this is something that is not really servo specific, so you should try to upstream it.

That is my plan. But I want to get it working with servo first, and then tackle rebasing to latest upstream (and hope upstream is interested).

sagudev:
All PRs that change mozjs-sys/mozjs or other cpp code need to bump mozjs-sys version.

sagudev (approved):

PR #489, opened 2024-07-31
Fix enforcerange for 64bit numbers

sagudev:
While working on WebGPU I noticed that `EnforceRange` is not working right, per [specs](https://webidl.spec.whatwg.org/#abstract-opdef-converttoint), limits for 64bit values are 2**+-53 -+1.

See also relevant limits overrides in mozilla: https://searchfox.org/mozilla-esr128/rev/1aa97f9d67f7a7231e62af283eaa02a6b31380e1/dom/bindings/PrimitiveConversions.h#166

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /doc
 /target
 /Cargo.lock
+/.vscode
 
 # Keep this in sync with ignore() in build.rs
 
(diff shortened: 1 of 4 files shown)


sagudev:
I think this needs companion PR in servo, just to be on the safe side (although WPT does not have real enforcerange tests).

sagudev:
Try run in servo has no changes: https://github.com/sagudev/servo/actions/runs/10185061444/job/28175496263

jdm (approved):
I really like the new test!

PR #490, opened 2024-08-04
Added Clang Format Checks to CI

Redfire75369:
Adds C++ Formatting to the CI under the Integrity Check.

I suspect we could switch to `check-path=mozjs-sys/src` instead of `exclude-regex` since `mozjs` doesn't have any c++ anymore.


sagudev:
> I suspect we could switch to check-path=mozjs-sys/src instead of exclude-regex since mozjs doesn't have any c++ anymore.

I think it's ok the way it is.

sagudev (approved):

PR #491, opened 2024-08-05
Fix mmap for JIT code on HarmonyOS NEXT Beta 2

jschwe:
The Security policy for JIT code is tightening on HarmonyOS . Beta 2 and newer require us to pass `MAP_EXECUTABLE` (internally same as `MAP_JIT`) to the committing mmap, otherwise it will fail.

### DIFF
--- a/mozjs-sys/etc/patches/0034-build__Add_compile_definition_for_ohos.patch
+++ b/mozjs-sys/etc/patches/0034-build__Add_compile_definition_for_ohos.patch
@@ -1,0 +1,18 @@
+diff --git a/build/moz.configure/init.configure b/build/moz.configure/init.configure
+--- a/build/moz.configure/init.configure	(revision dbffebd0937c14d3c73ce9be4798da15cb2f369d)
++++ b/build/moz.configure/init.configure	(revision 8a1dbb7da45148151dbb0d05d969118103cfa8d0)
+@@ -950,6 +950,14 @@
+ set_define("ANDROID", target_is_android)
+
+
++@depends(target)
++def target_is_ohos(target):
++    return target.raw_os.endswith("-ohos")
++
++
++set_define("XP_OHOS", target_is_ohos)
++
++
+ @depends(target)
+ def target_is_openbsd(target):
+     if target.kernel == "OpenBSD":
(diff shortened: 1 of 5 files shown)


mozjs-sys/etc/patches/0034-build__Add_compile_definition_for_ohos.patch
```diff
@@ -0,0 +1,20 @@
...
+--- a/build/moz.configure/init.configure	(revision 3720f6d208b45fb968961a8d5c97f64010ae3d81)
++++ b/build/moz.configure/init.configure	(revision 0aeaa43dab555891d1278a3931463414b7afb497)
+@@ -950,6 +950,16 @@
+ set_define("ANDROID", target_is_android)
+ 
+ 
++@depends(target)
++def target_is_ohos(target):
++    log.info(f"target {target}")
++    if target.raw_os.endswith("-ohos"):
```
Taym95, line 11:
```suggestion
+    return target.raw_os.endswith("-ohos"):
```
?

mozjs-sys/etc/patches/0034-build__Add_compile_definition_for_ohos.patch
```diff
@@ -0,0 +1,20 @@
+diff --git a/build/moz.configure/init.configure b/build/moz.configure/init.configure
+--- a/build/moz.configure/init.configure	(revision 3720f6d208b45fb968961a8d5c97f64010ae3d81)
++++ b/build/moz.configure/init.configure	(revision 0aeaa43dab555891d1278a3931463414b7afb497)
+@@ -950,6 +950,16 @@
+ set_define("ANDROID", target_is_android)
+ 
+ 
++@depends(target)
++def target_is_ohos(target):
++    log.info(f"target {target}")
```
Taym95, line 10: log still needed?

jschwe: No. Thanks for catching :).

sagudev (changes requested):
I think we need to bump mozjs-sys version for this and it would be nice to add ohos in CI (like it is in servo), so we can land with more confidence that it compiles (ideally in separate PR).

jschwe:
> it would be nice to add ohos in CI 

Sure, I'll open a separate PR for that as suggested.

sagudev (approved):

PR #496, opened 2024-08-17
Check for MOZJS_ARCHIVE before Building from Source

Redfire75369:
If the user specifies `MOZJS_ARCHIVE`, that should be used no matter what, instead of checking for activated features.
The current system does not allow me to create an archive for testing locally with `debugmozjs` and use it, since it checks for that and returns `true`.

I can add some output if you want, but I wasn't sure about what to add.


sagudev (approved):
You are absolutely right.

PR #497, opened 2024-08-17
Add `is_constructing` method to `JS::CallArgs`

saito828koki:
This PR is for fixing the issue in servo repo https://github.com/servo/servo/issues/29821.

We implemented a util function named [`callargs_is_constructing`](https://github.com/servo/servo/blob/20273b062af969152635306c3df2a3a1364ac4d1/components/script/dom/bindings/utils.rs#L629) in servo repo, but having an equivalent method in `JS::CallArgs` and using it should be clearer.

In this PR, a new method `is_constructing` is added to `JS::CallArgs`. It will be used in servo repo to replace all the usage of `callargs_is_constructing` function.

### DIFF
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -304,8 +304,8 @@ impl JS::CallArgs {
     }
 
     #[inline]
-    pub unsafe fn is_constructing(&self) -> bool {
-        (*self.argv_.offset(-1)).is_magic()
+    pub fn is_constructing(&self) -> bool {
+        unsafe { (*self.argv_.offset(-1)).is_magic() }
     }
 }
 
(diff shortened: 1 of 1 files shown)


jdm (approved):

mozjs-sys/src/jsimpls.rs
```diff
@@ -302,6 +302,11 @@ impl JS::CallArgs {
             JS::MutableHandleValue::from_marked_location(self.argv_.offset(self.argc_ as isize))
         }
     }
+
+    #[inline]
+    pub unsafe fn is_constructing(&self) -> bool {
```
jdm, line 307: Let's use an unsafe block instead of marking the method as unsafe.

saito828koki: Thanks for the review. I will fix it.

saito828koki: Fixed. Could you take a look again when you have time, please?

jdm (approved):

PR #501, opened 2024-09-09
Add prebuilt artifacts for OpenHarmony and test github attestions

jschwe:
Adding prebuilt artifacts should fix the OpenHarmony build on windows hosts, since it avoids needing to build spidermonkey from source (the spidermonkey build system seems to assume that on windows hosts, one will only cross-compile to other windows targets).

Additionally we can use this as a test for the github attestations api. 
It requires splitting the build and upload artifact into 2 different jobs, since the job that creates the artifact needs elevated permissions.

Write permissions can't and shouldn't be granted to forks, so I made the `ohos-release` job conditional.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -153,7 +153,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        target: [ 'aarch64-unknown-linux-ohos' ]
+        target: [ 'aarch64-unknown-linux-ohos', 'x86_64-unknown-linux-ohos' ]
     steps:
       - uses: actions/checkout@v4
       - name: Setup OpenHarmony SDK
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -171,6 +171,43 @@ jobs:
         run: |
           ./ohos-build cargo build --target="${{ matrix.target }}"
 
+  ohos-release:
+    name: "OpenHarmony release artifact"
+    runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/main' && github.event_name == 'push' }}
+    permissions:
+      id-token: write
+      contents: read
+      attestations: write
+    strategy:
+      matrix:
+        target: [ 'aarch64-unknown-linux-ohos', 'x86_64-unknown-linux-ohos' ]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup OpenHarmony SDK
+        id: setup_sdk
+        uses: openharmony-rs/setup-ohos-sdk@v0.1
+        with:
+          version: "4.1"
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+      - name: Build (arch ${{ matrix.target }} )
+        env:
+          OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+        run: |
+          ./ohos-build cargo build --release --target="${{ matrix.target }}"
+      - name: Generate artifact attestation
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -171,6 +171,43 @@ jobs:
...
+          targets: ${{ matrix.target }}
+      - name: Build (arch ${{ matrix.target }} )
+        env:
+          OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+        run: |
+          ./ohos-build cargo build --release --target="${{ matrix.target }}"
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-path: './target/libmozjs-x86_64-unknown-linux-ohos.tar.gz'
```
mukilan, line 204: This would need to be derived using `${{ matrix.target }}`?

.github/workflows/rust.yml
```diff
@@ -171,6 +171,43 @@ jobs:
...
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./target/libmozjs-x86_64-unknown-linux-ohos.tar.gz
```
mukilan, line 208: Same as above - this would need to be derived using `${{ matrix.target }}`?

.github/workflows/rust.yml
jschwe: Thanks for catching that!

.github/workflows/rust.yml
```diff
@@ -171,6 +171,43 @@ jobs:
...
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+      - name: Build (arch ${{ matrix.target }} )
+        env:
+          OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+        run: |
+          ./ohos-build cargo build --release --target="${{ matrix.target }}"
+#      - name: Generate artifact attestation
```
jschwe, line 201: Could someone with permissions comment this block out and push the result to a branch in the main repo, so we can test that the attestion works as expected before merging?

sagudev:
> Write permissions can't and shouldn't be granted to forks, so I made the ohos-release job conditional - However that might cause problems with the publish-release / build-result jobs, since those seem to be unconditional.

Maybe we should move attestation to publish-release? (and make it conditional to only happen on push).

Also you can do tests of this PR by pushing to the main of your own fork (if workflows are enabled there).

jschwe:
> Maybe we should move attestation to publish-release? (and make it conditional to only happen on push).

I would prefer to attest that the artifact is the output of `cargo build`. Currently the publish-release command downloads the already built artifact, so the attestation at this point is less meaningful.

> Also you can do tests of this PR by pushing to the main of your own fork (if workflows are enabled there).

Thanks, I did just that and the attestation seems to work just fine: https://github.com/jschwe/mozjs/actions/runs/10776709384/job/29884117507

mukilan (approved):

jschwe:
I went ahead and merged this branch into my forks main branch - CI passed [on the new main branch](https://github.com/jschwe/mozjs/actions/runs/10787159984/job/29916180888) and CI also passes [when opening a new PR against main](https://github.com/jschwe/mozjs/actions/runs/10787789956/job/29918001320?pr=3)

jschwe:
hmm. this PR seems to be stuck waiting on the DCO bot - did I break that somehow or is it just a spurios failure?

sagudev:
> hmm. this PR seems to be stuck waiting on the DCO bot - did I break that somehow or is it just a spurios failure?

Ah, we need https://github.com/servo/servo/pull/33290 and https://github.com/servo/servo/pull/33308#issuecomment-2327919183

PR #503, opened 2024-09-10
Use github artifact attestations for the prebuilt artifacts

jschwe:
I tested this on my local fork (changed line 908 to point to `jschwe/mozjs`.
On my work machine the attestation takes around 500ms, but network is generally slow (downloading the 16MB artifact via curl takes somewhere between 20 and 60s for some reason)

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -49,3 +49,38 @@ jobs:
         run: |
           cargo build --verbose --features streams
           cargo test --tests --examples --verbose --features streams
+
+  verify-archive-ohos:
+    name: "Verify archive OpenHarmony"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: ["aarch64-unknown-linux-ohos", "x86_64-unknown-linux-ohos"]
+    env:
+      RELEASE_TAG: ${{ github.event_name == 'release' && github.ref_name ||  inputs.release-tag }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup OpenHarmony SDK
+        id: setup_sdk
+        uses: openharmony-rs/setup-ohos-sdk@v0.1
+        with:
+          version: "4.1"
+      - name: Download prebuilt mozjs from artifact
+        if: ${{ env.RELEASE_TAG == '' }}
+        uses: actions/download-artifact@v4
+        with:
+          name: libmozjs-${{ matrix.platform.target }}.tar.gz
+      - name: Build from archive
+        if: ${{ env.RELEASE_TAG == '' }}
+        env:
+          OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+          MOZJS_ARCHIVE: libmozjs-${{ matrix.platform.target }}.tar.gz
+        run: |
+          ./ohos-build cargo build --target="${{ matrix.target }}" --verbose --features streams
+      - name: Build from auto-download (arch ${{ matrix.target }})
+        if: ${{ env.RELEASE_TAG != '' }}
+        env:
+          OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
+          MOZJS_ATTESTATION: strict
+        run: |
+          ./ohos-build cargo build --target="${{ matrix.target }}" --verbose --features "streams"
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -44,6 +44,8 @@ jobs:
           cargo test --tests --examples --verbose --features streams
       - name: Build from auto-download
         if: ${{ env.RELEASE_TAG != '' }}
+        env:
+          MOZJS_ATTESTATION: strict
         run: |
           cargo build --verbose --features streams
           cargo test --tests --examples --verbose --features streams
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -65,16 +65,20 @@ jobs:
         uses: openharmony-rs/setup-ohos-sdk@v0.1
         with:
           version: "4.1"
+      - name: Install Rust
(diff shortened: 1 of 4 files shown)


jschwe:
I would love to see this merged, simply since it would allow servo builds for OH to work on windows. If there are objections to using `gh` to verify the OH shared library, then I can also remove that for now.

jschwe:
true, I guess the issue is that we never cut a release after #501. Should i open a seperate PR which just bumps the version?

sagudev:
Ah we didn't bump mozjs-sys ver in https://github.com/servo/mozjs/pull/501, so that's why there is no ohos release. So what this PR actually does is use attestation for ohos artifact. If we decide to introduce attestation, we should introduce it for all targets.

jschwe:
>  If we decide to introduce attestation, we should introduce it for all targets.

I remember someone on zulip had reservations about potential influence on runtime of the build-script, so I thought perhaps we should first get some real-world experience on one target, before affecting the commonly used ones

sagudev:
> I remember someone on zulip had reservations about potential influence on runtime of the build-script, so I thought perhaps we should first get some real-world experience on one target, before affecting the commonly used ones

That was me :). I think we should make make it for all platforms but let it be disabled by default (enabled via `MOZJS_ATTESTATION`), this way we can test it reasonably faster and then decide.

jschwe:
> MOZJS_ATTESTATION

How fancy do we want to make this switch? I guess having an option like `MOZJS_ATTESTATION=force`, which causes the build-script to exit with an error if `gh` is not found would be nice. I personally would say we could just differentiate between the variants "force", "variable not set" and "set to something !=force" and wouldn't bother with checking for values like false, off, no, 0 etc. What do you think?

sagudev:
> > MOZJS_ATTESTATION
> 
> How fancy do we want to make this switch? I guess having an option like `MOZJS_ATTESTATION=force`, which causes the build-script to exit with an error if `gh` is not found would be nice. I personally would say we could just differentiate between the variants "force", "variable not set" and "set to something !=force" and wouldn't bother with checking for values like false, off, no, 0 etc. What do you think?

Sounds good.

jschwe:
Is it possible we need to update the upstream mozjs COMMIT again? seems like the artifact download is failing.

sagudev:
> Is it possible we need to update the upstream mozjs COMMIT again? seems like the artifact download is failing.

We need to update SM to new minor release. I will do it later today.

.github/workflows/release-check.yml
```diff
@@ -14,6 +14,8 @@ env:
   RUST_BACKTRACE: 1
   SHELL: /bin/bash
   CARGO_INCREMENTAL: 0
+  # Force attestation verification and fail build if verification fails.
+  MOZJS_ATTESTATION: forceforce
```
sagudev, line 18: typo?

jschwe: No. The name is ugly and subject to bikeshedding. Below a copy of the comment in the code explaining the available values.

>  ```
>      // MOZJS_ATTESTATION values:
>      // unset -> no verification
>      // force -> force verification and fail if `gh` is not installed or recent enough
>      // set to any other value -> enable verification if `gh` installed and new enough.
>      // Note: If verification is enabled and fails, then we fall back to compiling from source.
>      // forceforce -> As a debug option we provide forceforce, which fails the build instead
>      //               of falling back to compiling from source if verification fails

The `force` option exists to force verification of the artifact. Since we can't assume `gh` is installed (or new enough, e.g. fedora packages a quite outdated version), the default is to just accept the downloaded artifact, if `gh` is not available.
When artifact verification is possible AND fails, then I think it makes sense to fallback to compiling from source instead of failing the build. 
For debugging purposes, we do actually want the build to fail, otherwise it's hard to tell if the artififact verification is working as expected (hence the more strict `forceforce` option, which I would consider an internal debug option).

mukilan: maybe we can use unset/lenient/strict as the possible values.

jschwe: Do you also have a suggestion for the fourth option? ExtraStrict?

mukilan: Do we need four modes? 
unset -> no verification if not possible
lenient -> verify if possible, but fallback to source if not possible or fails verification
strict -> always verify

jschwe: >  unset -> no verification if not possible

I would be fine with that. In that case we could add an explicit  `off` switch, to disable verification, even if possible.

sagudev:
I will review this at the end of the week.

mozjs-sys/build.rs
```diff
@@ -902,6 +905,72 @@ fn decompress_static_lib(archive: &Path, build_dir: &Path) -> Result<(), std::io
...
+                );
+            }
+        }
+        // When the environment variable is not set or invalid,
+        // we enable attestation if possible.
+        if *ATTESTATION_AVAILABLE {
+            ArtifactAttestation::Enabled(AttestationType::Lenient)
+        } else {
+            ArtifactAttestation::Disabled
+        }
```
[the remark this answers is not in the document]
jschwe: We did, yes. But then [there was some discussion](https://github.com/servo/mozjs/pull/503#discussion_r1785727512) on the name of the fourth option, and Mukilan suggested that we don't need that last option if we enable by default when possible.
I'm fine either way, but that does return us to the original discussion on the naming of the options. Perhaps it's best to just not name the last option and only give it a number?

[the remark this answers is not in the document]
sagudev: I think we can continue to have only 3 options as we also handle `ATTESTATION_AVAILABLE` down. Just change default from current to 0/1 to always be 0 (just replace if-else with `ArtifactAttestation::Disabled`).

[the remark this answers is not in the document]
jschwe: done

mrobinson (approved):

mrobinson:
Seems good to me. Let's give this a shot.

PR #506, opened 2024-09-22
Reconfigure if configure inputs changed

### LINKED ISSUE #505: Force Reconfigure if CC / CXX changed
I noticed that when building servo for openharmony and upgrading the SDK, the mozjs build system will continue to use the old compiler paths. This becomes noticable, when deleting the old SDK, since that triggers errors like: 

```
make[4]: /path/to/old/llvm/bin/aarch64-unknown-linux-ohos-clang++: No such file or directory
```

I'm a bit more worried about the case, were the user did not delete the old SDK, and mozjs is just silently using the old compiler instead of the new compiler that servo is telling it to use.

Proposed solution: mozjs should force a complete reconfigure if variables like CC or CXX change.

jschwe:
Save all input variables to the configure script, so that we can re-run it when necessary.

Closes #505

### DIFF
--- a/mozjs-sys/makefile.cargo
+++ b/mozjs-sys/makefile.cargo
@@ -148,12 +148,18 @@ all: maybe-configure
 # The second two time checks handle the case of configure.in and configure having
 # the same timestamp (e.g. after a git checkout)
 JSSRC := '$(SRC_DIR)'/js/src
+# Keep track of all input variables to configure, so we don't skip reconfigure
+# when we do need to reconfigure
+CONFIGURE_INPUTS := "$(CC)$(CFLAGS)$(CPP)$(CPPFLAGS)$(CXX)$(CXXFLAGS)$(AS)$(AR)$(CONFIGURE_FLAGS)"
+LAST_CONFIGURE_INPUTS := "$(shell cat reconfigure.inputs)"
 maybe-configure:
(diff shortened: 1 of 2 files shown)


mozjs-sys/makefile.cargo
```diff
@@ -148,12 +148,18 @@ all: maybe-configure
 # The second two time checks handle the case of configure.in and configure having
 # the same timestamp (e.g. after a git checkout)
 JSSRC := '$(SRC_DIR)'/js/src
+# Keep track of all input variables to configure, so we don't skip reconfigure
+# when we do need to reconfigure
+CONFIGURE_INPUTS := "$(CC)$(CFLAGS)$(CPP)$(CPPFLAGS)$(CXX)$(CXXFLAGS)$(AS)$(AR)$(CONFIGURE_FLAGS)"
+LAST_CONFIGURE_INPUTS := "$(shell cat reconfigure.inputs)"
```
sagudev, line 154: what is shell command? It's not available in git bash.

jschwe: isn't that make syntax to call the shell (bash)? Just doing `$(cat reconfigure.inputs)` did not work for me.

sagudev: right, this is makefile not bash script!

mrobinson (approved):

PR #509, opened 2024-09-24
Respect "STRIP" env variable for prebuilt artifact 

### LINKED ISSUE #508: Investigate ohos release artifact size
the OH artifact is 118MB large when zipped, 600MB uncompressed. 
It's likely that the strip command is not working correctly. 
We could read `STRIP` as an environment variable, allowing the user to set `llvm-strip`, or perhaps we could use `${CC}/../llvm-strip`  if `CC.endswith("clang")`.

jschwe:
Allow the user to select the strip command via setting the STRIP environment variable (or the cc-rs variants of it).
For the OpenHarmony build the `ohos-build` script already sets `STRIP`, so this change also reduces the artifact size of the OpenHarmony release artifact.

Closes #508

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.0-10"
+version = "0.128.0-11"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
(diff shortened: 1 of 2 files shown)


jschwe:
Tested, by merging into my fork https://github.com/jschwe/mozjs/releases/tag/mozjs-sys-v0.128.0-11
(The commit merged into my fork is slightly different - it just read `STRIP`, and not any of the cc-rs variants, but that shouldn't make any difference)

sagudev (approved):

PR #510, opened 2024-09-29
Generate attestation for all artifacts

sagudev:
https://github.com/servo/mozjs/pull/501 caused that release checks where not run on PRs anymore, this PR fixes it.

cc @jschwe


sagudev:
this will still create attestation for unreleased artifacts!

.github/workflows/build.yml
```diff
@@ -200,8 +189,9 @@ jobs:
           ./ohos-build cargo build --release --target="${{ matrix.target }}"
```
jschwe, line 189: While reviewing, I noticed that the OpenHarmony archive should also enable the `streams` feature, since the build-script seems to expect the prebuild archive to have the streams feature enabled. Should I open a separate PR for this, or could you perhaps add this to this PR?

jschwe (approved):
Thanks for adding the attestation for all the other platforms! 
Looks good to me, just have some minor questions.

> https://github.com/servo/mozjs/pull/501 caused that release checks where not run on PRs anymore, this PR fixes it.

Just for future reference - what exactly caused the release checks to not run anymore? I looked through both 501 and this PR, but nothing in particular caught my eye,

> this will still create attestation for unreleased artifacts, but so was https://github.com/servo/mozjs/pull/501

Just to clarify for other readers: the attestation is generated for artifacts created in CI for pushes to `main`. In practice, this is quite close to "tagged releases", since we release a new version for almost every new commit to main.

sagudev:
> > #501 caused that release checks where not run on PRs anymore, this PR fixes it.
> 
> Just for future reference - what exactly caused the release checks to not run anymore? I looked through both 501 and this PR, but nothing in particular caught my eye,

I also needed some time to figure it out, but `ohos-release` is only run on pushes and [`publish-release` `needs` `ohos-release`](https://github.com/servo/mozjs/blob/446ca9765ac3fe582147fd57a51d9a3bcea676ec/.github/workflows/rust.yml#L260) hence `publish-release` is also only run on pushes.

.github/workflows/build.yml
sagudev: Let's add it here and bump mozjs-sys version (we need it anyway to get attestation on release).

mukilan (approved):

PR #512, opened 2024-10-09
Add `ThreadSafeJSContext`

sagudev:
https://github.com/servo/servo/issues/33759

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -55,6 +55,7 @@ use crate::jsapi::{JSString, Object, PersistentRootedIdVector};
 use crate::jsapi::{JS_DefineFunctions, JS_DefineProperties, JS_DestroyContext, JS_ShutDown};
 use crate::jsapi::{JS_EnumerateStandardClasses, JS_GetRuntime, JS_GlobalObjectTraceHook};
 use crate::jsapi::{JS_MayResolveStandardClass, JS_NewContext, JS_ResolveStandardClass};
+use crate::jsapi::{JS_RequestInterruptCallback, JS_RequestInterruptCallbackCanWait};
 use crate::jsapi::{JS_StackCapture_AllFrames, JS_StackCapture_MaxFrames};
 use crate::jsapi::{PersistentRootedObjectVector, ReadOnlyCompileOptions, RootingContext};
 use crate::jsapi::{SetWarningReporter, SourceText, ToBooleanSlow};
@@ -293,6 +294,8 @@ pub struct Runtime {
     /// to represent the resulting ownership graph and risk destroying a Runtime on
     /// the wrong thread.
     outstanding_children: Arc<()>,
+    /// Raw pointer used by thread safe SyncJSContext
+    mt_cx: Arc<Mutex<Option<*mut JSContext>>>,
 }
 
 impl Runtime {
@@ -301,6 +304,11 @@ impl Runtime {
         let cx = CONTEXT.with(|context| context.get());
         assert!(!cx.is_null());
         cx
+    }
+
+    /// Returns thread safe SyncJSContext
+    pub fn get_sync(&self) -> SyncJSContext {
+        SyncJSContext(self.mt_cx.clone())
     }
 
     /// Creates a new `JSContext`.
@@ -370,6 +378,7 @@ impl Runtime {
             _parent_child_count: parent.map(|p| p.children_of_parent),
             cx: js_context,
             outstanding_children: Arc::new(()),
+            mt_cx: Arc::new(Mutex::new(Some(js_context))),
         }
     }
 
@@ -417,6 +426,7 @@ impl Runtime {
 
 impl Drop for Runtime {
     fn drop(&mut self) {
+        *self.mt_cx.lock().unwrap() = None;
         assert_eq!(
             Arc::strong_count(&self.outstanding_children),
             1,
@@ -429,6 +439,35 @@ impl Drop for Runtime {
                 assert_eq!(context.get(), self.cx);
                 context.set(ptr::null_mut());
             });
+        }
+    }
+}
+
+/// JSContext that is Send and Sync
+/// but supports only limited (thread safe) operations
+pub struct SyncJSContext(Arc<Mutex<Option<*mut JSContext>>>);
+
+unsafe impl Send for SyncJSContext {}
+unsafe impl Sync for SyncJSContext {}
+
+// We expose only methods that are defined as thread safe in spidermonkey
+// https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/JSContext.h#848
+impl SyncJSContext {
+    /// JS_RequestInterruptCallback
+    pub fn request_interrupt_callback(&self) {
+        if let Some(&cx) = self.0.lock().unwrap().as_ref() {
+            unsafe {
+                JS_RequestInterruptCallback(cx);
+            }
+        }
+    }
+
+    /// JS_RequestInterruptCallbackCanWait
+    pub fn request_interrupt_callback_can_wait(&self) {
+        if let Some(&cx) = self.0.lock().unwrap().as_ref() {
+            unsafe {
+                JS_RequestInterruptCallbackCanWait(cx);
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
     }
 }
 
+/// JSContext that is Send and Sync
+/// but supports only limited (thread safe) operations
+pub struct SyncJSContext(Arc<Mutex<Option<*mut JSContext>>>);
```
mrobinson: I like the name using "ThreadSafe" a bit better here as it matches the terminology we use for DOM nodes in layout. `Sync` maybe leans to heavily on the sense of "locking."
 
```suggestion
/// A version of the [`JSContext`] that can be used from other threads and is thus
/// `Send` and `Sync`. This should only ever expose operations that are marked as
/// thread-safe by the SpiderMonkey API.
/// See <https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/JSContext.h#848>.
pub(crate) struct ThreadSafeJSContext(Arc<Mutex<Option<*mut JSContext>>>);
```

mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
...
+
+unsafe impl Send for SyncJSContext {}
+unsafe impl Sync for SyncJSContext {}
+
+// We expose only methods that are defined as thread safe in spidermonkey
+// https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/JSContext.h#848
```
mrobinson, line 455: I think it makes sense to remove this comment in exchange for the rustdoc above.

mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
...
+impl SyncJSContext {
+    /// JS_RequestInterruptCallback
```
mrobinson, line 457:
```suggestion
    /// Call`JS_RequestInterruptCallback` from the SpiderMonkey API. This is thread-safe according to
    /// <link-to-documentation>. 
```

mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
...
+    pub fn request_interrupt_callback(&self) {
+        if let Some(&cx) = self.0.lock().unwrap().as_ref() {
+            unsafe {
+                JS_RequestInterruptCallback(cx);
+            }
+        }
+    }
+
+    /// JS_RequestInterruptCallbackCanWait
```
mrobinson, line 466:
```suggestion
    ///  Call`JS_RequestInterruptCallbackCanWait` from the SpiderMonkey API. This is thread-safe according to
    /// <link-to-documentation>. 
```

mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
...
+pub struct SyncJSContext(Arc<Mutex<Option<*mut JSContext>>>);
+
+unsafe impl Send for SyncJSContext {}
+unsafe impl Sync for SyncJSContext {}
+
+// We expose only methods that are defined as thread safe in spidermonkey
+// https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/src/vm/JSContext.h#848
+impl SyncJSContext {
+    /// JS_RequestInterruptCallback
+    pub fn request_interrupt_callback(&self) {
```
mrobinson, line 458:
```suggestion
    pub(crate) fn request_interrupt_callback(&self) {
```

mozjs/src/rust.rs
```diff
@@ -433,6 +443,35 @@ impl Drop for Runtime {
...
+        if let Some(&cx) = self.0.lock().unwrap().as_ref() {
+            unsafe {
+                JS_RequestInterruptCallback(cx);
+            }
+        }
+    }
+
+    /// JS_RequestInterruptCallbackCanWait
+    pub fn request_interrupt_callback_can_wait(&self) {
```
mrobinson, line 467:
```suggestion
    pub(crate) fn request_interrupt_callback_can_wait(&self) {
```

mrobinson (changes requested):
This looks good to me. I've have some picky documentation and naming requests below. I'm being a bit pickier about this documentation since it is such and important and delicate concept in Servo and there was a bit of confusion about this before.

mozjs/src/rust.rs
sagudev: We need to be able to use this struct in servo so it needs to be pub.

mrobinson: Oh sorry! That makes sense. I thought this change was in Servo itself. Too many repositories.

mozjs/src/rust.rs
```diff
@@ -417,6 +426,7 @@ impl Runtime {
 
 impl Drop for Runtime {
     fn drop(&mut self) {
+        *self.mt_cx.lock().unwrap() = None;
```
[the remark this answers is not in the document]
sagudev: I will change it to rwlock but will keep using std impl (because it would require new dep and this should not happen frequently).

mozjs/src/rust.rs
```diff
@@ -417,6 +428,7 @@ impl Runtime {
 
 impl Drop for Runtime {
     fn drop(&mut self) {
+        *self.thread_safe_handle.write().unwrap() = None;
```
mukilan, line 431:
```suggestion
        self.thread_safe_handle.write().unwrap().take();
```

mozjs/src/rust.rs
```diff
@@ -293,6 +294,10 @@ pub struct Runtime {
     /// to represent the resulting ownership graph and risk destroying a Runtime on
     /// the wrong thread.
     outstanding_children: Arc<()>,
+    /// An `Option` that holds the same pointer as `cx`.
+    /// This is shared with all [`ThreadSafeJSContext`]s, so
+    /// they can detect when its destroyed on the main thread.
```
mrobinson, line 299: Apologies. I made a typo in my suggestion:

```suggestion
    /// they can detect when it's destroyed on the main thread.
```

mozjs/src/rust.rs
```diff
@@ -433,6 +445,40 @@ impl Drop for Runtime {
     }
 }
 
+/// A version of the [`JSContext`] that can be used from other threads and is thus
+/// `Send` and `Sync`. This should only ever expose operations that are marked as
+/// thread-safe by the SpiderMonkey API.
+// (only use atomic fields in JSContext)
```
mrobinson, line 451:
```suggestion
/// thread-safe by the SpiderMonkey API, ie ones that only atomic fields in JSContext.
```

mozjs/src/rust.rs
```diff
@@ -433,6 +445,40 @@ impl Drop for Runtime {
...
+#[derive(Clone)]
+pub struct ThreadSafeJSContext(Arc<RwLock<Option<*mut JSContext>>>);
+
+unsafe impl Send for ThreadSafeJSContext {}
+unsafe impl Sync for ThreadSafeJSContext {}
+
+impl ThreadSafeJSContext {
+    /// Call`JS_RequestInterruptCallback` from the SpiderMonkey API.
```
mrobinson, line 459:
```suggestion
    /// Call `JS_RequestInterruptCallback` from the SpiderMonkey API.
```

mozjs/src/rust.rs
```diff
@@ -433,6 +445,40 @@ impl Drop for Runtime {
...
+    /// <https://searchfox.org/mozilla-central/rev/7a85a111b5f42cdc07f438e36f9597c4c6dc1d48/js/public/Interrupt.h#19>
+    pub fn request_interrupt_callback(&self) {
+        if let Some(&cx) = self.0.read().unwrap().as_ref() {
+            unsafe {
+                JS_RequestInterruptCallback(cx);
+            }
+        }
+    }
+
+    /// Call`JS_RequestInterruptCallbackCanWait` from the SpiderMonkey API.
```
mrobinson, line 470:
```suggestion
    /// Call `JS_RequestInterruptCallbackCanWait` from the SpiderMonkey API.
```

mrobinson (approved):

mrobinson:
@sagudev Great work! With the new design and documentation is now *much* clearer what is going on here.

sagudev:
> @sagudev Great work! With the new design and documentation is now _much_ clearer what is going on here.

It took me a while to figure out what is going, so it only make sense to improve and document findings, or else I would forget it.

PR #514, opened 2024-10-26
Fix unsound use of Rooted<T> with RootKind::Traceable

jdm:
The existing implementation for `Rooted<T>` is correct for most JS types that are rooted, but there's one significant case that is broken. `Rooted<T>` uses C++ template magic to change its base class depending on the type that it's templated over, and certain types that map to `RootKind::Traceable` [end up with a vtable](https://searchfox.org/mozilla-central/source/js/public/RootingAPI.h#923) that contains a custom tracing implementation, while other types do not have any such vtable.

These template shenanigans are very likely the reason why we have a handwritten `Rooted<T>` definition, since bindgen is not able to handle this situation. This PR makes the following changes:
* we introduce a vtable field into `Rooted<T>` which is zero-sized for types that do not have Traceable RootKind value
* we introduce a `TraceableTrace` trait that `Rooted<T>`'s vtable delegates to, allowing implementers to write straightforward tracing implementations for their types

These changes were verified by running the property_descriptor unit test with zealous GC enabled (ie. perform a GC every time there's an allocation) and the test no longer crashes after these changes are applied.

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.3-0"
+version = "0.128.3-1"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
--- a/mozjs-sys/src/jsgc.rs
+++ b/mozjs-sys/src/jsgc.rs
@@ -68,7 +68,7 @@ impl RootKind for JS::Value {
     const KIND: JS::RootKind = JS::RootKind::Value;
 }
 
-impl RootKind for JS::PropertyDescriptor {
+impl<T: TraceableTrace> RootKind for T {
     type Vtable = *const RootedVFTable;
     const VTABLE: Self::Vtable = &<Self as TraceableTrace>::VTABLE;
     const KIND: JS::RootKind = JS::RootKind::Traceable;
@@ -96,7 +96,7 @@ impl RootedVFTable {
 /// `Rooted<T>` with a T that uses the Traceable RootKind uses dynamic dispatch on the C++ side
 /// for custom tracing. This trait provides trace logic via a vtable when creating a Rust instance
 /// of the object.
-pub unsafe trait TraceableTrace: Sized + RootKind {
+pub unsafe trait TraceableTrace: Sized {
     const VTABLE: RootedVFTable = RootedVFTable {
         padding: RootedVFTable::PADDING,
         trace: Self::trace,
@@ -265,12 +265,6 @@ impl<const N: usize> ValueArray<N> {
     pub unsafe fn get_mut_ptr(&self) -> *mut JS::Value {
         self.elements.as_ptr() as *mut _
     }
-}
-
-impl<const N: usize> RootKind for ValueArray<N> {
-    type Vtable = *const RootedVFTable;
-    const VTABLE: Self::Vtable = &<Self as TraceableTrace>::VTABLE;
-    const KIND: JS::RootKind = JS::RootKind::Traceable;
 }
 
 unsafe impl<const N: usize> TraceableTrace for ValueArray<N> {
--- a/mozjs-sys/src/jsgc.rs
+++ b/mozjs-sys/src/jsgc.rs
@@ -2,91 +2,145 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use crate::glue::CallPropertyDescriptorTracer;
+use crate::jsapi::js::TraceValueArray;
 use crate::jsapi::JS;
 use crate::jsapi::{jsid, JSFunction, JSObject, JSScript, JSString, JSTracer};
 
 use crate::jsid::VoidId;
 use std::cell::UnsafeCell;
-use std::ffi::c_void;
+use std::ffi::{c_char, c_void};
 use std::mem;
 use std::ptr;
 
 /// A trait for JS types that can be registered as roots.
 pub trait RootKind {
-    #[allow(non_snake_case)]
-    /// Returns the rooting kind for `Self`.
-    fn rootKind() -> JS::RootKind;
+    type Vtable;
+    const VTABLE: Self::Vtable;
+    const KIND: JS::RootKind;
 }
 
 impl RootKind for *mut JSObject {
-    #[inline(always)]
-    fn rootKind() -> JS::RootKind {
-        JS::RootKind::Object
-    }
+    type Vtable = ();
+    const VTABLE: Self::Vtable = ();
+    const KIND: JS::RootKind = JS::RootKind::Object;
 }
 
 impl RootKind for *mut JSFunction {
-    #[inline(always)]
-    fn rootKind() -> JS::RootKind {
-        JS::RootKind::Object
-    }
+    type Vtable = ();
+    const VTABLE: Self::Vtable = ();
+    const KIND: JS::RootKind = JS::RootKind::Object;
 }
 
 impl RootKind for *mut JSString {
-    #[inline(always)]
-    fn rootKind() -> JS::RootKind {
-        JS::RootKind::String
-    }
+    type Vtable = ();
+    const VTABLE: Self::Vtable = ();
+    const KIND: JS::RootKind = JS::RootKind::String;
 }
 
 impl RootKind for *mut JS::Symbol {
-    #[inline(always)]
-    fn rootKind() -> JS::RootKind {
-        JS::RootKind::Symbol
-    }
+    type Vtable = ();
+    const VTABLE: Self::Vtable = ();
+    const KIND: JS::RootKind = JS::RootKind::Symbol;
 }
 
 impl RootKind for *mut JS::BigInt {
-    #[inline(always)]
-    fn rootKind() -> JS::RootKind {
-        JS::RootKind::BigInt
-    }
+    type Vtable = ();
+    const VTABLE: Self::Vtable = ();
+    const KIND: JS::RootKind = JS::RootKind::BigInt;
 }
 
 impl RootKind for *mut JSScript {
-    #[inline(always)]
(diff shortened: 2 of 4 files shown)


mozjs-sys/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.3-0"
+version = "0.128.4-0"
```
jschwe, line 5:
```suggestion
version = "0.128.3-1"
```

I believe the version before the dash is meant to correspond with the upstream LTS version.

sagudev: That's right.

mozjs-sys/src/jsimpls.rs
```diff
@@ -386,30 +387,66 @@ impl JSNativeWrapper {
...
+    pub unsafe fn add_to_root_stack(&mut self, cx: *mut JSContext)
+    where
+        T: RootKind,
+    {
+        let kind = T::rootKind();
+        // Any stack root that uses the Traceable kind needs to use RootedTraceable instead.
+        // Using Traceable with Rooted<T> leads to crashes during GCs.
+        assert_ne!(kind, JS::RootKind::Traceable);
+        self.base.add_to_root_stack(cx, kind)
+    }
```
sagudev, line 407: I am not really thrilled with this runtime check, I think I managed to put this in compiletime by merging Rooted and RootedTraceable back into one type: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cc1af6037a0b3c6373e12116585d9c28

jdm: Oooh, that's really interesting. I'll see if I can integrate that!

jdm: Grumble. The data representation isn't quite correct when view from the C++ code in lldb, and I can't figure out why yet. The vtable pointer is being treated as the value of the stack pointer.

sagudev: do you have proper repr attribute everywhere (even on vtable type)?

jdm: https://github.com/servo/mozjs/compare/main...jdm:mozjs:propdesc2?expand=1

sagudev: Maybe try without TraceableTrace hop:
```
mod prop_desc {
    const vftable: RootedVFTable = RootedVFTable {
        padding: RootedVFTable::PADDING,
        trace,
    };

    unsafe extern "C" fn trace(this: *mut c_void, trc: *mut JSTracer, _name: *const c_char) {
        let rooted = this as *mut RootedTraceable<Self>;
        let rooted = rooted.as_mut().unwrap();
        CallPropertyDescriptorTracer(trc, &mut rooted.ptr);
    }
}

impl RootKind for JS::PropertyDescriptor {
    type Vtable = *const RootedVFTable;
    const VTABLE: Self::Vtable = &prop_desc::vftable;
    const KIND: JS::RootKind = JS::RootKind::Traceable;
}
```

sagudev: > The vtable pointer is being treated as the value of the stack pointer.

This could also be coming from using & instead of addr_of!(), another potential problem could be using `const` instead of `static` (I think consts are most of the time inlined).

jdm: Sadly none of those ideas made a difference.

jdm: Oh, I just realized what the difference in code is—in the first branch, I moved add_to_root_stack/remove_from_root_stack to RootedBase, which is after the vtable field, but it used self as `Rooted<T>`. That seems extremely sketchy, but it apparently worked :<

jdm: Yeah, https://github.com/servo/mozjs/commit/712f584e0b94674fdb9a157a1b060c39a87058af made the test pass. Not sure how I feel about that!

sagudev: We do not need RootedBase if we only have one rooted type.

sagudev: Ah, `TraceExactStackRootTraceableList` expects `StackRootedBase*` on its trace list and this is then casted to `*StackRootedTraceableBase`, so if we pass ptr to Rooted (that already has vtable inside) we get dummy data for vtable: https://searchfox.org/mozilla-central/rev/c0b40fc662d376ecdee66b5d7489f4be63ef88bc/js/src/gc/RootMarking.cpp#56

sagudev:
```patch
diff --git a/mozjs-sys/src/jsimpls.rs b/mozjs-sys/src/jsimpls.rs
index 3a9efc5f0..ca631cb37 100644
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -412,11 +412,11 @@ impl<T: RootKind> JS::Rooted<T> {
         self.base.stack = stack;
         self.base.prev = *stack;
 
-        *stack = self as *mut _ as usize as _;
+        *stack = ::std::ptr::addr_of_mut!(self.base) as usize as _;
     }
 
     pub unsafe fn remove_from_root_stack(&mut self) {
-        assert!(*self.base.stack == self as *mut _ as usize as _);
+        assert!(*self.base.stack == ::std::ptr::addr_of_mut!(self.base) as usize as _);
         *self.base.stack = self.base.prev;
     }
 }
```
on top of https://github.com/servo/mozjs/commit/712f584e0b94674fdb9a157a1b060c39a87058af also works, but I like RootedBase solution better, but we should write comment in code why is that also inline markings.

mozjs-sys/src/jsgc.rs
```diff
@@ -2,91 +2,145 @@
...
+        let rooted = this as *mut Rooted<Self>;
+        let rooted = rooted.as_mut().unwrap();
+        Self::do_trace(&mut rooted.ptr, trc);
     }
+
+    /// Used by `TraceableTrace` implementer to trace its contents.
+    /// Corresponds to virtual `trace` call in a `Rooted` that inherits from
+    /// StackRootedTraceableBase (C++).
+    fn do_trace(&mut self, trc: *mut JSTracer);
+}
```
sagudev, line 115: I think this trait is redundant as it's not used as bound but rather as namespace. It does provide wrapper for trace, but I think hops just complicates things further. I think this would be clearer:

```rust
#[repr(C)]
pub struct RootedVFTable {
    #[cfg(windows)]
    padding: [usize; 1],
    #[cfg(not(windows))]
    padding: [usize; 2],
    trace: unsafe extern "C" fn(this: *mut c_void, trc: *mut JSTracer, name: *const c_char),
}

impl RootedVFTable {
    const fn new(trace: unsafe extern "C" fn(this: *mut c_void, trc: *mut JSTracer, name: *const c_char)) -> Self {
        Self {
            padding: Self::PADDING,
            trace,
        }
    }
}

    unsafe extern "C" fn prop_desc_trace(this: *mut c_void, trc: *mut JSTracer, _name: *const c_char) {
        let rooted = this as *mut Rooted<Self>;
        let rooted = rooted.as_mut().unwrap();
        CallPropertyDescriptorTracer(trc, &mut rooted.ptr);
    }

const prop_desc_vtable = RootedVFTable::new(prop_desc_trace);

impl RootKind for JS::PropertyDescriptor {
    type Vtable = *const RootedVFTable;
    const VTABLE: Self::Vtable = &prop_desc_vtable;
    const KIND: JS::RootKind = JS::RootKind::Traceable;
}
```

jdm: Yeah, I suppose that makes sense. I tried to use it as a bound like:
```rust
impl<T: TraceableTrace> for T {
    type Vtable = *const RootedVFTable;
    const VTABLE: Self::Vtable = &<Self as TraceableTrace>::VTABLE;
    const KIND: JS::RootKind = JS::RootKind::Traceable;
}
```
but rustc complained that the type of `&<Self as TraceableTrace>::VTABLE` was not the expected associated type.

jdm: Actually, I got this to work in the latest commit and I'd prefer to keep TraceableTrace around for that reason. It makes it impossible to use the wrong vtable definition this way.

mozjs-sys/src/jsgc.rs
```diff
@@ -2,91 +2,145 @@
...
+    unsafe extern "C" fn trace(this: *mut c_void, trc: *mut JSTracer, _name: *const c_char) {
+        let rooted = this as *mut Rooted<Self>;
+        let rooted = rooted.as_mut().unwrap();
+        Self::do_trace(&mut rooted.ptr, trc);
     }
+
+    /// Used by `TraceableTrace` implementer to trace its contents.
+    /// Corresponds to virtual `trace` call in a `Rooted` that inherits from
+    /// StackRootedTraceableBase (C++).
+    fn do_trace(&mut self, trc: *mut JSTracer);
```
sagudev, line 114: I think this should also be marked as unsafe.

sagudev (approved):
LGTM

mozjs-sys/src/jsgc.rs
sagudev: That's perfect!

mozjs-sys/src/jsgc.rs
```diff
@@ -111,14 +111,12 @@ pub unsafe trait TraceableTrace: Sized {
...
 }
 
 unsafe impl TraceableTrace for JS::PropertyDescriptor {
-    fn do_trace(&mut self, trc: *mut JSTracer) {
-        unsafe {
-            CallPropertyDescriptorTracer(trc, self);
-        }
+    unsafe fn do_trace(&mut self, trc: *mut JSTracer) {
+        CallPropertyDescriptorTracer(trc, self);
     }
```
sagudev, line 120: Actually unsafe block in unsafe fn will be the only correct way in the future: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html, but that is not yet relevant.

PR #515, opened 2024-10-28
Use GCZeal in tests when run with debugmozjs

sagudev:
Discovered on work from https://github.com/servo/mozjs/pull/514 (failed before it landed: https://github.com/servo/mozjs/actions/runs/11548177590)


jdm (approved):
Good idea!

PR #519, opened 2024-10-31
Hide call to unsafe method inside of an unsafe block in rooting macro.

jdm:
This usage of initial() is safe because we're storing the result directly inside of a rooted location.

### DIFF
--- a/mozjs/src/gc/macros.rs
+++ b/mozjs/src/gc/macros.rs
@@ -18,10 +18,12 @@ macro_rules! rooted {
     };
 	(in($cx:expr) let $($var:ident)+: $type:ty) => {
         let mut __root = $crate::jsapi::Rooted::new_unrooted();
+        // SAFETY:
+        // We're immediately storing the initial value in a rooted location.
         let $($var)+: $crate::gc::RootedGuard<$type> = $crate::gc::RootedGuard::new(
             $cx,
             &mut __root,
-            <$type as $crate::gc::GCMethods>::initial(),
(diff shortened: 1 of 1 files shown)


sagudev (commented):
this could cause warnings for rooted! inside unsafe blocks (due to unsafe inside unsafe). see https://github.com/rust-lang/rust/issues/74838

jdm:
> this could cause warnings for rooted! inside unsafe blocks (due to unsafe inside unsafe). see [rust-lang/rust#74838](https://github.com/rust-lang/rust/issues/74838)

We can allow that when it happens. It's better than requiring unsafe blocks in code that does not require them, which is a problem I hit when trying to use this version of the macro.

sagudev:
> > this could cause warnings for rooted! inside unsafe blocks (due to unsafe inside unsafe). see [rust-lang/rust#74838](https://github.com/rust-lang/rust/issues/74838)
> 
> We can allow that when it happens. It's better than requiring unsafe blocks in code that does not require them, which is a problem I hit when trying to use this version of the macro.

I have a feeling this happens a lot in servo (if before unsafe was required). So we could simply put `#[allow(unused_unsafe)]` in macro: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=80541d5b4fbe5645e4f9e47d4d496d1b

jdm:
The isn't a version of the macro that we use very often in Servo right now. I did not need to modify any existing code with these changes.

sagudev (approved):

PR #522, opened 2024-11-09
Support Box<T>/Option<T> inside of non-JS value roots.

jdm:
The auto_root! macro creates a CustomAutoRoot value for the provided initializer. When this initializer contains a Box, the code fails to compile because we don't have custom tracing code for boxes.

### DIFF
--- a/mozjs-sys/src/jsgc.rs
+++ b/mozjs-sys/src/jsgc.rs
@@ -11,6 +11,7 @@ use crate::jsid::VoidId;
 use std::cell::UnsafeCell;
 use std::ffi::{c_char, c_void};
 use std::mem;
+use std::ops::DerefMut;
 use std::ptr;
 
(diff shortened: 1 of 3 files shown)


mozjs-sys/src/jsgc.rs
```diff
@@ -153,6 +169,14 @@ pub trait GCMethods {
...
 }
 
+impl<T> GCMethods for Option<T> {
+    unsafe fn initial() -> Self {
+        None
+    }
+
+    unsafe fn post_barrier(_v: *mut Self, _prev: Self, _next: Self) {}
+}
+
```
sagudev, line 179: Shouldn't we use bounds here (`T: GCMethods`) and call underlying methods (`T::post_barrier`)?

jdm:
Closing in favour of #535.

PR #523, opened 2024-11-11
Add default feature for JIT

jschwe:
This allows optionally disabling JIT by removing the feature. This is necessary on some variants of OpenHarmony, when JIT is forbidden by a system policy.

My idea is that on the servo side, I would disable `default-features` for mozjs when `target_env = "ohos"`, since the dependency on mozjs is in the script crate. One could then add a direct dependency in `servoshell` on mozjs (again only for OpenHarmony), so that we can expose a top-level feature `js-jit` to the user running `./mach`, since they are the only one that know if their distribution of openharmony allows jit or not.

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.3-4"
+version = "0.128.3-5"
 authors = ["Mozilla"]
 links = "mozjs"
 build = "build.rs"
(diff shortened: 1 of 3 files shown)


sagudev:
I wonder if android has same problem if we forbid JIT by system?

jschwe:
Is it possible to forbid JIT on android? If so, I'd assume we would hit the same issue.

sagudev:
> Is it possible to forbid JIT on android? If so, I'd assume we would hit the same issue.

I imagine it should be possible using SELinux policies.

BTW I think you also need to disable wasm JIT (wasm_baselinejit, wasm_ionjit) although I am not sure we expose those prefs.

sagudev:
Or maybe we should call https://doc.servo.org/mozjs/jsapi/fn.DisableJitBackend.html before SM init in servo to fully disable JIT. Some doc in mozilla: https://searchfox.org/mozilla-central/source/js/public/Initialization.h#172

sagudev:
I am going to close this given the main usecase was resolved by https://github.com/servo/servo/pull/34231.

PR #524, opened 2024-11-12
Fix mozjs-sys always being out of date

jschwe:
- `src/rustfmt.toml` does not exist, so always out of date
- the configure script is touched by the makefile, so the second build will always be out of date, even without changes.
- bindgen::CargoCallbacks will by default tell Cargo to rerun the build-script if any of the included files change.
  In our case some of the included files are generated by the   build-script, so their mtime will always be newer than
  when the build-script last ran causing cargo to needlessly  rerun the build-script. Writing a customized Callback that ignores generated include files avoids this problem


mrobinson:
Great investigation work. I'm really looking forward to this change.

jdm (approved):
Very nice!

PR #525, opened 2024-11-22
jsglue.cpp: Fix -Winconsistent-missing-override

jschwe:
`-Winconsistent-missing-override` is enabled by default when compiling with clang.


sagudev (approved):

sagudev:
Ideally we would fail cc if there are any warnings, but we would need to resolve warnings first.

PR #530, opened 2024-12-10
Bump bindgen to 0.71.1

jschwe:


sagudev:
This PR needs companion PR in servo, just to be safe.

jschwe:
I will do it together with the mozangle bump

jschwe:
Bindgen seems to have a bug, where specifying `stable(80, 0)` somehow enables features that are gated behind `Stable_1_82` (unsafe extern blocks)

sagudev (commented):
LGTM, but there are still failures in try run.

sagudev:
I think we can just put `ProfilingStackFrame` on block list, as it's not used in servo/mozjs. I am not sure if bindings are correct anyway as all fields are actually atomic (should we block all mozilla::atomic or at least make them opaque, to prevent user from changing them wrongly?).

jschwe:
Good point about the atomics. Just generating Rust atomics as bindings would also be wrong, since the types require at least `RelAcq` ordering. I opened an issue so we can follow up on this separately.

sagudev (approved):

PR #533, opened 2024-12-12
Remove HandleValueArray::from_rooted_slice footgun

jdm:
It's far too easy to misuse by just creating a slice on the stack. Instead, we can add From implementations for constructs that are known to be safe (based on https://searchfox.org/mozilla-central/rev/8f423e3c79147c65d078589f3666dd25ffe80840/js/public/ValueArray.h#54-69). These changes also exposed that Servo was using a duplicate copy of the RootedVec code that included crown annotations and a `from_iter` implementation, so this PR also adds those features to support building Servo successfully with crown linting enabled.

### DIFF
--- a/mozjs/src/lib.rs
+++ b/mozjs/src/lib.rs
@@ -10,6 +10,8 @@
     non_snake_case,
     improper_ctypes
 )]
+#![cfg_attr(feature = "crown", feature(register_tool))]
+#![cfg_attr(feature = "crown", register_tool(crown))]
 
 //!
 //! This crate contains Rust bindings to the [SpiderMonkey Javascript engine][1]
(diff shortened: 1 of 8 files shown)


mozjs/src/lib.rs
```diff
@@ -10,6 +10,8 @@
     non_snake_case,
     improper_ctypes
 )]
+#![cfg_attr(feature = "crown", feature(register_tool))]
+#![cfg_attr(feature = "crown", register_tool(crown))]
```
sagudev, line 14: I am wondering if we should put this behind cfg rather than having a feature, but using a feature makes it easier to use in servo.

sagudev (approved):
LGTM, needs fix of wasm example.

PR #538, opened 2024-12-15
ci: Pin ubuntu to 22.04 in linux builder

sagudev:
This will allow artifacts to be used on older systems (due to glibc).

Issue: https://github.com/servo/servo/pull/34562#issuecomment-2543342461

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       RUSTC_WRAPPER: "sccache"
       CCACHE: sccache
       SCCACHE_GHA_ENABLED: "true"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04 # Needed for artifacts to work on older systems (due to glibc)
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       RUSTC_WRAPPER: "sccache"
       CCACHE: sccache
       SCCACHE_GHA_ENABLED: "true"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04 # Needed for artifacts to work on older systems (due to glibc)
```
jschwe, line 64: I think servo is still using ubuntu-20.04 when creating its artifacts for publishing

jschwe (approved):

.github/workflows/build.yml
sagudev: I though 22.04 was oldest image Github actions supports, apparently not, but it seems that clang is too old.

PR #539, opened 2024-12-15
Allow iterating over the frames of a CapturedJSStack

simonwuelker:
This adds `CapturedJSStack::for_each_stack_frame` which calls the provided closure for each element in the linked list of `SavedFrame`s. Unfortunately, I don't think there's an easy way to make this an iterator over `SavedFrame`s, because of the lifetime associated with a `RootedGuard`.

`for_each_stack_frame` is useful for providing js backtraces in functions like `console.trace`.

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -65,6 +65,7 @@ use crate::jsval::ObjectValue;
 use crate::panic::maybe_resume_unwind;
 use lazy_static::lazy_static;
 use log::{debug, warn};
+use mozjs_sys::jsapi::JS::SavedFrameResult;
 pub use mozjs_sys::jsgc::{GCMethods, IntoHandle, IntoMutableHandle};
 
 use crate::rooted;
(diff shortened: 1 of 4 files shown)


jdm (approved):
Nice!

jdm:
Could you add a test in https://github.com/servo/mozjs/blob/main/mozjs/tests/capture_stack.rs ?

simonwuelker:
> Could you add a test in https://github.com/servo/mozjs/blob/main/mozjs/tests/capture_stack.rs ?

Done. Should I create a new release for mozjs or is it fine to pin the git commit in servo?

simonwuelker:
Ugh, the test passes individually but the other tests are interfering with the engine initialization (because the engine is global state).

jdm:
> > Could you add a test in https://github.com/servo/mozjs/blob/main/mozjs/tests/capture_stack.rs ?
> 
> Done. Should I create a new release for mozjs or is it fine to pin the git commit in servo?

Since this PR only modifies Rust code, we can just update the git revision. C++ changes need a version number update to ensure new static library binaries are published.

PR #540, opened 2024-12-18
Remove jsstreams

sagudev:
Now that https://github.com/servo/servo/pull/34064 landed we can remove this. This is needed so we can start using mozjs artifacts again in servo CI.

This is nice: 
-18k lines


.github/workflows/build.yml
```diff
@@ -135,12 +135,12 @@ jobs:
 
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@v1
-        if: ${{ inputs.release && !contains(matrix.target, 'aarch64') && contains(matrix.features, 'streams') }}
+        if: ${{ inputs.release && !contains(matrix.target, 'aarch64') && matrix.features != 'debugmozjs' }}
```
jschwe, line 138: I guess in the next step we could just rename the artifact to have `-${{ matrix.features}}` suffix, so we can upload multiple versions.

sagudev: Indeed, see https://github.com/servo/mozjs/issues/541

wusyong:
servo is building mozjs without artifact. I suppose this is the PR to fix it?

wusyong (approved):
Tested with manually setting `MOZJS_ARCHIVE` and it is working!
I'll approve here first in case you want to merge it when it's ready.
Whole Servo workflow became very slow today.

sagudev:
I'm waiting for https://github.com/servo/servo/pull/34629 to create companion PR in servo.

PR #543, opened 2024-12-20
Add helper to dump the current JS stack.

jdm:
This is implemented in C++ due to the use of UniqueChars, which has nontrivial C++ construction/.destruction logic associated with it.


sagudev (approved):
LGTM, but needs mozjs-sys version bump.

PR #545, opened 2024-12-24
Allow returning null context pointers from Runtime::get().

jdm:
This assertion fails in various tests and real browser usage, because it's still possible to call this API while shutdown is in progress. Instead of asserting against a valid use, we can use the type system to ensure that callers make informed choices.


sagudev (approved):

sagudev:
I will make PR to fix integrity today.

PR #551, opened 2025-01-20
Indicate to the compiler that Rooted<T> may be uninitialized

### LINKED ISSUE #548: `Rooted::new_unrooted` can easily cause undefined behaviour
`Rooted::new_unrooted` will zero-initialize the contained value (`RootedGuard::new` will usually be used to init it later)

https://github.com/servo/mozjs/blob/0081fc4a3f5fc9891d4377844a874651f7c46041/mozjs-sys/src/jsimpls.rs#L431-L440

This causes immediate undefined behaviour (or a panic in debug mode) if `T` is a type that must not be zero, like `std::ptr::NonNull<T>`.

simonwuelker:
Wraps `Rooted::ptr` in `MaybeUninit` so it can safely be zero-initialized.

Fixes #548

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -1,9 +1,10 @@
 use std::marker::PhantomData;
+use std::mem::MaybeUninit;
 use std::ops::{Deref, DerefMut};
 use std::ptr;
 
 use crate::jsapi::{jsid, JSContext, JSFunction, JSObject, JSScript, JSString, Symbol, Value, JS};
-use mozjs_sys::jsgc::{Initialize, RootKind, Rooted};
+use mozjs_sys::jsgc::{RootKind, Rooted};
 
 use crate::jsapi::Handle as RawHandle;
 use crate::jsapi::HandleValue as RawHandleValue;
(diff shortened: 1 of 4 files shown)


simonwuelker:
These changes also remove the `Initialize` bound on `RootedGuard`, because it is no longer needed (instead of being initialized to a default value, the `ptr` field is now uninitialized).

I think this is an improvement, because it is not always possible to cheaply construct default values for `Initialized`, for example if the type contains `Box` or similar things that need to be allocated.

mozjs/src/gc/root.rs
```diff
@@ -19,60 +20,64 @@ use mozjs_sys::jsgc::ValueArray;
...
         T: Copy,
     {
-        self.root.ptr
+        // SAFETY: The rooted value is initialized as long as we exist
+        unsafe { self.root.ptr.assume_init() }
     }
 
     pub fn set(&mut self, v: T) {
-        self.root.ptr = v;
+        self.root.ptr.write(v);
```
jdm, line 53: We will leak the previous initialized value if this is invoked.

jdm (approved):
This is subtle but looks like a solid fix for the issue that you identified. My only concern is with the `set` method.

mozjs/src/gc/root.rs
simonwuelker: Nice catch!

jdm (approved):

PR #553, opened 2025-01-30
Unified Builds of jsapi and jsglue

### LINKED ISSUE #435: Windows Builds Fail when `clang-cl` Flags are added to `CXXFLAGS`

redfire75369:
Unified flags, as well as the usage of `cc` and `bindgen` for `jsapi` and `jsglue`.
Reorganised the `build.rs` primarily for the unification, but there are some changes in ordering and module organisation.
I'm not 100% sure on the ordering, so if you would like, that can be changed as well.

Resolves #435 through the removal of `-x c++` and addition of `--driver-mode=cl` on windows from the invocations of `bindgen`. Replaces #437.

Note the failures in `aarch64-pc-windows-msvc` are due to an error in llvm 18 https://github.com/llvm/llvm-project/pull/93235 (see https://github.com/rust-lang/rust-bindgen/issues/2842 also). LLVM 18.1.8 is installed on github actions runners for windows-2022. I have tested that LLVM 19 works correctly on spiderfire.

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -633,241 +633,204 @@ fn ignore(path: &Path) -> bool {
     })
 }
 
-mod bindings {
-    use super::BuildTarget;
-
-    impl BuildTarget {
-        pub(crate) fn unsafe_impl_sync_types(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::UNSAFE_IMPL_SYNC_TYPES,
-                BuildTarget::JSGlue => &[],
-            }
-        }
-
-        pub(crate) fn whitelist_types(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::WHITELIST_TYPES,
-                BuildTarget::JSGlue => &[],
-            }
-        }
-
-        pub(crate) fn whitelist_vars(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::WHITELIST_VARS,
-                BuildTarget::JSGlue => &[],
-            }
-        }
-
-        pub(crate) fn whitelist_functions(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::WHITELIST_FUNCTIONS,
-                BuildTarget::JSGlue => &[],
-            }
-        }
-
-        pub(crate) fn blacklist_types(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::BLACKLIST_TYPES,
-                BuildTarget::JSGlue => jsglue::BLACKLIST_TYPES,
-            }
-        }
-
-        pub(crate) fn blacklist_functions(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::BLACKLIST_FUNCTIONS,
-                BuildTarget::JSGlue => &[],
-            }
-        }
-
-        pub(crate) fn opaque_types(self) -> &'static [&'static str] {
-            match self {
-                BuildTarget::JSApi => jsapi::OPAQUE_TYPES,
-                BuildTarget::JSGlue => jsglue::OPAQUE_TYPES,
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


sagudev:
> Note the failures in aarch64-pc-windows-msvc are due an error in llvm 18 https://github.com/llvm/llvm-project/pull/93235 (see https://github.com/rust-lang/rust-bindgen/issues/2842 also). LLVM 18.1.8 is installed on github actions runners for windows-2022. I have tested that LLVM 19 works correctly on spiderfire.

We should pin LLVM to 19 in CI then (but only for aarch64), this is not used in servo so it should not be a problem there.

mozjs-sys/build.rs
```diff
@@ -56,11 +47,7 @@ const SM_TARGET_ENV_VARS: &'static [&'static str] = &[
     "OBJCOPY",
 ];
 
-const EXTRA_FILES: &'static [&'static str] = &["makefile.cargo", "src/jsapi.cpp"];
```
sagudev, line 59: I think we only need makefile.cargo as cc should emit `cargo:rerun-if-changed` for us.

mozjs-sys/build.rs
```diff
@@ -696,434 +556,593 @@ impl ParseCallbacks for CustomCargoCallbacks {
     }
 }
 
-mod jsglue {
-    use crate::CustomCargoCallbacks;
-    use std::env;
-    use std::path::{Path, PathBuf};
+/// The version of moztools we expect.
+#[cfg(windows)]
+const MOZTOOLS_VERSION: &str = "4.0";
```
sagudev, line 561: I think we should keep this on top of file.

mozjs-sys/build.rs
```diff
@@ -696,434 +556,593 @@ impl ParseCallbacks for CustomCargoCallbacks {
...
+            match self {
+                BuildTarget::JSApi => jsapi::MODULE_RAW_LINES,
+                BuildTarget::JSGlue => jsglue::MODULE_RAW_LINES,
+            }
+        }
+    }
+
+    mod jsapi {
+        /// JSAPI types for which we should implement `Sync`.
+        pub(crate) const UNSAFE_IMPL_SYNC_TYPES: &'static [&'static str] = &[
```
sagudev, line 699: can we just inline uses and remove `mod`s

sagudev:
This makes everything much clear and consistent!

Before landing this we should make companion PR to servo to be sure we don't brake anything there.

mozjs-sys/build.rs
Redfire75369: This is possible, but I thought it would look weird if the functions had super long bodies that was just a constant.

sagudev:
I think this will need mozjs-sys bump.

Redfire75369:
Should it be a minor or patch bump?

jdm:
https://github.com/servo/servo/pull/37240

jdm:
All servo builds passed, so we should be good to merge here.

jdm (approved):

PR #554, opened 2025-02-10
Add JSContext wrapper for bindings with fewer raw pointers.

jdm:
This adapts the [JSContext](https://github.com/servo/servo/blob/d733abfca02cdb9fd2af4f0d82ff050e25f71829/components/script/script_runtime.rs#L893) type from Servo's bindings, making it more reliable because we can associate it directly with the underlying runtime wrapper.

This wrapper will enable us in the future to remove the unsafety from traits like ToJSValConvertible and FromJSValConvertible, as well as introduce higher-level wrappers of the JSAPI that do not require using unsafe.

### DIFF
--- a/mozjs/src/lib.rs
+++ b/mozjs/src/lib.rs
@@ -46,6 +46,7 @@ pub mod jsapi {
 pub mod rust;
 
 mod consts;
+pub mod context;
 pub mod conversions;
 pub mod error;
 pub mod gc;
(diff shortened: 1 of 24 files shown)


mozjs/src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub mod jsapi {
 pub mod rust;
 
 mod consts;
+pub mod context;
```
gmorenz, line 49: The file for this module appears to be missing.

jdm: Oops. I'll fix that up tonight, hopefully.

PR #555, opened 2025-03-02
Use Arc::get_mut instead of Arc::strong_count == 1

gmorenz:
Arc::strong_count doesn't come with memory ordering guarantees. As such it should be possible to destory the context multiple times with the previous version of this check. I haven't actually seen or attempted to force that to happen, just came across this while reading code.

See https://github.com/rust-lang/rust/issues/117485

Note: The new code is slightly slower (3 atomic operations when only 1 should be necessary, because it checks for weak pointers as well). I can't imagine we care.


sagudev (approved):
Wow, everyday I learn something new. Great work!

PR #558, opened 2025-03-04
Rename include files to support rust-analyzer

gmorenz:
Rust analyzer can only see files then end in `.rs`, so it was failing to load these, resulting in false errors in vscode (and other rust-analyzer based IDEs).

https://github.com/rust-lang/rust-analyzer/issues/10178


sagudev (approved):
This is great. Thank you!

PR #559, opened 2025-03-04
Remove Copy/Clone from MutableHandle

### LINKED ISSUE #520: MutableHandle are clonable breaking safety provided by &mut
mut is required for https://doc.servo.org/mozjs/gc/root/struct.MutableHandle.html#tymethod.set, but this is still weird as MutableHandle is clonable so `&mut` does not help a lot with safety.

_Originally posted by @sagudev in https://github.com/servo/servo/pull/34087#discussion_r1825494608_

When cloning MutableHandle we still have underlying pointer point to same location, so we can have two &mut that points to same location.

gmorenz:
Fixes #520

Replaces `Copy` and `Clone` with a `reborrow` method that handles lifetimes correctly. Also merges `wrappers` and `jsapi_wrapped` since they now have the same API.

As discussed in #520 there are multiple options here, with no clear "best option". I currently favor this since

 - There's enough complexity in how data is accessed through `RootedGuard` that safe-interior-mutability apis are likely to be unwieldy (these would let us keep `Copy/Clone`.
 - It feels like the least "weird" of the other alternatives.
 
Accompanying servo branch [here](https://github.com/gmorenz/servo/tree/reborrow).

This doesn't fix  `JS::MutableHandle` being  `Copy` and `Clone`. `JS::MutableHandle` doesn't carry a lifetime annotation, so there's no reasonable way to define `reborrow` on it. The fix here is probably to remove the `Deref` and `DerefMut` impls on it (since without lifetime annotations there's no way those are safe functions) and  move servo to always using the lifetime annotated versions.

It also doesn't fix other aliasing issues surrounding this type, those can/should be fixed separately, and apart from an approach that leans entirely into interior mutability, those fixes should be compatible with this change.


gmorenz:
A previous version of this change is [here](https://github.com/gmorenz/mozjs/tree/soundness_1), with an accompying servo branch [here](https://github.com/gmorenz/servo/tree/mutable_handle).

It took the approach of passing around `&mut MutableHandle` instead of explicit reborrows. I decided that I didn't favor that approach because the extra temporary variables felt like more of a compromise than the explicit reborrow calls. The explicit reborrow calls are also more like `clone` and perhaps easier to teach.

That branch also includes a commit removing `Deref[Mut]`. The new branch does not as that can be done as a separate change - though at least removing the `DerefMut` impl is going to be required to fix other aliasing issues with this api.

gmorenz:
Marking this as ready for review with my preferred fix to #520. The design question asked in #520 is still valid, and if we'd rather go another way... this shouldn't be merged as is (though it wouldn't be hard to revert).

I'm also going to open a draft servo PR simultaneously. I'm not sure what the preferred procedure for updating mozjs and servo in lockstep is? If that's a process we'd really rather avoid, this could be split into two. First add `reborrow`. Then update servo to use it and to use `wrappers` apis. Then drop `Copy/Clone` and the `jsapi_wrapped` modules here in a subsequent commit.

mrobinson:
> I'm also going to open a draft servo PR simultaneously. I'm not sure what the preferred procedure for updating mozjs and servo in lockstep is? If that's a process we'd really rather avoid, this could be split into two. First add `reborrow`. Then update servo to use it and to use `wrappers` apis. Then drop `Copy/Clone` and the `jsapi_wrapped` modules here in a subsequent commit.

Typically what we do for stylo is to post the stylo PR, and then post the corresponding Servo PR (with a temporary change to `Cargo.toml` to override the [stylo git ref to the PR commit](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit)). Once both PRs are approved we merge the stylo one first and then undo the temporary change to Servo's `Cargo.toml`, update stylo in `Cargo.lock`, and then merge the Servo PR.

mozjs/src/rust.rs
```diff
@@ -1065,13 +1065,7 @@ macro_rules! capture_stack {
...
 }
 
-/** Wrappers for JSAPI methods that should NOT be used.
- *
- * The wrapped methods are identical except that they accept Handle and MutableHandle arguments
- * that include lifetimes instead.
- *
- * They require MutableHandles to implement Copy. All code should migrate to jsapi_wrapped instead.
- * */
+/// Wrappers for JSAPI methods that ačept lifetimed Handle and MutableHandle arguments
```
sagudev, line 1068:
```suggestion
/// Wrappers for JSAPI methods that accept lifetimed Handle and MutableHandle arguments
```

sagudev (approved):
This make sense to me.

Also for context, second wrapped were introduced in https://github.com/servo/rust-mozjs/pull/422/commits/6ca0c136ab1279e79392b4af4d792cfa5753ca66

jdm (approved):
Thank you for your patience; this is great! I was pleasantly surprised by how few changes were necessary in the Servo codebase, too.

sagudev:
Ah we need to do mozjs bump again, I will take care of it.

gmorenz:
@sagudev - I can get on rebasing the servo PR if that will help?

sagudev:
> @sagudev - I can get on rebasing the servo PR if that will help?

Nah, we just needs to rebase this on top of https://github.com/servo/mozjs/pull/566 (once it lands). Actually MQ will do rebase for use so we just need to resend to queue after https://github.com/servo/mozjs/pull/566 lands.

PR #562, opened 2025-03-06
Fix all the warnings

gmorenz:
- `alloc_error_hook` has evidently been unused for... quite awhile. As far as I can tell the code should still work, up to adding back (conditionally) a nightly feature, so I've done that. 
 - `EncodedStringCallback` is a function called from the C++ glue, it needs to be `extern "C"`. I've also marked this as an `unsafe` fn since it takes a raw pointer as an argument, and to be able to do anything with that it needs to be `unsafe`.
 -  Remove an unused import, this one is my fault.
 - jsapi has conflicting imports from two sets of bindgen, `js` and `JS` - currently this means that neither are imported. To silence these warnings, we pick one.
   - Both have a `mod detail` with different contents, `detail::*` is also re-exported for both, so just pick one arbitrarily.
   - `FrontendContext` exists and has different functions using it in both, the `JS` functions look more likely to be called via servo, so re-export that one.
   - `MemoryUse` exists in both, but only the `JS` one has any functions using it, so re-export that one.

### DIFF
--- a/mozjs/tests/callback.rs
+++ b/mozjs/tests/callback.rs
@@ -64,12 +64,13 @@ unsafe extern "C" fn puts(context: *mut JSContext, argc: u32, vp: *mut Value) ->
     let arg = mozjs::rust::Handle::from_raw(args.get(0));
     let js = mozjs::rust::ToString(context, arg);
     rooted!(in(context) let message_root = js);
-    EncodeStringToUTF8(context, message_root.handle().into(), |message| {
+    unsafe extern "C" fn cb(message: *const core::ffi::c_char) {
         let message = CStr::from_ptr(message);
         let message = str::from_utf8(message.to_bytes()).unwrap();
         assert_eq!(message, "Test Iñtërnâtiônàlizætiøn ┬─┬ノ( º _ ºノ) ");
         println!("{}", message);
-    });
+    }
+    EncodeStringToUTF8(context, message_root.handle().into(), cb);
 
     args.rval().set(UndefinedValue());
     true
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -27,6 +27,7 @@ debugmozjs = []
 profilemozjs = []
 jitspew = []
 crown = []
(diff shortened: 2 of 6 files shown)


mozjs/tests/callback.rs
```diff
@@ -64,12 +64,13 @@ unsafe extern "C" fn puts(context: *mut JSContext, argc: u32, vp: *mut Value) ->
     let arg = mozjs::rust::Handle::from_raw(args.get(0));
     let js = mozjs::rust::ToString(context, arg);
     rooted!(in(context) let message_root = js);
-    EncodeStringToUTF8(context, message_root.handle().into(), |message| {
```
gmorenz, line 67: Unfortunately closures won't coerce to C-ffi functions, so this becomes a bit more awkward to use.

There are currently no uses of this in servo, so no coordination is needed.

sagudev: I wonder if we can make it generic over closure `rust::EncodeStringToUTF8<F>(..., cb_c: F)` then use it like:
```rust
unsafe extern "C" fn cb(message: *const core::ffi::c_char) {
    cb_c()
}
EncodeStringToUTF8(..., cb)
```

but this is just an idea.

sagudev:
> jsapi has conflicting imports from two sets of bindgen, js and JS - currently this means that neither are imported. To silence these warnings, we pick one. 

Our exports are a mess.

sagudev (approved):

mozjs/tests/callback.rs
sagudev: I think it's not really needed (at least while we [store DOMString as rust str internally instead of utf-16](https://github.com/servo/servo/blob/37634b1251a10cdb873750456026ff3bf68693c3/components/script_bindings/str.rs#L156)). Also SMDOC says its [DEPRECATED](https://searchfox.org/mozilla-central/rev/5c3577d8865aa2d61fa89fd453d803b7c4f8b4ed/js/public/CharacterEncoding.h#410), so let's not deal with this.

PR #563, opened 2025-03-10
Implement `Copy` and `Clone` for `Handle<T>` where `T: !Copy`

gmorenz:
Derive only derives these impls for types where the generics satisfy them, but shared references (and thus `Handle`) can implement them for any type `T`.

I've run into this a few times while playing around now, because `Cell` doesn't implement `Copy`. It's not entirely clear that users of the library will run into this, but at worst it doesn't hurt to define these generally.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -100,10 +100,17 @@ impl<'a, const N: usize> From<&RootedGuard<'a, ValueArray<N>>> for JS::HandleVal
     }
 }
 
-#[derive(Clone, Copy)]
 pub struct Handle<'a, T: 'a> {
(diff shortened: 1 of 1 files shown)


gmorenz:
PS. It's also pretty likely that users will run into this if/when `Handle` is used more broadly. Right now the reason no one is running into this is because for the most part `&`/`&mut` referencess into `RootedGuard` are used.

jdm (approved):
Makes sense.

sagudev:
This will need a real rebase after https://github.com/servo/mozjs/pull/559 lands.

gmorenz:
Rebased, should be good to go.

PR #568, opened 2025-03-22
Support creating Rc<T> values from JS values.

jdm:
As part of https://github.com/servo/servo/issues/36086 we need to support creating Rust Promise wrappers from JS values. We can do this through a layer of indirection for types that need to be wrapped inside of an Rc value: we introduce a new conversion trait that can be implemented by those types in other crates, which allows us to avoid the orphan rule.

### DIFF
--- a/mozjs/src/conversions.rs
+++ b/mozjs/src/conversions.rs
@@ -166,6 +166,29 @@ pub trait FromJSValConvertible: Sized {
     ) -> Result<ConversionResult<Self>, ()>;
 }
 
+/// A trait to convert `JSVal`s to Rust types inside of Rc wrappers.
+pub trait FromJSValConvertibleRc: Sized {
+    /// Convert `val` to type `Self`.
+    /// If it returns `Err(())`, a JSAPI exception is pending.
+    /// If it returns `Ok(Failure(reason))`, there is no pending JSAPI exception.
+    unsafe fn from_jsval(
+        cx: *mut JSContext,
+        val: HandleValue,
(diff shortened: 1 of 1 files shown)


mozjs/src/conversions.rs
```diff
@@ -166,6 +166,29 @@ pub trait FromJSValConvertible: Sized {
...
+    /// If it returns `Err(())`, a JSAPI exception is pending.
+    /// If it returns `Ok(Failure(reason))`, there is no pending JSAPI exception.
+    unsafe fn from_jsval(
+        cx: *mut JSContext,
+        val: HandleValue,
+    ) -> Result<ConversionResult<Rc<Self>>, ()>;
+}
+
+impl<T: FromJSValConvertibleRc> FromJSValConvertible for Rc<T> {
+    type Config = ();
```
jdm, line 181: While I would prefer to use `type Config = T::Config;` here, I was not able to make that compile in Servo's generated bindings. I kept running into compiler errors where passing the `()` config option to FromJSValConvertible for a `Rc<Promise>` where the compiler expected an associated type.

sagudev: Maybe because T (`FromJSValConvertibleRc`) does not have any `type Config` in trait definition.

sagudev (approved):

mozjs/src/conversions.rs
jdm: The errors I refer to.iccurred when both traits were defined identically (apart from the Rc return value).

Issue #569, opened 2025-03-23
GC related safety and soundness issues


gmorenz:
While working on #520 other soundness and safety issues came up, this is something of an omnibus issue regarding them.

By "handle like types" I mean `RootedGuard`, `RootedVec`, `MutableHandle`, and `JS::MutableHandle`, as well as potentially `RootedTraceableBox`, and `CustomAutoRooter`.

## Soundness - correct use of the public API results in undefined behavior

 - [ ] `Traceable` modifies values behind a `&T` reference (#560).
 - [x] `RootedGuard` stores a `&mut` reference to data the GC looks at and modifies over GC pauses (#564).
 - [ ] `Handle` stores an `&` reference to data (not wrapped in a `Cell` type) that the GC modifies over GC pauses.
 - [ ] The `Deref` implementations for handle like types allow for safe code to hold an `&` reference to data (not wrapped in a `Cell` type) that the GC will modify at over GC pauses. (Arguably safety, but I think we'll end up allowing borrowing over GC pauses).
    - [ ] Some of this data is modified via `Traceable`.
    - [ ] Some of this data is modified directly in `Rooted` via `RootKind`.
 - [ ] (TODO: Verify) [`ForOfIteratorGuard::new`](https://github.com/servo/mozjs/blob/26c35182c1b2b7fb2fbfa93b82669533d92c3a60/mozjs/src/conversions.rs#L666) is suspicious, it likely has issues similar to those described in #564.

## Safety - it's possible to use this improperly in safe code

 - [ ] The `DerefMut` implementations for handle like types allow for safe code to hold an `&mut` reference to data the GC will look at and modify over GC pauses.
 - [ ] #520
    - [x] #559
    - [ ] `JS::MutableHandle` still has this issue.
 - [ ] `JS::[Mutable]Handle` allows use after free in safe code via `Deref[Mut]`, `get` and `set`.

gmorenz:
Proposed fixes (some of this I'm posting for the first time, other of this I'm copying in from previous discussions):

## GC using values while we have a `&mut` reference to them.

Apart from stopping holding `&mut` references ourselves, we want to stop the user from doing so. I.e. we need to remove `DerefMut` impls.

We can replace these with

 - `Cell` like `get` and `set` functions (since we're single threaded these are safe from GC pauses)
[...]

PR #572, opened 2025-03-26
Remove DerefMut impl for RootedGuard

gmorenz:
- We no longer construct `MutableHandle` via `deref_mut`. Doing so meant that the raw pointer in `MutableHandle` was only valid for as long as the mutable reference created in `deref_mut` would be. I.e. until the next time GC runs. This is an extension of #564, and doesn't require a change in public API.
 - It removes the `deref_mut` impl entirely, as it's an unsafe footgun that isn't just marked as safe, but is silently invoked.
 - Creates a helper `take` function for `RootedGuard<Option<T>>`, since that was the primary use of `DerefMut` in servo, and is a safe thing to do.
 - It adds `allow_unrooted_interior` to `MutableHandle` because the accompanying servo PR requires that crown understand `MutableHandle` may point to things that must be rooted over GC passes.
 
Accompanying servo PR: https://github.com/servo/servo/pull/36158

 Addresses a small part of #569.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -50,6 +50,13 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
         // SAFETY: self.root points to an inbounds allocation
         unsafe { (&raw mut (*self.root).ptr).cast() }
     }
+
+    /// Safety: GC must not run during the lifetime of the returned reference.
+    pub unsafe fn as_mut<'b>(&'b mut self) -> &'b mut T
+    where
+        'a: 'b,
+    {
+        &mut *(self.as_ptr())
     }
 
     pub fn get(&self) -> T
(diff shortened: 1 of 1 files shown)


gmorenz:
Servo's CI is complaining because `MutableHandle<T>` is unrooted, and crown doesn't know about that type. So I've added another commit to this PR that adds `crown::unrooted_must_root_lint::allow_unrooted_interior` to `MutableHandle`. I believe that's correct because the purpose of this `Handle` is to point at something that is rooted... and thus the GC will update everything it needs to. That said I don't deeply understand what crown is checking.

mozjs/src/gc/root.rs
```diff
@@ -43,7 +43,17 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
...
 
     pub fn handle_mut(&mut self) -> MutableHandle<T> {
-        unsafe { MutableHandle::from_marked_location(self.deref_mut()) }
+        unsafe { MutableHandle::from_marked_location(self.as_ptr()) }
+    }
+
+    pub fn as_ptr(&self) -> *mut T {
+        // SAFETY: self.root points to an inbounds allocation
+        unsafe { (&raw mut (*self.root).ptr).cast() }
+    }
```
sagudev, line 52: This ptr can become invalid if GC runs, right?

gmorenz: No, I don't think so, happy to be proven wrong if I am though. This pointer should be equivalent [^1] to the `self.root` ptr, which should be either equivalent to or a parent of any pointer the GC uses, so the GC shouldn't invalidate it.

Miri seems to agree it isn't invalidated in [this toy example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4222e47257fdcb7e6d7c8dbeb126f85d).

[^1]: Equivalent in the tree borrows model. In the same contiguous block of `SharedRW` tags in the stacked borrows model.

sagudev: Indeed, I got confuse with all indirection.

mozjs/src/gc/root.rs
```diff
@@ -43,7 +43,17 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
...
+
+    /// Safety: GC must not run during the lifetime of the returned reference.
+    pub unsafe fn as_mut<'b: 'a>(&mut self) -> &'b mut T {
+        &mut *(self.as_ptr())
```
sagudev, line 56: Those lifetimes are wrong as per https://github.com/servo/mozjs/pull/573#discussion_r2013405578

gmorenz: Oops, forgot I copy and pasted this, thanks for catching this twice now

gmorenz: Updated again as per discussion in https://github.com/servo/mozjs/issues/569

sagudev (approved):
LGTM, but with this GC stuff it's always good to have another pair of eyes.

jdm (approved):

PR #573, opened 2025-03-26
Remove DerefMut impl for Mutable Handle

gmorenz:
Removes the `deref_mut` impl, and replaces it with an unsafe `as_mut` method. As discussed in #569 (and other recent issues) `deref_mut` is unsafe since it aliases with the GC, and cannot be made safe or marked unsafe.

The accompanying servo change (servo/servo#36160) does not need the `as_mut` method, it feels like an appropriate escape hatch to include though.

It also includes a `MutableHandle::<Option<_>>::take` method, since like in #572 that's a safe method, and #572's servo change results in it being used on `MutableHandle`.

Independent from (the obviously related) #572 and the similar incoming PR for `JS::MutableHandle`. Doing these all at roughly the same time feels like the right move, since otherwise we might accidentally encourage someone to convert from one of these types another to access the `DerefMut` function instead of using actual alternatives.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -216,6 +216,14 @@ impl<'a, T> MutableHandle<'a, T> {
         unsafe { *self.ptr = v }
     }
 
+    /// Safety: GC must not run during the lifetime of the returned reference.
+    pub unsafe fn as_mut<'b>(&'b mut self) -> &'b mut T
+    where
+        'a: 'b,
+    {
+        &mut *(self.ptr)
+    }
+
     /// Creates a copy of this object, with a shorter lifetime, that holds a
     /// mutable borrow on the original object. When you write code that wants
     /// to use a `MutableHandle` more than once, you will typically need to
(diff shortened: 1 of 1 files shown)


gmorenz:
Oops, this is not actually independent from #572. #572 creates an instance where `MutableHandle::<Option<_>>::take` is called... I'll fix that like I fixed the many more `take` calls in #572 by adding a `take` helper function.

mozjs/src/gc/root.rs
```diff
@@ -216,6 +216,11 @@ impl<'a, T> MutableHandle<'a, T> {
         unsafe { *self.ptr = v }
     }
 
+    /// Safety: GC must not run during the lifetime of the returned reference.
+    pub unsafe fn as_mut<'b: 'a>(&mut self) -> &'b mut T {
```
sagudev, line 220: I think those lifetimes are wrong. We do not want for `'b` to outlive `'a`, right?

Rust lifetimes `'long: 'short` are read as "'long outlives 'short".`

sagudev: And these are right: https://github.com/servo/mozjs/blob/3e8e9d38531e74c75b3ad4921e895c9e8a496082/mozjs/src/gc/root.rs#L229-L232

gmorenz: You're right, this is backwards... I'm also not sure why I thought this was a good idea. The lifetime should be bounded by the lifetime of the mutable reference to self, not just the lifetime of the data it points at. Will fix.

gmorenz: Updated again as per discussion in #569

sagudev (approved):

jdm (approved):

PR #574, opened 2025-03-26
Remove DerefMut impl for `JS::MutableHandle`

gmorenz:
Like with #572 and #573 this impl allows for improper aliasing with the GC. With this type it also allows for outright use after frees, since `JS::MutableHandle`, like a raw pointer, has no lifetime attached to it. Eventually that's a justification for removing the `Deref` impl as well, but once thing at a time.

There's only one use of this in servo, which is removed in the accompanying PR: https://github.com/servo/servo/pull/36161

Like with #572 and #573 I added an unsafe `as_mut` method, though it is unused locally this time as well as in servo. It feels like a reasonable escape hatch to share even if it isn't currently used.

Part of #569, and can be merged entirely independently from #572 and #573.

### DIFF
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -112,6 +112,13 @@ impl<T> JS::MutableHandle<T> {
     {
         unsafe { *self.ptr = v }
     }
+
+    /// The returned pointer is aliased by a pointer that the GC will read
+    /// through, and thus `&mut` references created from it must not be held
+    /// across GC pauses.
+    pub fn as_ptr(self) -> *mut T {
+        self.ptr
+    }
 }
 
 impl JS::HandleValue {
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -27,7 +27,6 @@ use crate::jsval::{JSVal, UndefinedValue};
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::Deref;
-use std::ops::DerefMut;
 use std::ptr;
 
 impl<T> Deref for JS::Handle<T> {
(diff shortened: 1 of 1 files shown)


mozjs-sys/src/jsimpls.rs
```diff
@@ -119,6 +112,15 @@ impl<T> JS::MutableHandle<T> {
...
     }
+
+    /// SAFETY: The returned reference must not be held across a GC pause,
+    /// and this method must not be used to violate rust's aliasing rules.
+    ///
+    /// Note that it is possible to create multiple `JS::MutableHandle<T>`
+    /// objects that point to the same value.
+    pub unsafe fn as_mut(&mut self) -> &mut T {
+        &mut *self.ptr
+    }
```
sagudev, line 123: I am wondering if it's even reasonable to have this escape hatch? Or at least should it really return ref instead of ptr?

gmorenz: I think some sort of way to get a pointer-type to the data makes sense. Any time you have a handle to something with a complex type (like an Option), there's a lot of perfectly valid things to do to it (like call `.take()`) that can't easily be done without exposing this. I suppose we could say for this type the way to do that is "cast to a `MutableHandle` and use its `as_mut`, but that feels like creating unnecessary friction for anyone who ends up needing this.

Thankfully servo doesn't currently use the `JS::MutableHandle` type in that way, but I'd rather not be getting in the way of some future developer who ends up with one of them and needs to do something with it.

You're right that a conversion to a pointer might make more sense for this type though.... I'll change this to that.

mozjs-sys/src/jsimpls.rs
```diff
@@ -119,6 +112,13 @@ impl<T> JS::MutableHandle<T> {
     {
         unsafe { *self.ptr = v }
     }
+
+    /// The returned pointer is aliased by a pointer that the GC will read
```
gmorenz, line 116: I've added a task to the tracking issue to update this comment if we don't convert the GC to only write via interior mutability.

sagudev (approved):

PR #576, opened 2025-03-29
Refactor Rooted<T>

gmorenz:
- Instead of storing `MaybeUninit<T>` in `Rooted<T>`, and initializing `RootedGuard` from a `&mut Rooted<T>`, store `T` in `Rooted<T>` and initialize `RootedGuard` from an `&mut MaybeUninit<Rooted<T>>`.
  
   Currently every time we touch the data through `RootedGuard<T>`, we need to re-assert to the compiler that it is initialized. Adding one more thing to think about in an already complicated safety dance. By instead getting rid of the half initialized state, and passing the entire place in as uninitialized memory, this part of initialization can happen through entirely safe functions, and access never needs to worry about it.
 - Correctly mark `ValueArray::get_ptr` as safe - it's doing things with the pointer that might be unsafe.
 
   This lets us drop the unsafe block in `fn from(array: &Rooted<ValueArray<N>>) -> JS::HandleValueArray`
 - Rename `Rooted::ptr` to `Rooted::data`, whenever working with this type via `RootedGuard`, we currently have a pointer to a field name `ptr` which is in fact not (necessarily) a pointer. This is... confusing. Let's change the name to something else.
 
None of these changes are necessary for correctness or anything, they would just make me happier when reading the code.

Servo still builds with this change, we're technically changing public APIs, but not ones that are actually directly used.

### DIFF
--- a/mozjs-sys/src/jsgc.rs
+++ b/mozjs-sys/src/jsgc.rs
@@ -104,7 +104,7 @@ pub trait Rootable: crate::trace::Traceable + Sized {
     unsafe extern "C" fn trace(this: *mut c_void, trc: *mut JSTracer, _name: *const c_char) {
         let rooted = this as *mut Rooted<Self>;
         let rooted = rooted.as_mut().unwrap();
-        <Self as crate::trace::Traceable>::trace(rooted.ptr.assume_init_mut(), trc);
+        <Self as crate::trace::Traceable>::trace(&mut rooted.ptr, trc);
     }
 }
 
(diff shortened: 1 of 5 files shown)


jdm (approved):
Clever! I appreciate how this makes a bunch of code easier to read, too.

sagudev:
> Rename Rooted::ptr to Rooted::data, whenever working with this type via RootedGuard, we currently have a pointer to a field name ptr which is in fact not (necessarily) a pointer. This is... confusing. Let's change the name to something else.

This is actually from mozilla: https://searchfox.org/mozilla-central/source/js/public/RootingAPI.h#1250, but I agree the naming is not really clear.

PR #577, opened 2025-04-10
Allow re-initializing the JavaScript engine after shutting down

mrobinson:
As far as I can tell, there's no reason that once the engine it shut
down, it cannot be started again. This will allow creating multiple
`Servo` instances which is important for unit testing Servo.

Signed-off-by: Martin Robinson <mrobinson@igalia.com>

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -155,7 +155,6 @@ enum EngineState {
     Uninitialized,
     InitFailed,
     Initialized,
-    ShutDown,
 }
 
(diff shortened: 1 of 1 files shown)


redfire75369:
According to the docs for `JS_ShutDown`, initialising SpiderMonkey multiple times is not allowed.
https://searchfox.org/mozilla-central/rev/601256c3cc6f397b018995810fd3f586570f50ee/js/public/Initialization.h#188-190

mrobinson:
Hrm. That's unfortunate. I guess Servo simply needs to wait until program termination to call this method.

PR #578, opened 2025-04-12
Expose function to obtain a PropertyKey from a JSString.

jdm:
This is needed to replicate https://searchfox.org/mozilla-central/rev/1f65969e57c757146e3e548614b49d3a4168eeb8/js/public/Id.h#183 for https://github.com/servo/servo/issues/34686.

### DIFF
--- a/mozjs-sys/src/jsid.rs
+++ b/mozjs-sys/src/jsid.rs
@@ -44,6 +44,13 @@ pub fn SymbolId(symbol: *mut Symbol) -> jsid {
     AsPropertyKey((symbol as usize) | (PropertyKeyTag::Symbol as usize))
 }
 
+#[inline(always)]
+pub fn StringId(s: *mut JSString) -> jsid {
+    assert!(!s.is_null());
+    assert_eq!((s as usize) & JSID_TYPE_MASK, 0);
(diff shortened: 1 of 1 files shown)


mozjs-sys/src/jsid.rs
```diff
@@ -44,6 +44,13 @@ pub fn SymbolId(symbol: *mut Symbol) -> jsid {
     AsPropertyKey((symbol as usize) | (PropertyKeyTag::Symbol as usize))
 }
 
+#[inline(always)]
+pub fn StringId(s: *mut JSString) -> jsid {
```
sagudev, line 48:
```suggestion
#[inline(always)]
/// <https://searchfox.org/mozilla-central/rev/1f65969e57c757146e3e548614b49d3a4168eeb8/js/public/Id.h#183>
pub fn StringId(s: *mut JSString) -> jsid {
```

sagudev (approved):
Instead of rewriting SM in mozjs we should probably call original C++ functions via glue, but given that we already have other `*Id` functions this will do.

redfire75369:
This looks to be missing the safety criteria that the `JSString*` is a non-integer Atom. An integer atom is a string with the value < 2^32 - 1.

PR #580, opened 2025-04-13
Bump mozjs-sys version.

jdm:
https://github.com/servo/mozjs/pull/579/ modified the compiled C++ code, so we need to publish new compiled libraries.


sagudev (approved):
Whoops, that slipped my mind.

PR #581, opened 2025-05-06
Respect MOZJS_FROM_SOURCE=0

jschwe:
Allow disabling the from source build explicitly.


mozjs-sys/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.128.9-1"
```
sagudev, line 5: we do not actually need to do bump here (because there is no C/C++ changes), but nothing is wrong if we do.

sagudev (approved):

PR #583, opened 2025-05-21
ohos: Use custom domain when logging with hilog

### LINKED ISSUE #582: ohos: Set log domain
To make filtering logs easier we should set the log domain to something unique, non-zero. 
The following patch should be adjusted accordingly: https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch

Previous servo PR: https://github.com/servo/servo/pull/36913 
Probably we should choose a different, but adjacent log domain than servo, since it's nicer for filtering purposes to separate what we can.

jschwe:
This modifies the existing patch to redirect log messages to hilog to use a non-zero domain, which allows us to filter logs via the domain on ohos devices.
We set 0xE0C3 for Rust code in servo, using a different number allows us to seperate spidermonkey from servo logs if wanted.

Fixes #582

### DIFF
--- a/mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
+++ b/mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
@@ -2,11 +2,12 @@ diff --git a/build/moz.configure/init.configure b/build/moz.configure/init.confi
 index e0e14bf60..179de2592 100644
 --- a/build/moz.configure/init.configure
 +++ b/build/moz.configure/init.configure
-@@ -956,6 +956,7 @@ def target_is_ohos(target):
+@@ -956,6 +956,8 @@ def target_is_ohos(target):
 
 
  set_define("XP_OHOS", target_is_ohos)
 +set_config("OHOS", target_is_ohos)
++set_define("OHOS_LOG_DOMAIN", "0xE0C4", when=target_is_ohos)
 
 
  @depends(target)
@@ -34,7 +35,7 @@ index 3cfc92533..9c487ac45 100644
 +#ifdef ANDROID
 +  __android_log_print(ANDROID_LOG_ERROR, "Gecko", "mozalloc_abort: %s", msg);
 +#elif defined(XP_OHOS)
-+    (void) OH_LOG_Print(0 /* LOG_APP */, 7 /* LOG_FATAL */, 0, "Gecko",
++    (void) OH_LOG_Print(0 /* LOG_APP */, 7 /* LOG_FATAL */, OHOS_LOG_DOMAIN, "Gecko",
 +         "mozalloc_abort: %{public}s\n", msg);
 +#else
    fputs(msg, stderr);
@@ -65,7 +66,7 @@ index 0b7395177..e84d76aba 100644
                              /* aMaxFrames */ 0);
  #  endif
 +#elif defined(XP_OHOS)
-+    (void) OH_LOG_Print(0 /* LOG_APP */, 7 /* LOG_FATAL */, 0, "MOZ_Assert",
++    (void) OH_LOG_Print(0 /* LOG_APP */, 7 /* LOG_FATAL */, OHOS_LOG_DOMAIN, "MOZ_Assert",
 +     "Assertion failure: %{public}s, at %{public}s:%{public}d\n",
 +     aStr, aFilename, aLine);
  #else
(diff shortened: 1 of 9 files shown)


mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
```diff
@@ -47,7 +47,7 @@ index 3cfc92533..9c487ac45 100644
 +#ifdef ANDROID
 +  __android_log_print(ANDROID_LOG_ERROR, "Gecko", "mozalloc_abort: %s", msg);
 +#elif defined(OHOS)
-+    (void) OH_LOG_Print(LOG_APP, LOG_ERROR, 0, "Gecko",
++    (void) OH_LOG_Print(LOG_APP, LOG_ERROR, 0xEOC4, "Gecko",
```
mukilan, line 50: Would it make sense to define a constant for this domain value, maybe in the [init.configure](https://github.com/servo/mozjs/blob/728acdf3d4ce0604e9f75dd1d539dc6f291ccec7/mozjs-sys/mozjs/build/moz.configure/init.configure#L958), rather than hard coding it?

jschwe: Yeah, that is definitely better than copying the same value everywhere.

mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
```diff
@@ -10,6 +10,19 @@ index e0e14bf60..179de2592 100644
...
  @depends(target)
+@@ -1019,6 +1019,12 @@
+
+ set_config("ANDROID_CPU_ARCH", android_cpu_arch)
+
++@depends(target, when=target_is_ohos)
++def ohos_hilog_domain(target):
++    return "0xE0C4"
++
++set_define("OHOS_LOG_DOMAIN", ohos_hilog_domain)
```
mukilan, line 21: It looks like set_define [also accepts a `when` argument](https://github.com/servo/mozjs/blob/728acdf3d4ce0604e9f75dd1d539dc6f291ccec7/mozjs-sys/mozjs/build/moz.configure/arm.configure#L251). Is it not possible to use it here?

mukilan (approved):

mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
jschwe: I checked the implementation (set_define_impl) and the doc string says that when the value is false, the define will be explicitly undefined (-U).

jschwe: Ah, did you want to suggest to just remove the `ohos_hilog_domain` function and instead use when? that makes sense to me.

mukilan: Right, my idea was to see if we can just do `set_define("OHOS_LOG_DOMAIN", "0xE0C4", when=target_is_ohos)`

jschwe:
I'm not really sure what is going on here:

```
@depends(target)
def target_is_ohos(target):
    return False


set_define("XP_OHOS", target_is_ohos)
```

This causes `XP_OHOS` to be defined. 

```
@depends(target)
def target_is_ohos(target):
    return None


set_define("XP_OHOS", target_is_ohos)
```

This does not define XP_OHOS. This seems really weird to me, and goes against the documentation of `set_define_impl` and also how other defines are set.

jschwe:
IMHO yes. I didn't proceed, because it seems I don't understand `set_define` and am not sure if I may be doing something wrong.

mukilan:
> This does not define XP_OHOS. This seems really weird to me, and goes against the documentation of `set_define_impl` and also how other defines are set.

I think the documentation of `set_define_impl` is just wrong, or its a bug in `set_define_impl`. The exisiting [CROSS_COMPILE](https://github.com/servo/mozjs/blob/ae642f8cd3d2b462b40b99a11da06df9912ecc2c/mozjs-sys/mozjs/build/moz.configure/init.configure#L758) define is controlled by a boolean as well and I can see in my macOS builds it is defined to be an empty value in `out/build/config.status` which translates to `-DCROSS_COMPILE='' `. The same happens for XP_OHOS when False is returned from `target_is_ohos`.

The existing [defines for XP_DARWIN](https://github.com/servo/mozjs/blob/ae642f8cd3d2b462b40b99a11da06df9912ecc2c/mozjs-sys/mozjs/build/moz.configure/init.configure#L855), XP_WINDOW don't have a False return and instead let the predicate methods just return `None` by default, when the if doesn't succeed, which matches what you have in this PR.

jdm (approved):

PR #589, opened 2025-07-07
Add default feature for JIT

jschwe:
This is a revival of https://github.com/servo/mozjs/pull/523 - but with a slightly different background.
When building servo on Linux with the profile `production-stripped`, disabling JIT with this change shaves off 9MiB from the servo binary size, which is around 10% (80MiB vs. 89MiB).

Companion PR: https://github.com/servo/servo/pull/37972

### DIFF
--- a/mozjs/Cargo.toml
+++ b/mozjs/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.14.1"
+version = "0.14.2"
 authors = ["The Servo Project Developers"]
 license.workspace = true
 edition.workspace = true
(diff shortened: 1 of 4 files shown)


jdm (approved):

mozjs/Cargo.toml
```diff
@@ -12,6 +12,7 @@ doctest = false
 
 [features]
 debugmozjs = ['mozjs_sys/debugmozjs']
+jit = ['mozjs_sys/jit']
```
jdm, line 17: This should be a default feature for the mozjs crate as well.

jschwe: I moved the default feature to mozjs and removed it as default feature from mozjs_sys. Otherwise we couldn't disable it, unless `mozjs` would depend on `monz_js_sys` with `default-features = false`. Since I don't think there would be any other users of the sys crate, besides `mozjs`, I think it makes the most sense to expose the jit feature as default in `mozjs` and don't add any default features to mozjs_sys. 
What do you think?

jdm: Yep, I agree!

PR #590, opened 2025-07-08
Update Spidermonkey to 128.13 and fix CI

jschwe:
The integrity check is failing, so bump SM to the latest ESR version.
This also fixes the failing windows CI job.


.github/workflows/build.yml
```diff
@@ -121,16 +121,16 @@ jobs:
         run: |
           curl -SL "https://github.com/servo/servo-build-deps/releases/download/msvc-deps/moztools-4.0.zip" --create-dirs -o target/dependencies/moztools.zip
           cd target/dependencies && unzip -qo moztools.zip -d .
-      - name: Install LLVM 19 on aarch64
-        if: contains(matrix.target, 'aarch64')
+      - name: Install LLVM 19
         run: |
           # install due to build errors caused by a bug in llvm 18
           # https://github.com/llvm/llvm-project/pull/93235
           # remove when windows runner updates to llvm 19
```
sagudev, line 127: We need this now because MSVC has headers that requires clang 19: https://github.com/actions/runner-images/issues/12435, so this comment is stale.

jschwe:
```suggestion
          # MSVC has headers that require clang-19.
          # remove when windows runner updates to llvm 19
```

Would this be sufficient, or should we link the issue in the comment?

PR #592, opened 2025-07-12
Add AsHandleValue trait to safely convert Heap<Value> to HandleValue

Taym95:
Needed for https://github.com/servo/servo/issues/37258

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -4,6 +4,7 @@ use std::ops::Deref;
 use std::ptr;
 
 use crate::jsapi::{jsid, JSContext, JSFunction, JSObject, JSScript, JSString, Symbol, Value, JS};
+use mozjs_sys::jsapi::JS::Heap;
 use mozjs_sys::jsgc::{RootKind, Rooted};
 
 use crate::jsapi::Handle as RawHandle;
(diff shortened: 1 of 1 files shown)


redfire75369:
`Heap<Value>` doesn't root anything on its own. For this to be safe, it needs to be on a traced `Heap<T>`.

Taym95:
> `Heap<Value>` doesn't root anything on its own. For this to be safe, it needs to be on a traced `Heap<T>`. So I'm pretty sure this isn't safe.

Indeed, as_handle_value Is only safe If it's stored inside a structure that is traced by SpiderMonkey (e.g. inside a RootedGuard, JS::Rooted) I can rename it to `AsHandleValueIfRooted` with more details doc.

context:
We need this trait to safely convert a rooted Heap<Value> into a HandleValue<'a> without repeating unsafe { HandleValue::from_raw(self.handle()) } at every call site

jdm:
I'm not a fan of this solution because it doesn't actually force the caller to use it in a safe way; it just hides an unsafe conversion.

jdm:
I think if we annotate Heap with the crown::unrooted_must_root so that we can have Unrooted values be compiler errors I would be willing to reevaluate this.

jdm:
I would also prefer creating the rust::Handle values from the Heap values directly so the lifetime comes from the self value. Handle::from_raw is _rrally_ unsafe because the lifetime has no relationship with the original source.

sagudev:
> I think if we annotate Heap with the crown::unrooted_must_root so that we can have unrooted values be compiler errors I would be willing to reevaluate this.

Wouldn't that be to strict in some cases?

sagudev:
> I would also prefer creating the rust::Handle values from the Heap values directly so the lifetime comes from the self value. Handle::from_raw is _really_ unsafe because the lifetime has no relationship with the original source.

In this case we should put the trait in servo, as we need crown to ensure it's safe.

Taym95:
> I would also prefer creating the rust::Handle values from the Heap values directly so the lifetime comes from the self value. Handle::from_raw is _really_ unsafe because the lifetime has no relationship with the original source.

I will do that in Servo.

Taym95:
> I think if we annotate Heap with the crown::unrooted_must_root so that we can have unrooted values be compiler errors I would be willing to reevaluate this.

sounds good.

PR #594, opened 2025-07-18
Add crown::unrooted_must_root_lint::must_root to Heap

Taym95:
Add crown::unrooted_must_root_lint::must_root to Heap

### DIFF
--- a/mozjs-sys/src/jsgc.rs
+++ b/mozjs-sys/src/jsgc.rs
@@ -324,6 +324,7 @@ pub type RootedValueArray<const N: usize> = Rooted<ValueArray<N>>;
 /// SpiderMonkey.
 ///
 /// For safe `Heap` construction with value see `Heap::boxed` function.
+#[cfg_attr(feature = "crown", crown::unrooted_must_root_lint::must_root)]
 #[repr(C)]
 #[derive(Debug)]
 pub struct Heap<T: GCMethods + Copy> {


mozjs/src/gc/root.rs
sagudev, line 1: Should this be in this PR? I though we will put this in servo.

mozjs-sys/src/jsgc.rs
```diff
@@ -324,6 +324,7 @@ pub type RootedValueArray<const N: usize> = Rooted<ValueArray<N>>;
 /// SpiderMonkey.
 ///
 /// For safe `Heap` construction with value see `Heap::boxed` function.
+#[cfg_attr(feature = "crown", crown::unrooted_must_root_lint::must_root)]
```
Taym95, line 327: @jdm this is cause problem in typedarray's codegen in Servo:
```
Expression of type js::typedarray::TypedArray<js::typedarray::ArrayBufferU8, std::boxed::Box<mozjs_sys::jsgc::Heap<*mut js::jsapi::JSObject>, std::alloc::Global>> must be rooted.
```

python code:
```
    if is_typed_array(type):
        if failureCode is None:
            unwrapFailureCode = (f'throw_type_error(*cx, "{sourceDescription} is not a typed array.");\n'
                                 f'{exceptionCode}')
        else:
            unwrapFailureCode = failureCode

        typeName = type.unroll().name  # unroll because it may be nullable

        if isMember == "Union":
            typeName = f"Heap{typeName}"

        templateBody = fill(
            """
            match typedarray::${ty}::from($${val}.get().to_object()) {
                Ok(val) => val,
                Err(()) => {
                    $*{failureCode}
                }
            }
            """,
            ty=typeName,
            failureCode=f"{unwrapFailureCode}\n",
        )

        if isMember == "Union":
            templateBody = f"RootedTraceableBox::new({templateBody})"
```

genersted code:
```
impl ArrayBufferOrArrayBufferView {
    unsafe fn TryConvertToArrayBuffer(cx: SafeJSContext, value: HandleValue) -> Result<Option<RootedTraceableBox<typedarray::HeapArrayBuffer>>, ()> {
        Ok(Some(RootedTraceableBox::new(match typedarray::HeapArrayBuffer::from(value.get().to_object()) {
            Ok(val) => val,
            Err(()) => {
                return Ok(None);
            }
        }
        )))
    }

    unsafe fn TryConvertToArrayBufferView(cx: SafeJSContext, value: HandleValue) -> Result<Option<RootedTraceableBox<typedarray::HeapArrayBufferView>>, ()> {
        Ok(Some(RootedTraceableBox::new(match typedarray::HeapArrayBufferView::from(value.get().to_object()) {
            Ok(val) => val,
            Err(()) => {
                return Ok(None);
            }
        }
        )))
    }
}
```

error:
```
Building `release` build with crown enabled.
   Compiling script_bindings v0.0.1 (servo/components/script_bindings)
error: Expression of type js::typedarray::TypedArray<js::typedarray::ArrayBufferU8, std::boxed::Box<mozjs_sys::jsgc::Heap<*mut js::jsapi::JSObject>, std::alloc::Global>> must be rooted.
   --> servo/target/release/build/script_bindings-38ffb9456b24b1a2/out/GenericUnionTypes.rs:147:16
    |
147 |             Ok(val) => val,
    |                ^^^
    |
    = note: `#[deny(crown::unrooted_must_root)]` on by default
   ```
I could ignore it but I really want to knew what is happening I tried several solutions no success.

mozjs/src/gc/root.rs
Taym95: Heap is defined here, do you want add crown to Heap in servo?

mozjs-sys/src/jsgc.rs
jdm: I wonder if `typedarray::HeapArrayBuffer::from(value.get().to_object()).map(RootedTraceableBox::new)` would pass the crown checks?

Taym95: I played with it a bit. Since only the Heap version implements JSTraceable and can safely be rooted, only members where isMember == "Union" can use .map(RootedTraceableBox::new). non union members are not rooted, so I had to add a small if/else condition.

I got that working, but now I’m dealing with another error in the Servo code, where Heap<JSVal> is used without proper rooting.

mozjs/src/gc/root.rs
sagudev: I am just wondering why do we need this trait here, if we can have it in servo, per https://github.com/servo/mozjs/pull/592#issuecomment-3066599372

Because some users of mozjs are not using crown.

Taym95: Ah yes trait can be move to servo indeed

sagudev (approved):

PR #595, opened 2025-07-22
Implement saveJobQueue() for the SpiderMonkey Debugger API

delan:
in the [SpiderMonkey Debugger API](https://firefox-source-docs.mozilla.org/js/Debugger/), hooks like [onNewGlobalObject()](https://firefox-source-docs.mozilla.org/js/Debugger/Debugger.html#onnewglobalobject-global) use an AutoDebuggerJobQueueInterruption to [switch to a new microtask queue](https://github.com/servo/mozjs/blob/b14aebff23ac4d5b0652060ef949334bda08b22f/mozjs-sys/mozjs/js/src/debugger/Debugger.cpp#L2834-L2841) and avoid clobbering the debuggee’s microtask queue. this in turn relies on JobQueue::saveJobQueue(), which is [not yet implemented in RustJobQueue](https://github.com/servo/mozjs/blob/b14aebff23ac4d5b0652060ef949334bda08b22f/mozjs-sys/src/jsglue.cpp#L83-L86).

this patch implements [saveJobQueue() and SavedJobQueue](https://github.com/servo/mozjs/blob/b14aebff23ac4d5b0652060ef949334bda08b22f/mozjs-sys/mozjs/js/public/Promise.h#L92-L114) for RustJobQueue by calling into Servo via two new JobQueueTraps that create and destroy extra “interrupt” queues for use by the debugger.

SpiderMonkey [does not own any job queues](https://github.com/servo/mozjs/blob/b14aebff23ac4d5b0652060ef949334bda08b22f/mozjs-sys/mozjs/js/public/Promise.h#L117-L123), so the lifetime of these queues is managed in Servo, where they are stored in a Vec-based stack. stack-like behaviour is adequate for SpiderMonkey’s save and restore patterns, as far as we can tell, but we’ve added an assertion just in case.

Servo patch: servo/servo#38232

### DIFF
--- a/mozjs-sys/src/jsglue.cpp
+++ b/mozjs-sys/src/jsglue.cpp
@@ -49,7 +49,11 @@ struct JobQueueTraps {
                             JS::HandleObject allocationSite,
                             JS::HandleObject incumbentGlobal) = 0;
   bool (*empty)(const void* queue);
+
+  // Create a new queue, push it onto an embedder-side stack, and return the new queue.
   void* (*pushNewInterruptQueue)(void* aInterruptQueues);
+  // Destroy the queue most recently created by pushNewInterruptQueue(), returning its address
+  // so we can check if we are restoring the correct queue.
   void* (*popInterruptQueue)(void* aInterruptQueues);
 };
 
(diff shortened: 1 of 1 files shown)


mozjs-sys/src/jsglue.cpp
```diff
@@ -80,9 +90,85 @@ class RustJobQueue : public JS::JobQueue {
...
+          mInterruptQueues(aInterruptQueues),
+          mCurrentQueue(aCurrentQueue),
+          mNewQueue(aNewQueue),
+          mSavedQueue(*aCurrentQueue) {
+      // TODO: assert that the context’s jobQueue hasn’t been cleared with
+      // SetJobQueue(nullptr) or DestroyContext(). Don’t know how to do this
+      // with only an opaque JSContext decl. Are we allowed to #include
+      // "vm/JSContext.h"?
+      //
+      // MOZ_ASSERT(cx->jobQueue.ref());
```
jdm, line 111: I don't think this comment makes any sense here. It looks like you're using https://searchfox.org/mozilla-central/rev/1838f847aa3bf909c3d34a94a8f0cd7e37fca086/js/src/vm/JSContext.cpp#926 as a reference, but that's internal engine code.

mozjs-sys/src/jsglue.cpp
```diff
@@ -80,9 +90,85 @@ class RustJobQueue : public JS::JobQueue {
...
+      *mCurrentQueue = mNewQueue;
+    }
+
+    ~SavedQueue() {
+      // TODO: assert that the context’s jobQueue hasn’t been cleared with
+      // SetJobQueue(nullptr) or DestroyContext(). Don’t know how to do this
+      // with only an opaque JSContext decl. Are we allowed to #include
+      // "vm/JSContext.h"?
+      //
+      // MOZ_ASSERT(cx->jobQueue.ref());
```
jdm, line 125: Same here.

mozjs-sys/src/jsglue.cpp
```diff
@@ -80,9 +90,85 @@ class RustJobQueue : public JS::JobQueue {
...
+      // (SavedQueue from C to B). If the SavedQueue from B to A is restored
+      // before the SavedQueue from C to B, the embedder will destroy both C and
+      // B, but in the end, the queue will be set to B, a freed queue.
+      MOZ_ASSERT(mTraps.popInterruptQueue(mInterruptQueues) == mNewQueue);
+
+      *mCurrentQueue = mSavedQueue;
+    }
+
+   private:
+    JSContext* cx;
```
jdm: Why do we need to store the context?

mozjs-sys/src/jsglue.cpp
```diff
@@ -80,9 +90,85 @@ class RustJobQueue : public JS::JobQueue {
...
+    auto newQueue = mTraps.pushNewInterruptQueue(mInterruptQueues);
+    auto result = js::MakeUnique<SavedQueue>(cx, mTraps, mInterruptQueues,
+                                             &mQueue, newQueue);
+    if (!result) {
+      // “On OOM, this should call JS_ReportOutOfMemory on the given JSContext,
+      // and return a null UniquePtr.” When the allocation in MakeUnique()
+      // fails, the SavedQueue constructor is never called, so this->mQueue is
+      // still set to the old queue.
+      // TODO: should we propagate any OOM condition from Servo by returning
+      // nullptr in pushNewInterruptQueue()?
```
jdm: In general we have handled OOM in SpiderMonkey APIs by aggressively asserting non-null allocation responses, and we rarely use fallible allocation in Rust code in Servo.

mozjs-sys/src/jsglue.cpp
```diff
@@ -49,15 +49,25 @@ struct JobQueueTraps {
...
                             JS::HandleObject incumbentGlobal) = 0;
   bool (*empty)(const void* queue);
+
+  // Create a new queue, push it onto an embedder-side stack, and return the new
+  // queue.
+  void* (*pushNewInterruptQueue)(void* aInterruptQueues);
+  // Destroy the queue most recently created by pushNewInterruptQueue(),
+  // returning its address so we can check if we are restoring the saved queue
+  // over the correct queue.
+  void* (*popInterruptQueue)(void* aInterruptQueues);
```
jdm: Can these traps return `const void*` instead? I think that would help the Rc questions in the servo PR.

mozjs-sys/src/jsglue.cpp
delan: we did base our impl on the InternalJobQueue, that’s true. in the case of InternalJobQueue, they assert that `cx->internalJobQueue` (the SpiderMonkey-owned internal job queue) is not null, which might happen if UseInternalJobQueues() was not called.

there’s no equivalent to that in RustJobQueue, but it did make us aware that `cx->jobQueue` (the borrowed job queue) may be null, if it was cleared with SetJobQueue(nullptr) or DestroyContext(). we wanted to check for that situation, but we couldn’t, because the fields of JSContext are not public api.

mozjs-sys/src/jsglue.cpp
delan: left over from those assertions; removed in 3a06b6c2ba822dcfc1b56ab1567076a9ce1794bf

mozjs-sys/src/jsglue.cpp
delan: amended in 9399a9d26c738ba72fc211e5ac38a2dd601a7290

mozjs-sys/src/jsglue.cpp
delan: i think so! done in 22547bef60767b31fb71d86edcd67570e6ea724d

delan:
we think we’ve addressed your feedback, and we’ve also pushed ccc1423c5dedd86c34c74b7eda7d62be1b26b7ee, which helps Servo avoid a memory leak when doing the Box::new → Box::into_raw → (Box::from_raw → forget)+ dance.

jdm (approved):

sagudev:
This PR should have bumped mozjs-sys version.

PR #596, opened 2025-07-24
Use cargo provided icu_capi

jschwe:
Using vendored rust crates causes problems with `cargo vendor` and offline builds as well as cargo publish. 
Using `cargo metadata` is currently the officially recommended solution to get the location of the c header files, but ideally we add a patch upstream to support reading the location from a `DEP_` environment variable, that cargo could set for us (posted https://github.com/unicode-org/icu4x/issues/6769).

For ease of reviewing the changes to the build system and removing the now unnecessary vendored code are separate commits.

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -213,7 +213,7 @@ fn build_spidermonkey(build_dir: &Path) {
 
     let icu_c_include_path = get_icu_capi_include_path();
     let mut cxxflags = vec![];
-    cxxflags.push(format!("-I{}", &icu_c_include_path));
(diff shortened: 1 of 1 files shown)


jschwe:
@sagudev @jdm Does either of you know if I can set includes in moz.build, that point to somewhere in `$CARGO_HOME`? Currently I'm patching out the `LOCAL_INCLUDES` line, and setting `-I<path>` via `CXXFLAGS`, but that doesn't seem to work on windows (probably I need to escape the `\` component in the path or similar). But I guess it would be nicer to use `moz.build` instead, so I'm wondering if there is a way to set INCLUDES to an arbitrary path.

sagudev (approved):

PR #598, opened 2025-07-25
Expose ScriptEnvironmentPreparer to Rust

delan:
# NOTE: do not merge until #597 lands

### DIFF
--- a/mozjs-sys/src/jsglue.cpp
+++ b/mozjs-sys/src/jsglue.cpp
@@ -173,6 +173,21 @@ class RustJobQueue : public JS::JobQueue {
   }
 };
 
+struct RustEnvironmentPreparer : public js::ScriptEnvironmentPreparer {
+  JSContext* cx;
+  explicit RustEnvironmentPreparer(JSContext* cx) : cx(cx) {}
+  void invoke(JS::HandleObject global, Closure& closure) override {
+    MOZ_ASSERT(JS_IsGlobalObject(global));
+    MOZ_ASSERT(!JS_IsExceptionPending(cx));
+
+    JSAutoRealm ar(cx, global);
+    // TODO: AutoReportException are(cx);
+    if (!closure(cx)) {
+      return;
+    }
+  }
+};
+
 struct JSExternalStringCallbacksTraps {
   void (*latin1Finalize)(const void* privateData, JS::Latin1Char* chars);
   void (*utf16Finalize)(const void* privateData, char16_t* chars);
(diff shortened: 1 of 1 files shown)


mozjs-sys/src/jsglue.cpp
```diff
@@ -173,6 +173,21 @@ class RustJobQueue : public JS::JobQueue {
...
 };
 
+struct RustEnvironmentPreparer : public js::ScriptEnvironmentPreparer {
+  JSContext* cx;
+  explicit RustEnvironmentPreparer(JSContext* cx) : cx(cx) {}
+  void invoke(JS::HandleObject global, Closure& closure) override {
+    MOZ_ASSERT(JS_IsGlobalObject(global));
+    MOZ_ASSERT(!JS_IsExceptionPending(cx));
+
+    JSAutoRealm ar(cx, global);
```
jdm, line 183: This looks like you're modelling it off of the jsshell version, but I intend to use this to replicate https://searchfox.org/mozilla-central/rev/fcc0cfc7f5b1df957ae29f6b2ccfeb6df47b6be8/xpcom/base/CycleCollectedJSRuntime.cpp#2177 . Since AutoEntryScript is a Servo Rust type, can we make this class implement a hook for Rust code to invoke the closure instead?

delan: > I intend to use this to replicate https://searchfox.org/mozilla-central/rev/fcc0cfc7f5b1df957ae29f6b2ccfeb6df47b6be8/xpcom/base/CycleCollectedJSRuntime.cpp#2177

can you elaborate on this? i’m not sure i follow, did you have existing plans to set up a ScriptEnvironmentPreparer for Servo?

> Since AutoEntryScript is a Servo Rust type, can we make this class implement a hook for Rust code to invoke the closure instead?

yeah, that could work. would the idea be that RustEnvironmentPreparer::invoke() just passes the closure to Servo, which would enter the AutoEntryScript and call the closure?

also, what do we do with uncaught exceptions? the jsshell uses a thing called AutoReportException, but i’m not sure what the Servo counterpart is for that.

delan:
heads up: this is a lower priority for us now, since it only seems to be needed for handling uncaught exceptions in the debugger script (#38331).

mozjs-sys/src/jsglue.cpp
jdm: >can you elaborate on this? i’m not sure i follow, did you have existing plans to set up a ScriptEnvironmentPreparer for Servo?

I didn't know it existed until this PR, but now I want to replicate the way Gecko uses it.

>yeah, that could work. would the idea be that RustEnvironmentPreparer::invoke() just passes the closure to Servo, which would enter the AutoEntryScript and call the closure

Yes.

>also, what do we do with uncaught exceptions? the jsshell uses a thing called AutoReportException, but i’m not sure what the Servo counterpart is for that.

We have code that calls report_pending_exception in various places right now. Gecko uses AutoJsApi, which is another thing I want to replicate.

Gae24:
Closing this since it was superseded by #704

PR #599, opened 2025-07-27
Implement JSAutoStructuredCloneBufferWrapper

kkoyung:
Implement a safe wrapper of `*mut JSAutoStructuredCloneBuffer` that Implements the Drop trait to prevent leakage.

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -48,10 +48,7 @@ use crate::jsapi::{InitSelfHostedCode, IsWindowSlow};
 use crate::jsapi::{
     JSAutoRealm, JS_SetGCParameter, JS_SetNativeStackQuota, JS_WrapObject, JS_WrapValue,
 };
-use crate::jsapi::{
-    JSAutoStructuredCloneBuffer, JSStructuredCloneCallbacks, JSStructuredCloneData,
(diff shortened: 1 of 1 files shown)


sagudev (approved):

mozjs/src/rust.rs
```diff
@@ -532,6 +537,29 @@ impl Drop for CompileOptionsWrapper {
     }
 }
 
+pub struct JSAutoStructuredCloneBufferWrapper {
+    pub ptr: *mut JSAutoStructuredCloneBuffer,
```
sagudev: I think we could make wrap this in nonnull.

kkoyung: How about adding `assert!(!ptr.is_null());`, like `CompileOptionsWrapper`?

sagudev: I mean it would still be nice to tell the compiler that wrapped pointer is never null, by wrapping it in nonull.

kkoyung: Just suggesting an alternative approach that's consistent with the existing code.
Anyway, the internal pointer is now wrapped in NonNull in the new commit.

sagudev (approved):

PR #602, opened 2025-07-28
Add check for mozjs-sys bumps

### LINKED ISSUE #601: Gate pull requests on mozjs-sys bump if there are any C++ changes
in #595, we made Servo-breaking changes in mozjs-sys without bumping the mozjs-sys crate version, which meant that no new artifact was released for prebuilt mozjs. this broke Servo in servo/servo#38311.

we should prevent pull requests from merging if they have any changes (or at least any C++ changes?) in mozjs-sys without bumping the version. this can be done in build.yml’s integrity job.

([zulip thread](https://servo.zulipchat.com/#narrow/channel/263398-general/topic/Building.20new.20mozjs-sys.20artifacts/near/531305693), cc @sagudev)

sagudev:
Fixes #601 

To avoid problems we should also enable "Require branches to be up to date before merging" in GitHub for this repo: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging

Testing:
- no mozjs_sys changes: this PRs checks
- mozjs_sys changes without bump: https://github.com/servo/mozjs/actions/runs/16565592710/job/46844809943
- mozjs_sys changes with bump: https://github.com/servo/mozjs/actions/runs/16565626054/job/46844920211

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -248,6 +248,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
       - uses: dtolnay/rust-toolchain@stable
       - name: Check Rust formatting
         run: cargo fmt --check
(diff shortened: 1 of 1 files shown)


sagudev:
> how come only C++ changes require a version bump? do the artifacts not contain any of the rust parts of mozjs-sys?

Artifacts contain compiled SM (so all C/C++ code), all glues (src/*.cpp) and rust bindings of them, hence I also included mozjs-sys/* in filter to catch changes to build.rs (defines how artifacts are packed) and Cargo.toml (defines bindgen version).

PR #603, opened 2025-07-30
Allow embedder to indicate whether scripts are inline

delan:
to use the [SpiderMonkey Debugger API](https://firefox-source-docs.mozilla.org/js/Debugger/) as the single source of truth about scripts and their sources for devtools purposes (servo/servo#38334), the debugger script needs to be able to distinguish inline scripts from other scripts, because inline scripts are a special case where the source contents need to come from the Servo parser.

the mechanism for this is [Debugger.Script.prototype.**introductionType**](https://firefox-source-docs.mozilla.org/js/Debugger/Debugger.Source.html#introductiontype), which is `inlineScript` for inline scripts or a variety of other values for other kinds of scripts, but only the embedder can provide this information.

this patch expands on CompileOptionsWrapper, making it a safe wrapper around CompileOptions. to construct one from safe code, use Runtime::new_compile_options(). then you can call `set_introduction_type(&'static CStr)` on the new instance. we also make Runtime::evaluate_script() take a CompileOptionsWrapper from the caller, instead of constructing one internally.

Servo patch: servo/servo#38363
Fixes: part of servo/servo#38378

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -7,8 +7,8 @@
 use std::cell::Cell;
 use std::char;
 use std::default::Default;
-use std::ffi;
 use std::ffi::CStr;
+use std::ffi::{self, CString};
 use std::marker::PhantomData;
 use std::mem::MaybeUninit;
 use std::ops::{Deref, DerefMut};
@@ -399,13 +399,13 @@ impl Runtime {
         &self,
         glob: HandleObject,
         script: &str,
-        filename: &str,
         rval: MutableHandleValue,
         options: CompileOptionsWrapper,
     ) -> Result<(), ()> {
         debug!(
             "Evaluating script from {} with content {}",
-            filename, script
+            options.filename(),
+            script
         );
 
         let _ac = JSAutoRealm::new(self.cx(), glob.get());
@@ -521,6 +521,7 @@ impl Drop for RootedObjectVectorWrapper {
 
 pub struct CompileOptionsWrapper {
     pub ptr: *mut ReadOnlyCompileOptions,
+    filename: CString,
 }
 
 impl CompileOptionsWrapper {
@@ -528,10 +529,14 @@ impl CompileOptionsWrapper {
     /// `cx` must point to a non-null, valid [`JSContext`].
     /// To create an instance from safe code, use [`Runtime::new_compile_options`].
     pub unsafe fn new(cx: *mut JSContext, filename: &str, line: u32) -> Self {
-        let filename_cstr = ffi::CString::new(filename.as_bytes()).unwrap();
-        let ptr = NewCompileOptions(cx, filename_cstr.as_ptr(), line);
+        let filename = CString::new(filename.as_bytes()).unwrap();
+        let ptr = NewCompileOptions(cx, filename.as_ptr(), line);
         assert!(!ptr.is_null());
-        Self { ptr }
+        Self { ptr, filename }
+    }
+
+    pub fn filename(&self) -> &str {
+        self.filename.to_str().expect("Guaranteed by new")
     }
 
     pub fn set_introduction_type(&mut self, introduction_type: &'static CStr) {
--- a/mozjs/examples/eval.rs
+++ b/mozjs/examples/eval.rs
@@ -45,13 +45,7 @@ fn run(rt: Runtime) {
     let source: &'static str = "40 + 2";
 
     let options = rt.new_compile_options(filename, lineno);
-    let res = rt.evaluate_script(
-        global.handle(),
-        source,
-        filename,
-        rval.handle_mut(),
-        options,
-    );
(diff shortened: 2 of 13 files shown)


mozjs-sys/src/jsglue.cpp
```diff
@@ -692,10 +692,26 @@ void DeleteCompileOptions(JS::ReadOnlyCompileOptions* aOpts) {
...
-JS::ReadOnlyCompileOptions* NewCompileOptions(JSContext* aCx, const char* aFile,
-                                              unsigned aLine) {
+enum class IntroductionType : unsigned {
+  Undefined,
+  InlineScript,
+};
+
+JS::ReadOnlyCompileOptions* NewCompileOptions(
+    JSContext* aCx, const char* aFile, unsigned aLine,
+    IntroductionType introductionType) {
```
sagudev:
```suggestion
    const char* introductionType) {
```
I think we should just use here `const char*` thus let user of mozjs define custom values, because we will likely need more of them in servo eventually. And it also reduces amount of C++ code.

Also isn't this already available here: https://doc.servo.org/mozjs/jsapi/struct.TransitiveCompileOptions.html#structfield.introductionType

We could still use safe wrappers accepting `&'static CStr`.

delan: > I think we should just use here `const char*` thus let user of mozjs define custom values, because we will likely need more of them in servo eventually. And it also reduces amount of C++ code. […] We could still use safe wrappers accepting `&'static CStr`.

yeah, we considered doing that initially, but NewCompileOptions() and Runtime::evaluate_script() are not marked unsafe, and the static C string invariant would need to be proven by the caller. i guess like you said, we could make this unsafe and create a safe wrapper around it.

> Also isn't this already available here: https://doc.servo.org/mozjs/jsapi/struct.TransitiveCompileOptions.html#structfield.introductionType

hmm, good point, i didn’t realise we could set the field manually. i’ll try it :)

sagudev: > NewCompileOptions() and Runtime::evaluate_script() are not marked unsafe, and the static C string invariant would need to be proven by the caller

[NewCompileOptions](https://doc.servo.org/mozjs/glue/fn.NewCompileOptions.html) is unsafe because it's from FFI, but Runtime::evaluate_script() isn't, but that one is already safe wrapper so we could just add `&'static CStr`.

sagudev: actually given https://github.com/servo/servo/issues/38378 we should write safe wrapper for compile options and let them pass into evaluate_script.

delan: it looks like [CompileOptionsWrapper](https://doc.servo.org/mozjs/rust/struct.CompileOptionsWrapper.html) is pretty much what we need; we were just missing a safe way to construct new instances without the unchecked JSContext pointer.

249ac54367fbdb247444a87d46f48601d3df6b06 removes IntroductionType, reverts the changes to NewCompileOptions() and CompileOptionsWrapper, makes Runtime::evaluate_script() take &amp;'static CStr (superseded by 4b431ccb270adfeefb98b8a6e55cc73c541e5e44), and adds CompileOptionsWrapper::set_introduction_type().

4b431ccb270adfeefb98b8a6e55cc73c541e5e44 makes Runtime::evaluate_script() take CompileOptionsWrapper instead of constructing one internally from the given &amp;'static CStr.

64604b0561fc0617e2a1ca85dcd98eaf9845791c adds Runtime::new_compile_options(), a safe way to construct a CompileOptionsWrapper. we had to guess the unsafe invariants in CompileOptionsWrapper::new(), so please let us know if we missed any.

c8ecb34b215684a0cf59cbfaa92a77a713549c66 fixes the tests.

mozjs/src/rust.rs
```diff
@@ -400,16 +400,15 @@ impl Runtime {
...
         script: &str,
         filename: &str,
-        line_num: u32,
         rval: MutableHandleValue,
+        options: CompileOptionsWrapper,
     ) -> Result<(), ()> {
         debug!(
             "Evaluating script from {} with content {}",
             filename, script
         );
```
sagudev: the only reason why we pass filename here is this debug statement, can we just extract filename from `CompileOptionsWrapper`?

delan: unfortunately it’s kinda annoying to get the filename out of CompileOptionsWrapper, because there we convert the &amp;str to CString. not sure what’s the best way to deal with this :/

sagudev: I think just doing `CStr::from_ptr(...).to_string_lossy()` would be enough.

delan: done in b6d7bce2bc165fb555ce3f254404aecc89627388

delan:
thanks for the feedback!

PR #606, opened 2025-08-08
fix icu_capi version

jschwe:
The vendored version of icu_capi was bumped to 1.5.0 in the SM 137
 update, so we need to update the version in the Cargo.toml too.
Additionally, we use `=1.5.0` to force selecting the same version as the vendored one.
The vendored icu_capi is never built, only its headers are used.


jschwe:
#596 presumably is blocked until we get the integrity test back, but I think this change can be merged earlier.

sagudev (approved):
While this change technically does not need to bump mozjs-sys version, but CI isn't smart enough to know, so we will need to do it anyway to make it happy.

PR #608, opened 2025-08-13
Remove unneeded icu_capi features

jschwe:
We don't need all default features, which are quite a few: https://github.com/unicode-org/icu4x/blob/599f3366f901eb0cc0af9de6c9de99998734bb8a/ffi/capi/Cargo.toml#L38
This reduces servo binary size by around 15% in production mode.

Companion PR: https://github.com/servo/servo/pull/38666

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -36,7 +36,7 @@ encoding_c = "0.9.8"
 encoding_c_mem = "0.2.6"
 # unicode-bidi-ffi = { path = "./mozjs/intl/bidi/rust/unicode-bidi-ffi" }
 # keep in sync with intl/icu_capi/Cargo.toml
-icu_capi = { version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
+icu_capi = { version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_calendar", "icu_properties", "icu_segmenter"] }
 
 [build-dependencies]
 bindgen.workspace = true
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -35,7 +35,8 @@ libz-sys = "1.1.19"
 encoding_c = "0.9.8"
 encoding_c_mem = "0.2.6"
 # unicode-bidi-ffi = { path = "./mozjs/intl/bidi/rust/unicode-bidi-ffi" }
-icu_capi = "=1.5.0" # keep in sync with intl/icu_capi/Cargo.toml
+# keep in sync with intl/icu_capi/Cargo.toml
+icu_capi = {  version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
 
 [build-dependencies]
 bindgen.workspace = true
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.137.0-1"
+version = "0.137.0-2"
 authors = ["Mozilla"]
 links = "mozjs"
 license.workspace = true
@@ -36,7 +36,7 @@ encoding_c = "0.9.8"
 encoding_c_mem = "0.2.6"
 # unicode-bidi-ffi = { path = "./mozjs/intl/bidi/rust/unicode-bidi-ffi" }
 # keep in sync with intl/icu_capi/Cargo.toml
-icu_capi = {  version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
+icu_capi = { version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
 
 [build-dependencies]
 bindgen.workspace = true
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -385,6 +385,7 @@ fn link_static_lib_binaries(build_dir: &Path) {
         println!("cargo:rustc-link-lib=psapi");
         println!("cargo:rustc-link-lib=user32");
         println!("cargo:rustc-link-lib=Dbghelp");
+        println!("cargo:rustc-link-lib=advapi32");
         if target.contains("gnu") {
             println!("cargo:rustc-link-lib=stdc++");
         }


jdm (approved):
Very nice find!

mozjs-sys/Cargo.toml
```diff
@@ -35,7 +35,8 @@ libz-sys = "1.1.19"
 encoding_c = "0.9.8"
 encoding_c_mem = "0.2.6"
 # unicode-bidi-ffi = { path = "./mozjs/intl/bidi/rust/unicode-bidi-ffi" }
-icu_capi = "=1.5.0" # keep in sync with intl/icu_capi/Cargo.toml
+# keep in sync with intl/icu_capi/Cargo.toml
+icu_capi = {  version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
```
sagudev:
```suggestion
icu_capi = { version = "=1.5.0", default-features = false, features = ["compiled_data", "icu_segmenter"] }
```

How did you determine the features needed? Because from what I see firefox keeps all default (`["compiled_data", "default_components", "logging", "simple_logger", "std"]`).

jschwe: https://github.com/mozilla-firefox/firefox/blob/esr128/intl/icu4x-patches/001-Cargo.toml.patch

Firefox applies a patch to reduce the `compiled_data`, and we tried if we can achieve the same by just removing components. We need `icu_segmenter` and the data it references, and that seems to be sufficient for mozjs to compile and servo to pass all wpt tests (linked a try run in the companion PR). 
Do you think keeping the logging feature might be valuable?

sagudev: No, but I think this might be reason for compilation problems. I suspect we need at least std and all features enabled here: https://github.com/servo/mozjs/blob/0bebd6872b1a061b69bcfcf73162abf0af27eebc/mozjs-sys/mozjs/intl/icu_capi/Cargo.toml#L192 (this is from SM 137 not 128).

sagudev:
```suggestion
icu_capi = { version = "=1.5.0", default-features = false, features = ["std", "compiled_data", "icu_calendar", "icu_properties", "icu_segmenter"] }
```

jschwe: Downgrading the compiler does fix the problem on windows. Looking at the undefined functions, it seems that the problem should be fixed by linking`Advapi32.dll`. That file does appear in the [mozglue/moz.build](https://github.com/servo/mozjs/blob/0bebd6872b1a061b69bcfcf73162abf0af27eebc/mozjs-sys/mozjs/mozglue/build/moz.build#L35), but I guess the linker is invoked via rust, so we need to add a line to our build-script. 

Edit: my suspicion is that we were relying on unused code / rust stdlib linking in that library.

sagudev: I still think we need "icu_calendar", "icu_properties" as they might not be needed for servo because of feature unification.

jschwe: That's a good point. I'll add them.

jschwe: Wouldn't it still cause link issues when building the mozjs tests, if we didn't add a required feature?

sagudev: Hm, depending on how icu works, it could have just one entrypoint function (this one is linked) and features enable it's inner functionality.

jschwe:
What is a bit worrying, is that while the ohos binary size even dropped by 22MB, the linux binary size remains unchanged [bencher](https://bencher.dev/console/projects/servo/perf?lower_value=false&upper_value=false&lower_boundary=false&upper_boundary=false&x_axis=date_time&branches=52e1e9bb-959c-4171-a53d-e06bd694a6c1&testbeds=d742c702-3842-4108-9d0c-2db74e57599a&benchmarks=864410ba-a36c-4e5e-a822-64d828f89160%2C46c23cda-549d-4ffc-88ae-0c5298aab5b9&measures=678e4118-c8a5-494d-8799-08abc3021cd5&start_time=1752817255121&end_time=1755236455121&tab=plots&plot=9b77de8b-0330-44cb-805c-4cbf8743b115&plots_search=9b77de8b-0330-44cb-805c-4cbf8743b115&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=) - and we only run WPT on Linux - I hope this didn't break anything on ohos ...
Time to investigate feature unification differences on the linux version vs OH I guess.

sagudev:
> What is a bit worrying, is that while the ohos binary size even dropped by 22MB, the linux binary size remains unchanged [bencher](https://bencher.dev/console/projects/servo/perf?lower_value=false&upper_value=false&lower_boundary=false&upper_boundary=false&x_axis=date_time&branches=52e1e9bb-959c-4171-a53d-e06bd694a6c1&testbeds=d742c702-3842-4108-9d0c-2db74e57599a&benchmarks=864410ba-a36c-4e5e-a822-64d828f89160%2C46c23cda-549d-4ffc-88ae-0c5298aab5b9&measures=678e4118-c8a5-494d-8799-08abc3021cd5&start_time=1752817255121&end_time=1755236455121&tab=plots&plot=9b77de8b-0330-44cb-805c-4cbf8743b115&plots_search=9b77de8b-0330-44cb-805c-4cbf8743b115&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=) - and we only run WPT on Linux - I hope this didn't break anything on ohos ... Time to investigate feature unification differences on the linux version vs OH I guess.

https://doc.rust-lang.org/cargo/reference/features.html#inspecting-resolved-features

jschwe:
```
#!/usr/bin/env bash


for crate_name in $(cargo tree | grep icu_ |  sed -e 's/.*\(icu_.*\)v[0-9]\..*/\1/' | sort | uniq)
do
  echo "Comparing features for crate $crate_name"
  linux_tree=$(cargo tree -i "${crate_name}" -e features --depth 2 --target x86_64-unknown-linux-gnu)
  ohos_tree=$(cargo tree -i "${crate_name}" -e features --depth 2 --target aarch64-unknown-linux-ohos)
  diff <(echo "$linux_tree") <(echo "$ohos_tree")
done
```

I created the following small script to compare the enabled features of the icu crates and this doesn't show any differences (I also manually diffed the complete output of `cargo tree` first, which also didn't show anything. 
The binary size is completely unchanged on windows, linux and mac. Android also has a similar [binary size reduction](https://bencher.dev/console/projects/servo/perf?clear=true&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=95&testbeds_page=1&benchmarks_page=1&plots_page=1&measures=678e4118-c8a5-494d-8799-08abc3021cd5&tab=benchmarks&branches_search=main&branches=52e1e9bb-959c-4171-a53d-e06bd694a6c1&heads=3dbe3681-11b1-4e30-b482-4ee72dc0960c&testbeds=d742c702-3842-4108-9d0c-2db74e57599a%2Cce4dc48a-8ae7-44e7-94de-f4815a7dfdad&benchmarks=42b8e664-90ce-434b-a11a-a41d1cdc0932%2C588aed8a-7309-4c7d-963c-462b90e47f40%2C09e53e6a-d783-4033-afe2-05a917dc5188&benchmarks_search=-release) (we only test release mode binary size for android though).

I've always found it a bit strange that the ohos shared library was so much larger than the linux binary - I wonder if the difference might just be that the linker already stripped out the unneeded / unused parts on the major desktop platforms? 
It's really a pity we can't run wpt yet on mobile...

PR #610, opened 2025-08-14
Remove unneeded icu_capi features 

jschwe:
We don't need all default features, which are quite a few: https://github.com/unicode-org/icu4x/blob/599f3366f901eb0cc0af9de6c9de99998734bb8a/ffi/capi/Cargo.toml#L38
This reduces servo binary size by around 15% in production mode.

Companion PR: https://github.com/servo/servo/pull/38666

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,7 @@ jobs:
         features: ["debugmozjs", '""']
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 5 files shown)


jschwe:
Sames as #608, but pushed to a branch in this repo, so hopefully the .yml changes will have an effect.

sagudev:
> Sames as #608, but pushed to a branch in this repo, so hopefully the .yml changes will have an effect.

Yaml files should always have effect (except in servo if they are triggered from try labels).

jschwe:
yeah, it seems I overlooked some occurrences of `@stable` the first time :'(

PR #611, opened 2025-08-14
Use Rust 1.85 toolchain

jschwe:
We should ensure mozjs doesn't break servos MSRV accidentally. The easiest way is to just consistently use servos MSRV as the compiler when developing mozjs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
         type: boolean
       rust_version:
         required: false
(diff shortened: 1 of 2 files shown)


sagudev:
Both mozangle and mozjs have weekly runs to detect breaking changes in environment (either in runner images or in rust). Can we preserve this by adding input on https://github.com/servo/mozjs/blob/0bebd6872b1a061b69bcfcf73162abf0af27eebc/.github/workflows/build.yml#L5

sagudev (approved):
This is great, thanks.

PR #614, opened 2025-08-16
Replace null bytes in error text with escaped bytes.

jdm:
This is easy to trigger from code that throws a JS error and helpfully provides the original failed input (eg. `throw_js_error(cx, format!("\"{input}\" is not a valid ReadableStreamType value", 0)`). To avoid unnecessary allocations, we only perform the escaping if CString::new reports that the original string contains invalid bytes.


sagudev (approved):
LGTM, although I am not sure about replacement behavior (FF just removes nuls). It would be interesting to see what expects WPT tests.

PR #617, opened 2025-08-20
Prepare mozjs-sys for publishing

jschwe:
-  Update the publish exclude list to trim the size down
- Prevent pollution of the source directory, by disabling python byte cache generation and fixing makefile.cargo touching old-configure, which doesn't exist anymore.


jdm (approved):

sagudev:
We could probably also request for higher size limit: https://github.com/rust-lang/crates.io/issues/195, but we should reduce our package as much as possible before that.

I think we would need to patch build sys to not run SM vm ops tests (python script) to be able to remove more.

mozjs-sys/Cargo.toml
```diff
@@ -2,16 +2,24 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.140.0-2"
+version = "0.140.0-3"
 authors = ["Mozilla"]
```
jschwe: What about the authors fields? Should we add "The servo community" or just remove the field? I wouldn't want to create a false impression that this crate is maintained by mozilla when its not.

sagudev: We use "The Servo Project Developers": https://github.com/servo/servo/blob/9c840b05bc53502555ec5848a0c6b7434c78d650/Cargo.toml#L12C13-L12C41

mozjs-sys/Cargo.toml
```diff
@@ -2,16 +2,24 @@
...
+  "/mozjs/python/mozperftest/mozperftest/system/example.zip",
+  "/mozjs/python/mozperftest/mozperftest/tests/data",
+  "/mozjs/third_party/python/aiohttp",
+  # pip is 8MiB and we can assume its installed.
+  "/mozjs/third_party/python/pip",
+  "/mozjs/testing/mozharness",
+  "/mozjs/testing/mozbase/mozproxy",
+  "/mozjs/testing/mozbase/**/tests",
+  "/mozjs/testing/performance/",
+  "/mozjs/testing/test/",
```
sagudev: Why not add these to filters: https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/filters.txt

jschwe: I was thinking it might make it easier to tweak the excludes if we can easily inspect the contents of excluded folders locally. 
I'm not particulary opposed though - If you prefer, then we can use filters.txt instead.

sagudev: The idea is to make repository (checkout) more lightweight (thus also cargo checkout when using git) and diffs of SM bumps smaller. If anybody wants to inspect content of filtered files it can read SM archive. I also think it makes sense to have all SM filtering/excluding in one place.

jschwe: Does cargo checkout repositories using a shallow clone? I guess it does make sense to at least try to prevent the repository bloating further than it already has, but this won't reduce the size of standard checkouts, since everything is still in the history.

sagudev: > Does cargo checkout repositories using a shallow clone?

Unfortunately no, but we want to avoid future pollution.

PR #619, opened 2025-08-23
Add Cargo Feature for zlib-rs

### LINKED ISSUE #586: Add Cargo Feature to Use zlib-rs (instead of system zlib)
As of FF 137, `zlib-rs` is supported. By adding a feature flag for `zlib-rs`, it would reduce the amount of C compiled, and would reduce the number of system dependencies.

The `USE_LIBZ_RS` environment variable would be required, along with the `libz-rs-sys` crate (w/ the `export-symbols` feature). In theory, the `zlib.h` header should work, and cargo/linker should handle the rest. Given that `zlib-rs` doesn't provide headers, I'm not sure where we should get them from.

redfire75369:
Fixes #586

Added configure flag `--enable-libz-rs`.
Adds `zlib.h` and `zconf.h` from mozjs, to replace the system zlib header as `libz-rs-sys` does not provide its own.
Modifies some of the zlib patches in `0033-no-zlib.patch` to re-enable using zlib-rs.

Added two features for each of `mozjs` and `mozjs-sys`, `libz-sys` to either link to system `libz.so`, or compile `libz` from source, and `libz-rs` to use `libz-rs-sys` to mimic the C API of `libz.so`.
`mozjs` is given the default feature, `libz-sys`, which enables it for `mozjs-sys` as well.

The tarball distributed in releases is still compatible, and no changes are required for downloading tarballs for faster compilation, to work.
In a release build, the size penalty is only about 532 KiB.

<details>
<summary>Testing to show zlib-rs is compiled and linked properly.</summary>
All tests performed on WSL 2 Debian with `zlib1g-dev` installed.

```sh
$ cargo build --test evaluate
$ ldd target/debug/deps/evaluate-3d8e1e0dea36e1d8
target/debug/deps/evaluate-3d8e1e0dea36e1d8:
        linux-vdso.so.1 (0x00007ffdd9795000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x0000787f9c000000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x0000787f9e7a5000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x0000787f9e778000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000787f9e688000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000787f9be0a000)
        /lib64/ld-linux-x86-64.so.2 (0x0000787f9e7ce000)
```
```sh
$ cargo build --test evaluate --no-default-features --features libz-rs
$ ldd target/debug/deps/evaluate-55eb3881f68ef35b
target/debug/deps/evaluate-55eb3881f68ef35b:
        linux-vdso.so.1 (0x00007ffdae737000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f67bc800000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f67bef72000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f67bcb10000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f67bc60a000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f67befa8000)
```
</details>


mozjs-sys/etc/patches/0033-zlib.patch
```diff
@@ -1,28 +1,24 @@
...
- Library("zlib")
- 
--if CONFIG["MOZ_SYSTEM_ZLIB"]:
--    OS_LIBS += CONFIG["MOZ_ZLIB_LIBS"]
+ if CONFIG["MOZ_SYSTEM_ZLIB"]:
+     OS_LIBS += CONFIG["MOZ_ZLIB_LIBS"]
 -else:
 -    DIRS += [
 -        "../../../modules/zlib",
 -    ]
```
sagudev: I think we can write includes here and remove `if libz_rs {` from build.rs

mozjs-sys/build.rs
```diff
@@ -187,23 +187,43 @@ fn build_spidermonkey(build_dir: &Path) {
...
-            "{}/lib/pkgconfig",
-            zlib_root_dir.replace("\\", "/")
-        ));
-        if let Some(env_pkg_config_path) = get_cc_rs_env_os("PKG_CONFIG_PATH") {
-            pkg_config_path.push(":");
-            pkg_config_path.push(env_pkg_config_path);
+    let libz_rs = env::var_os("CARGO_FEATURE_LIBZ_RS").is_some();
+    let libz_sys = env::var_os("CARGO_FEATURE_LIBZ_SYS").is_some();
+
+    if libz_rs && libz_sys {
```
sagudev: Should we also fail if both are disabled?

sagudev: Also what happens if somebody else in dep tree brings libz-sys, linker errors presumably?

mozjs-sys/etc/patches/0033-zlib.patch
redfire75369: I considered that, but it'd need `modules/zlib/moz.build` and `modules/zlib/src/moz.build`, the latter of which I wasn't too sure about.

sagudev: can't we just do `INCLUDES +=` (or maybe it's `EXPORTS +=`) without any other build files?

mozjs-sys/build.rs
redfire75369: Without both features, the pkg-config paths aren't added, and it fails as it can't find zlib. If you have the relevant dev packages installed on your system, it fails at link time with "undefined reference" errors. 
We could try a single cargo feature instead, but that leads to a footgun if a something else in the tree brings it in with `libz-rs`, and you wanted `libz-sys`. (I guess that's kind of still a problem here, it just fails immediately though)

I've been testing. These are the results: GNU `ld` seems to just link in `libz-rs-sys` for all, and not link `libz.so` at all, `lld` and `mold` fail with "duplicate symbol" errors

mozjs-sys/etc/patches/0033-zlib.patch
redfire75369: `EXPORTS+=` does work.

```py
else:
    EXPORTS += [
        "../../../modules/zlib/src/zconf.h",
        "../../../modules/zlib/src/zlib.h",
    ]
```

mozjs-sys/build.rs
```diff
@@ -187,23 +187,32 @@ fn build_spidermonkey(build_dir: &Path) {
...
-    // this is only needed when libz-sys builds zlib from source
-    if let Ok(zlib_root_dir) = env::var("DEP_Z_ROOT") {
-        let mut pkg_config_path = OsString::from(format!(
-            "{}/lib/pkgconfig",
-            zlib_root_dir.replace("\\", "/")
-        ));
-        if let Some(env_pkg_config_path) = get_cc_rs_env_os("PKG_CONFIG_PATH") {
-            pkg_config_path.push(":");
-            pkg_config_path.push(env_pkg_config_path);
+    if cfg!(all(feature = "libz-rs", feature = "libz-sys")) {
```
sagudev, line 190: I didn't know we can use cfg to access features in build scripts, but it does work.

jschwe: I think the recommended way is to inspect the feature via the `CARGO_FEATURE` environment variables at runtime of the build-script. I'm not sure if we can rely on the build-script being recompiled if the selected features change!

sagudev: They got passed to rustc so one would expect they also got passed on change: https://github.com/servo/mozjs/actions/runs/17174634797/job/48728865465#step:6:489

redfire75369: From my research, `libz-sys` seems to use `cfg!` so I doubt there'll be any problems.

PR #620, opened 2025-08-27
Add minimal bindings for `Handle<StackGCVector<T>>`.

jdm:
Bindgen can't handle `StackGCVector<T>` values automatically, but Servo only needs to support a couple operations on handle values that contain them. This change sets up a fake Rust `StackGCVector` type which can be used to shuttle real values between handwritten C++ glue and Rust callers.

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -1164,9 +1164,8 @@ impl<'a> Handle<'a, StackGCVector<JSVal, js::TempAllocPolicy>> {
         if index >= self.len() {
             return None;
         }
-        let handle = unsafe {
-            Handle::from_marked_location(StackGCVectorValueAtIndex(*self, index))
-        };
+        let handle =
+            unsafe { Handle::from_marked_location(StackGCVectorValueAtIndex(*self, index)) };
         Some(handle)
     }
 
@@ -1180,9 +1179,8 @@ impl<'a> Handle<'a, StackGCVector<*mut JSString, js::TempAllocPolicy>> {
         if index >= self.len() {
             return None;
         }
-        let handle = unsafe {
-            Handle::from_marked_location(StackGCVectorStringAtIndex(*self, index))
-        };
+        let handle =
+            unsafe { Handle::from_marked_location(StackGCVectorStringAtIndex(*self, index)) };
         Some(handle)
     }
 


mozjs-sys/build.rs
```diff
@@ -758,6 +758,9 @@ impl BuildTarget {
                 // is more than bindgen can cope with.
                 "JS::Rooted",
                 // We don't need them and bindgen doesn't like them.
+                "JS::StackGCVector.*",
+                "JS::RootedVector_Vec",
+                "JS::RootedVector_Base",
```
sagudev, line 763: The comment is wrong for those types, beacuse we obviously need them. Maybe they should be in their own section.

mozjs-sys/src/jsgc.rs
```diff
@@ -2,11 +2,17 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::jsapi::JS;
+use crate::glue::{
+    HandleStringFromStackGCVector, HandleValueFromStackGCVector, StackGCVectorStringLength,
+    StackGCVectorValueLength,
+};
+use crate::jsapi::{js, JS};
```
sagudev: Why do we need this? They are not used anywhere, are they?

mozjs/src/rust.rs
```diff
@@ -1154,6 +1159,26 @@ impl Drop for EnvironmentChain {
     }
 }
 
+impl<'a> Handle<'a, StackGCVector<JSVal, js::TempAllocPolicy>> {
+    pub fn at(&self, index: u32) -> Handle<'a, JSVal> {
```
sagudev: IIRC last time we talked about this it was decided that we need 'a on self too:
```suggestion
    pub fn at(&'a self, index: u32) -> Handle<'a, JSVal> {
```

mozjs-sys/src/jsgc.rs
```diff
@@ -462,3 +468,7 @@ impl CustomAutoRooterVFTable {
     #[cfg(not(windows))]
     pub const PADDING: [usize; 2] = [0, 0];
 }
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct StackGCVector<T, AllocPolicy>(PhantomData<(T, AllocPolicy)>, u8);
```
sagudev:
```suggestion
pub struct StackGCVector<T, AllocPolicy = TempAllocPolicy>(PhantomData<(T, AllocPolicy)>, u8);
```
per https://searchfox.org/firefox-main/rev/856a307913c2b73765b4e88d32cf15ed05549cae/js/public/TypeDecls.h#69

sagudev (commented):
I think for cases where we shadow C++ struct manually we should have tests to ensure stuff works and will continue to work.

mozjs-sys/src/jsgc.rs
jdm: Oh yeah, that's from an older version of the changes.

mozjs/src/rust.rs
```diff
@@ -1154,6 +1159,26 @@ impl Drop for EnvironmentChain {
...
+    }
+
+    pub fn len(&self) -> u32 {
+        unsafe { StackGCVectorValueLength((*self).into()) }
+    }
+}
+
+impl<'a> Handle<'a, StackGCVector<*mut JSString, js::TempAllocPolicy>> {
+    pub fn at(&self, index: u32) -> Handle<'a, *mut JSString> {
+        unsafe { Handle::from_raw(HandleStringFromStackGCVector((*self).into(), index)) }
```
Taym95: question: are we indexing without validating index < len() ? what will happen if If a caller passes an out of bounds index? 

if so what about:

```
pub fn len(&self) -> u32 {  }

    pub fn get(&self, index: u32) -> Option<Handle<'a, JSVal>> {
        if index >= self.len() { return None; }
        Some(unsafe { Handle::from_raw(HandleStringFromStackGCVector((*self).into(), index)) })
    }
```

jdm: That's true, we will crash if we don't check. Good catch!

jdm:
These changes now include a test.

sagudev:
I suspect we need to add some padding on windows. If only there was a way to still get bindgen size/align asserts on shadowd types.

jdm:
Theory confirmed: for HandleStringFromStackGCVector, the handle that is received on the C++ side contains a pointer value that is actually the `uint32_t index` argument for that function.

jdm:
I fixed the Windows x64 issue by returning pointers to the array elements instead of Handles, so now the Rust bindings need to use Handle::from_marked_location.

mozjs/src/generate_wrappers.sh
```diff
@@ -24,14 +24,16 @@ grep_heur() {
...
+  grep -F -v '> HandleValue' |
+  grep -F -v '> HandleString' |
   grep -v 'MutableHandleObjectVector' # GetDebuggeeGlobals has it
 }
 
 # usage find_latest_version_of_file_and_parse $input_file $out_wrapper_module_name
 find_latest_version_of_file_and_parse() {
   # clone file and reformat (this is needed for grep_heur to work properly)
   # this $(find) only gets last modified file
-  cp $(find target -name "$1" -printf "%T@ %p\n" | sort -n | tail -n 1 | tr ' ' '\n' | tail -n 1) "target/wrap_$1"
```
sagudev, line 34: we should probably make this var like we do for `gsed`

jdm: Done.

jdm:
Not sure how the build errors snuck in but it all compiles now.

sagudev (approved):

mozjs-sys/build.rs
```diff
@@ -882,6 +883,7 @@ impl BuildTarget {
                 ("root", "pub type FILE = ::libc::FILE;"),
                 ("root::JS", "pub type Heap<T> = crate::jsgc::Heap<T>;"),
                 ("root::JS", "pub type Rooted<T> = crate::jsgc::Rooted<T>;"),
+                ("root::JS", "pub type StackGCVector<T, AllocPolicy> = crate::jsgc::StackGCVector<T, AllocPolicy>;"),
```
TimvdLippe, line 886: Should `AllocPolicy` be removed here? Running into the following failure:

```
   --> components/script/script_runtime.rs:507:31
    |
507 |     parameter_strings: Handle<StackGCVector<HandleString>>,
    |                               ^^^^^^^^^^^^^ ------------ supplied 1 generic argument
    |                               |
    |                               expected 2 generic arguments
    |
note: type alias defined here, with 2 generic parameters: `T`, `AllocPolicy`
   --> /home/tvanderlippe/Projects/servo/target/debug/build/mozjs_sys-1250879c4857322e/out/build/jsapi.rs:9:330793
    |
9   | ... >; pub type Rooted < T > = crate :: jsgc :: Rooted < T >; pub type StackGCVector < T , AllocPolicy > = crate :: jsgc :: StackGCVector < T , AllocPolicy >; # [doc = " Ty...
    |                                                                        ^^^^^^^^^^^^^   -   -----------
help: add missing generic argument
    |
507 |     parameter_strings: Handle<StackGCVector<HandleString, AllocPolicy>>,
    |                                                         +++++++++++++
```

We already add a default `AllocPolicy` in `jsgc.rs`.

mozjs/tests/stack_gc_vector.rs
```diff
@@ -0,0 +1,103 @@
...
+use mozjs::rooted;
+use mozjs::rust::{Handle as SafeHandle, JSEngine, RealmOptions, Runtime, SIMPLE_GLOBAL_CLASS};
+
+static SECURITY_CALLBACKS: JSSecurityCallbacks = JSSecurityCallbacks {
+    contentSecurityPolicyAllows: Some(content_security_policy_allows),
+    codeForEvalGets: None,
+    subsumes: None,
+};
+
+unsafe extern "C" fn content_security_policy_allows(
```
TimvdLippe, line 24: I am surprised that this compiles. I have the exact same definition, yet run into the following:

```
error[E0308]: mismatched types
   --> components/script/script_runtime.rs:103:39
    |
103 |     contentSecurityPolicyAllows: Some(content_security_policy_allows),
    |                                  ---- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected fn pointer, found fn item
    |                                  |
    |                                  arguments to this enum variant are incorrect
    |
    = note: expected fn pointer `unsafe extern "C" fn(*mut JSContext, ..., ..., ..., ..., ..., ..., ..., ...) -> bool`
                  found fn item `unsafe extern "C" fn(..., ..., ..., ..., ..., ..., ..., ..., ...) -> bool {content_security_policy_allows}`
help: the type constructed contains `for<'a, 'b> unsafe extern "C" fn(*mut js::jsapi::JSContext, js::jsapi::RuntimeCode, js::jsapi::Handle<*mut js::jsapi::JSString>, js::jsapi::CompilationType, js::rust::Handle<'a, js::rust::StackGCVector<*mut js::jsapi::JSString, {type error}>>, js::jsapi::Handle<*mut js::jsapi::JSString>, js::rust::Handle<'b, js::rust::StackGCVector<js::jsapi::Handle<*mut js::jsapi::JSObject>, {type error}>>, js::jsapi::Handle<js::jsapi::Value>, *mut bool) -> bool {script_runtime::content_security_policy_allows}` due to the type of the argument passed
   --> components/script/script_runtime.rs:103:34
    |
103 |     contentSecurityPolicyAllows: Some(content_security_policy_allows),
    |                                  ^^^^^------------------------------^
    |                                       |
    |                                       this argument influences the type of `Some`
note: tuple variant defined here
   --> /home/tvanderlippe/.rustup/toolchains/1.89.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:599:5
    |
599 |     Some(#[stable(feature = "rust1", since = "1.0.0")] T),
    |     ^^^^
    = note: the full name for the type has been written to '/home/tvanderlippe/Projects/servo/target/debug/deps/script-259d16a73cbb7fb5.long-type-6006025203782094944.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

Did some googling, but unfortunately it's a bit above my head.

sagudev: Hm, I think you need to do cargo clean as it looks like dirty target folder.

TimvdLippe: I ran `./mach clean && ./mach clippy` and it fails with the exact same failure :/

jdm: You're using the wrong Handle type: `js::rust::Handle<'a,` is the Rust lifetime version; you want jsapi::Handle.

mozjs-sys/build.rs
jdm: You can work around it in the meantime if you import mozjs::gc::StackGCVector instead.

PR #623, opened 2025-09-30
Support build configurations without Intl API

tschneidereit:
The Intl API requires bundling a very large amount of data, and isn't needed in absolutely all environments.

This commit introduces an `intl` feature, which is enabled by default on `mozjs`, and builds SpiderMonkey without Intl support if that feature is disabled.

On MacOS, a mozjs embedding I'm working on shrinks from 32MB to 12MB, and a wasm32 build from also 32MB to 8.5MB (all stripped).

### DIFF
--- a/mozjs-sys/Cargo.toml
+++ b/mozjs-sys/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.140.0-4"
+version = "0.140.0-5"
 authors = ["Mozilla", "The Servo Project Developers"]
 links = "mozjs"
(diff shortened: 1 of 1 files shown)


jdm (approved):
Makes sense!

mozjs-sys/build.rs
sagudev, line 1: If the future is disabled we need to build from source like for debugmozjs: https://github.com/servo/mozjs/blob/f5d9a371d55935c6184dece01ae3deb13c99a872/mozjs-sys/build.rs#L460

tschneidereit: Done

tschneidereit:
Merge queues experienced some hiccups yesterday, which I think were the reason for this bouncing. A re-queue should fix that.

PR #624, opened 2025-10-01
Support building for Wasm/WASI

tschneidereit:
This commit changes the `build.rs` and `makefile.cargo` files to support compiling mozjs for `wasm32-wasip1` and `wasm32-wasip2`. SpiderMonkey itself has long supported targeting WASI, but the build system requires some finagling to line everything up just right for these targets.

It also frobs things a little bit to make tests pass (disabling the one test that uses threads and thus isn't supported at all on wasm32-wasi right now, and making `criterion` not use `rayon`.)

Eventually it'd be nice to also do build artifacts as for other platforms, but for now that's not a blocker.

Note that this currently only works with the Intl disabled using #623, because `diplomat-runtime` unfortunately assumes that `wasm32` as a target arch implies the availability of web APIs. With that applied and building without default features, this passes a `cargo test --target=wasm32-wasip1` run.


tschneidereit:
Just to feel things out: would you by any chance be open at all to building wasm artifacts (potentially without Intl)? Absolutely fine and understandable if not, in which case we'd probably do so in a mirror over in the `bytecodealliance` org, same was we currently do with [our own SpiderMonkey builds](https://github.com/bytecodealliance/StarlingMonkey/releases/tag/libspidermonkey_FIREFOX_140_0_4_RELEASE_STARLING).

It's possible that we'll have to keep an additional set of patches anyway, in which case this would be moot, but I hope to be able to get rid of those before too long.

tschneidereit:
I wasn't sure whether you'd want to spend CI time on this—see also my previous question. But I'm happy to add a configuration for that for sure.

tschneidereit:
I added CI, which is passing, so this should be ready, I think.

.github/workflows/build.yml
```diff
@@ -247,6 +247,39 @@ jobs:
...
+        run: |
+          cd /tmp
+          curl --retry 5 --retry-all-errors -OL https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-25/wasi-sdk-25.0-x86_64-linux.tar.gz
+          tar -xzf wasi-sdk-25.0-x86_64-linux.tar.gz
+          mv wasi-sdk-25.0-x86_64-linux wasi-sdk
+      - name: Build
+        env:
+          WASI_SDK_PATH: "/tmp/wasi-sdk"
+        run: |
+          cargo +${{ steps.toolchain.outputs.name }} build --target=wasm32-wasip1
```
sagudev: Can we also run some tests in some wasi runtime, probably using something like https://users.rust-lang.org/t/testing-code-compiled-to-webassembly/93237

tschneidereit: This turned into more of a rabbit hole than I expected: as mentioned in the last paragraph, builds only work with Intl disabled. We could test without the `intl` feature, but it seems unfortunate to test a different configuration from the one that'd be commonly used.

I opened https://github.com/rust-diplomat/diplomat/pull/977, hoping that we might get a patch release for `diplomat-runtime`, after which getting tests to pass should only require a lockfile update.

I'll see how that goes and push changes here if it happens quickly enough. If not, maybe testing without Intl is okay? (Testing locally, it doesn't make any difference to the results, at least.)

tschneidereit: This is all moot: thanks to excellent maintenance of Diplomat, we now have an `0.8.3` release that resolves this issue.

mozjs-sys/makefile.cargo
```diff
@@ -29,6 +35,9 @@ endif
 
 ifneq (,$(CARGO_FEATURE_LIBZ_RS))
 	CONFIGURE_FLAGS += --enable-libz-rs
+else ifneq (,$(findstring wasi,$(TARGET)))
+	# system libz is not available on wasi
+	CONFIGURE_FLAGS += --enable-libz-rs
```
sagudev: Maybe we should just error if libz-rs feature is not enabled on wasi and enable it unconditionally in mozjs for wasi.

tschneidereit: I agree that that'd be best. Unfortunately I can't see a way to do this while keeping `libz-sys` as a default feature for other platforms: features, and in particular the default feature set, can't be defined on a per-platform level AFAICT.

tschneidereit: The only way I can think of is to always have a `libz-sys` dependency, but on WASI targets define it as an alias for `libz-rs`. That way, the feature flag works and all defaults can be kept as-is. Seems terribly hacky and non-obvious to me though

sagudev: I think we can enable libz-sys feature of mozjs-sys unconditionally for wasm in mozjs by something like:
```
[target.'cfg(target_arch = "wasm32")'.dependencies]
mozjs_sys = { path = "", features = "libz-rs" }
```

tschneidereit: That takes care of the `libz-rs` dependency, yes. The problem is removing the `libz-sys` dependency, which comes in via the default feature set.

sagudev: Hm, I just checked and it seems that libz-sys supports wasm/wasi (I think it should do static build there), so I am not sure why we need this at all.

tschneidereit: I think the Gecko build system doesn't support statically linking a system-provided libz. Or at least I don't know of a way to tell it to do so.

sagudev: It should support is if it's provided via pkgconfig (which we use on windows). The problem that on cross compiling PKG_CONFIG_PATH gets reset. I have patch in progress for this that I will push on top of your changes.

tschneidereit: Ooh, very nice! Thank you for making this change, much appreciated.

PR #628, opened 2025-10-25
CI: Do not upload artifacts stabel/beta toolchains tests

sagudev:
This will make scheduled runs actually useful, because right now they fail due too conflict in artifact names.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
           subject-path: ./target/libmozjs-${{ matrix.platform.target }}.tar.gz
 
       - name: Upload artifact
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -18,6 +18,7 @@ env:
   CARGO_INCREMENTAL: 0
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MOZJS_CREATE_ARCHIVE: 1
+  NEW_RUST_CHECK: ${{ inputs.rust_version == 'stable' || inputs.rust_version == 'beta' }}
```
sagudev, line 21: Ah, env is actually stored as string so ! does not work on bool but on string (emptiness).

PR #630, opened 2025-10-25
Introduce RootedFixedValueArray and RootedDynamicValueArray

Taym95:


mozjs/src/gc/collections.rs
```diff
@@ -157,3 +158,219 @@ impl<T: Traceable + 'static> Drop for RootedTraceableBox<T> {
         }
     }
 }
+
+/// Inline, fixed capacity buffer of JS values with GC barriers.
+/// Backed by `[Heap<JSVal>; N]`, and a manual `len`.
+pub struct FixedValueArray<const N: usize> {
+    elems: [Heap<JSVal>; N],
```
Taym95, line 165: @jdm what do you think about this? and how it is used in https://github.com/servo/servo/pull/40166 ?

Taym95:
> This isn't quite what I had in mind—we should not need new macros for this, and we should not need heap values for the fixed size array. While I was looking for links for how I would restructure, it, I came across
> 
> https://github.com/servo/mozjs/blob/f5ed6f95f58e4ba683afcbbba46fd309c3ca8839/mozjs-sys/src/jsgc.rs#L279
> 
> which is exactly what I was trying to describe in the original issue. It's all set up for HandleValueArray, too: https://github.com/servo/mozjs/blob/main/mozjs/src/gc/root.rs#L108 . Try following the example of
> https://github.com/servo/mozjs/blob/31afcde5e66498594687ab75498fa49e1201a296/mozjs/examples/wasm.rs#L135
> 
> to use it in Servo?

I see, let me see what I can do

Issue #588, opened 2025-07-02
Upgrade to icu4x 2.0


nicoburns:
Mozilla upstream has upgraded `icu4x` to 2.0, and Servo is starting to pull that in via Stylo (https://github.com/servo/servo/pull/37822). We'd like to avoid duplicating icu4x in the tree if possible, so we need a `mozjs` that uses `icu4x` 2.0. This looks like it will likely be SpiderMonkey 142 from upstream, or we alternatively we could backport the change.

Servo upgrade PR: https://github.com/servo/servo/pull/37823
Mozilla upgrade bugzilla issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1960300

PR #636, opened 2025-10-30
Further fixes to offline builds with vendored sources

jschwe:
- When invoking cargo metadata explicitly set the manifest path and the cwd. Since the cwd is in the target directory, and vendored sources can be placed at arbitrary locations, not setting the cwd / manifest path can cause the invocation to fail.
- We need `--offline` to prevent cargo metadata from attempting to access the internet (which fails in offline builds). Internet access is not needed for the data we are interested in. 

Manually tested in servo with the following process:

1. checkout servo main
2. patch mozjs to point to this PRs repo and branch
3. `cargo update -p mozjs`
4. `cargo clean`
5. `cargo vendor`
6. edit `.cargo/config.toml` according to the vendor instructions
7. disconnect internet
8. `./mach build --frozen`

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -149,6 +149,7 @@ fn get_icu_capi_include_path() -> String {
         .manifest_path(manifest_path)
         // icu_capi is feature guarded behind the `intl` feature.
         .features(CargoOpt::SomeFeatures(vec!["intl".into()]))
+        .other_options(vec!["--offline".into()])
         .exec()
         .unwrap();
     let packages = metadata.packages;


sagudev (approved):
LGTM, but it will also need mozjs-sys version bump.

sagudev:
Hm, I am starting to think it would be easier to just have patch that would update to icu2 per https://github.com/servo/mozjs/issues/588 or try to get https://github.com/unicode-org/icu4x/pull/6887 backported to icu1.5.

sagudev:
> I'm starting to think, that perhaps just vendoring the c header files might be the reasonable thing to do.

Per https://github.com/unicode-org/icu4x/pull/7204#issuecomment-3473566170, I think we should just keep vendored headers in SM checkout, but only the headers (so only intl/icu_capi/bindings): https://github.com/servo/mozjs/blob/31afcde5e66498594687ab75498fa49e1201a296/mozjs-sys/etc/filters.txt#L23

jschwe:
Not necessary anymore after #639.

PR #637, opened 2025-10-31
Use `DEP_ICU_CAPI1_C_INCLUDE_DIR`

sagudev:
draft because we are waiting for https://github.com/unicode-org/icu4x/pull/7203 to be released

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -220,7 +220,7 @@ fn build_spidermonkey(build_dir: &Path) {
 
     #[cfg(feature = "intl")]
     {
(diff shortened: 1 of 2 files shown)


mozjs-sys/build.rs
```diff
@@ -249,7 +220,7 @@ fn build_spidermonkey(build_dir: &Path) {
 
     #[cfg(feature = "intl")]
     {
-        let icu_c_include_path = get_icu_capi_include_path();
+        let icu_c_include_path = env::var("DEP_ICU_CAPI_C_INCLUDE_DIR").unwrap();
```
jschwe: This (Edit: the name after `DEP_`) needs to match the name from the `links` key, i.e. you need to add the `1` you added in the links key.

Issue #560, opened 2025-03-05
Traceable is unsound


gmorenz:
I'll probably fix this at the same time as fixing #520, but just in case creating an issue for it.

`Traceable` takes an `&self` and eventually calls into spidermonkey's tracing functions. Spidermonkey may then mutate the thing being traced to point to a new location where it moved the object. [Per this comment](https://github.com/mozilla/gecko-dev/commits/HEAD/js/public/TracingAPI.h).

Traceable is defined as

```
pub unsafe trait Traceable {
    /// Trace `self`.
    unsafe fn trace(&self, trc: *mut JSTracer);
}
```

It takes self as a immutable reference, despite mutating it. Short of interior mutability, this is unsound. It is implemented for `Value` and `ValueArray`, which do not (and IMHO should not) have interior mutability. So the type signature of `trace` needs to change to either `&mut self`, `self: &UnsafeCell<Self>`.

jdm:
Thinking out loud: is this is only unsound when implemented against types that can mutate due to GC: JSVal, `*mut JSObject`, etc? Or is the presence of any field that can be mutated enough to make this unsound?

jdm:
I'm pretty sure I attempted to use a `&mut self` type signature back when I first implemented this 10 years ago, but there was no way to guarantee exclusive mutable access since GCs can be triggered at just about any time.

gmorenz:
> I think it's both.

From an `&mut` point of view I agree, from a interior mutability point of view, only the things that the GC modifies would need to be wrapped in a `Cell` type.

I think we could narrowly fix this issue, and not the more structural issues issues in #520, by replacing the definition of `Value` with `struct Value(Cell<OldValue>)`. Wrap things like `*mut JSObject` into `Cell<*mut JSObject>` *everywhere*. Which, so long as rust code never meaningfully de-references anything the GC mutates, would be a relatively minor API change. I'm not sure if that's the case? Dereferencing *mut JSObject seems unlikely, but I remember seeing a reference to updating hash values and so on that made me think structs with their internals defined on the rust side might need to be mutated as well?

> I think that is exactly the problem. When we are doing GC there should be no & to js managed (as it can be mutated/moved). Handles/MutableHandles are OK (because they are effectively JS pointers that will be corrected by SM when GC is done).

[...]

PR #638, opened 2025-10-31
Add safe JSContext wrapper with strict safety rules

sagudev:
Somehow related to https://github.com/servo/mozjs/pull/554, but it creates more strict JSContext. Companion PR in servo: https://github.com/servo/servo/pull/40465

More concrete documentation is written at types: https://github.com/sagudev/mozjs/blob/cx/mozjs/src/context.rs and also includes some compile tests. Here I will only describe rough idea and motivation.

If we want to build any proper safe wrappers we need to have a notion of GC/NoGC, `JSContext` somehow fulfills the role of GC (generally every SM function that can trigger GC takes JSContext). But to be able to use JSContext (and it's lifetime) for this notion we need to make it not globally obtainable obtainable (it needs to be passed down as an arg as mentioned in https://github.com/asajeffrey/josephine). This is already happening with `CanGC` we have in servo (this new `JSContext` would replace `CanGC`).

This PR also introduce new wrappers that use lifetimed handles and new safe JSContext. More specifically they use `&mut JSContext` (for functions that can trigger GC) or `&JSContext` (for functions that cannot trigger GC). We also remove https://doc.servo.org/mozjs/jsapi/struct.AutoRequireNoGC.html from args (as NoGC is already encoded in `&JSContext` argument).

While this model is safe and usable from mozjs, it's somehow less usable in servo, where Runtime is always part of some *GlobalScope dom_struct or TLS. So in practice all entrypoints (where we get initial cx for passdown) will likely need to be unbounded (unsafe).

#### The plan forward

The idea is that we can incrementally introduce safety to servo:

- use known list of GC/NoGC functions https://github.com/servo/mozjs/pull/638#issuecomment-3495297584
- start using and passing down this new `JSContext` everywhere in servo (this will also replace can_gc effort) and use wrappers2
    - it make make more sense to pass down &mut InRealm (which is JSAutoRealm that has cx inside)
- write some more wrappers mainly we need one for JSAutoRealm, js_str stuff, to_jsval, RootingGuard, ...
- Handles/Roots would use NoGC to create & of inner data
- DomRefCell would also take NoGC on borrow, borrow_mut to prevent borrows across GC.
- I think we could also write part of unrooted with this, when creating unrooted value it would have `'cx` lifetime (bounded to cx) and rooting would erase the lifetime (make it `'static` thus unbound it from cx).
   - Something like this is already presented in https://github.com/asajeffrey/josephine
   - The biggest problems are return values, which also need to be bounded. This will require some additional work in mozjs

#### Notes on the model

While GC is run only on some compartments/realm, this becomes complicated to model because of cross references. So this model assumes all compartments in cx will get traced/GC (even if that is not the case; but that does happen on full GC I think). This is more strict than per compartments, but much easier to model and reason about (look at how complicated https://github.com/asajeffrey/josephine is). We can always introduced more relaxed model later.

This model is already discussed in https://github.com/servo/mozjs/issues/520, https://github.com/servo/mozjs/issues/560, https://github.com/servo/mozjs/issues/569 and something similar is also used in nova js engine: https://trynova.dev/blog/guide-to-nova-gc (I borrowed idea of explicit NoGC token).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v6
         with:
-          python-version: '3.13' 
+          python-version: "3.13"
       - name: Install deps
         run: |
           brew install llvm yasm
@@ -90,9 +90,11 @@ jobs:
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.9
       - name: Build
+        # doc tests on mozjs_sys fail because they include stuff from SpiderMonkey
         run: |
           cargo +${{ steps.toolchain.outputs.name }} build --verbose --features ${{ matrix.features }}
           cargo +${{ steps.toolchain.outputs.name }} test --tests --examples --verbose --features ${{ matrix.features }}
+          cargo +${{ steps.toolchain.outputs.name }} test --doc -p mozjs --verbose --features ${{ matrix.features }}
       - name: Check wrappers integrity
         # we generate wrappers only without debugmozjs
         if: ${{ matrix.features != 'debugmozjs' }}


sagudev:
While this model is safe from mozjs POV, the question is how usable it is in servo, where `Runtime` is always part of some GlobalScope dom_struct. So in practice all entrypoints (where we get initial cx for passdown) will need to be unbounded (unsafe).

Also doc tests are not run in CI and I think we would benefit with DerefMut hatch too.

sagudev:
Yesterday I learned that searchfox actually has trigger gc/not gc labels on functions: https://matrix.to/#/!CZEbuMfGYVdQMIBKeP:mozilla.org/$gMT8sxO85ePN_QOJ6Paa0q-nRILwXATU2iJ_EzoNhq8

These are generated by CI, so I wonder if we could use this data for autogenerating wrappers.

sagudev:
cc @gterzian @Taym95 who might also be interested in this

jdm (approved):
This is exciting and very promising! My only complaint is the new & syntax in the rooting macros, but I don't actually have any suggestions.

PR #639, opened 2025-10-31
Keep icu_capi headers in SM

sagudev:
Per https://github.com/servo/mozjs/pull/636#issuecomment-3473574950

Reviewable per commit (second commit is just application of first commit).


jschwe (approved):
Thanks!

PR #640, opened 2025-11-02
re-export mozjs_sys::jsgc::ValueArray

Taym95:


mozjs/src/lib.rs
```diff
@@ -56,6 +56,7 @@ pub mod typedarray;
 
 pub use crate::consts::*;
 pub use mozjs_sys::glue;
+pub use mozjs_sys::jsgc::ValueArray;
```
jdm: Let's add this in https://github.com/Taym95/mozjs/blob/2f723e74593e1da4d0cc666f4e0141cc862f9c10/mozjs/src/gc/mod.rs#L5 for consistency.

jdm (approved):

PR #650, opened 2025-11-11
Create safe realm wrappers

sagudev:
Main usages of using global JSContext in servo are from realms, so with this PR we introduce safe realm concepts that are derived from JSContext (this allows us to enforce LIFO of realms). We also introduce counterpart to servos `AlreadyInRealm` and `InRealm`: https://github.com/servo/servo/blob/main/components/script_bindings/realms.rs. Eventually these new safe types should replace old servo types. Because this change make it hard to use runtime I removed most methods from it.

### DIFF
--- a/mozjs/src/realm.rs
+++ b/mozjs/src/realm.rs
@@ -1,4 +1,5 @@
 use std::marker::PhantomData;
+use std::ops::{Deref, DerefMut};
 use std::ptr::NonNull;
 
 use crate::jsapi::JS::Realm;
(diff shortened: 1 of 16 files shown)


mozjs/src/realm.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    pub fn realm(&self) -> &NonNull<Realm> {
+        &self.realm
+    }
+}
+
+pub enum InRealm<'cx> {
+    Already(&'cx mut AlreadyInRealm<'cx>),
+    Entered(&'cx mut AutoRealm<'cx>),
```
sagudev: There are 3 cases:
- realm (not entered)
- entered
- current (already)

entering realm adds realm to stack, the top one being current

PR #655, opened 2025-11-12
Add embedder defined allocator support (mimalloc example) 

jschwe:
This is a PR to explore adding support for using an allocator provided by the servo / the embedder.
An example is provided (`examples/embedder_allocator`), which uses `mimalloc` to show how an embedder is expected to provide a custom allocator for Spidermonkey to use. To summarize:

> # Example: Using mozjs with an embedder-provided allocator

> 1. In your project setup an `include` directory containing `servo_embedder_allocator.h` defining
   `SERVO_EMBEDDER_MALLOC_PREFIX` and all required allocation functions with the corresponding prefix.
    See `mimalloc/include` for an example based on mimalloc.
> 2. Enable the feature `custom-alloc` and set the environment variable `SERVO_CUSTOM_ALLOC_INCLUDE_DIR` to
   point to the include directory created in step 1.

This design ensures that the allocator functions can be inlined, and minimize boilerplate for allocators which provide prefixed versions of standard allocator APIs. 

This PR does not eliminate all usages of standard malloc / free, but covers the most important ones. This is not a major issue, as long as the remaining usages are isolated in the respective third-party library.
Most significantly, `icu` is not covered, however that is exposed via `JS_SetICUMemoryFunctions`, so we could initialize that from the embedder / servo side. 

Notes: 

- `-DJS_USE_CUSTOM_ALLOCATOR` doesn't seem to be actually useful / maintained, since any custom implementation still need to copy lots of code from `Utility.h`
-  Some code does make assumption that `moz_arena_` allocated pointers can be freed / re-alloced by unprefixed `free()`, this lead to a patch to StringBuffer.h

### DIFF
--- a/mozjs-sys/mozjs/js/public/Utility.h
+++ b/mozjs-sys/mozjs/js/public/Utility.h
@@ -431,7 +431,7 @@ static inline void js_free(void* p) {
   // currently can't enforce that all memory freed here was allocated by
   // js_malloc(). All other memory should go through a different allocator and
   // deallocator.
-  free(p);
+  free_impl(p);
 }
 #endif /* JS_USE_CUSTOM_ALLOCATOR */
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = ["mozjs-sys", "mozjs"]
+exclude = ["examples"]
 resolver = "2"
 
 [workspace.package]
--- a/examples/embedder_allocator/Cargo.toml
+++ b/examples/embedder_allocator/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "embedder_allocator"
+version = "0.1.0"
+license = "MPL-2.0"
+edition = "2024"
+description = "Example usage of mozjs, with the embedder (this crate) providing a custom allocator for mozjs"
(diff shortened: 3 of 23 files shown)


mozjs-sys/mozjs/js/public/Utility.h
```diff
@@ -431,7 +431,7 @@ static inline void js_free(void* p) {
   // currently can't enforce that all memory freed here was allocated by
   // js_malloc(). All other memory should go through a different allocator and
   // deallocator.
-  free(p);
+  free_impl(p);
```
sagudev, line 434: Is this even used given that we do not enable JS_USE_CUSTOM_ALLOCATOR?

jschwe: yes! If you replace this with `free` you will get crashes due to the wrong allocator freeing.

jschwe: Note that the allocations use `moz_arena_<>`, which will allocate via our embedder provided functions.

sagudev: Ah, got confused by #endif guards comment but there is else in between.

Issue #652, opened 2025-11-11
Fix mismatched_lifetime_syntaxes warning


jschwe:
When compiling mozjs with stable rust `mismatched_lifetime_syntaxes` is now warn by default.

```
warning: hiding a lifetime that's elided elsewhere is confusing
   --> mozjs/src/gc/collections.rs:128:19
    |
128 |     pub fn handle(&self) -> Handle<T> {
    |                   ^^^^^     ^^^^^^^^^ the same lifetime is hidden here
    |                   |
    |                   the lifetime is elided here
    |
    = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
    |
128 |     pub fn handle(&self) -> Handle<'_, T> {
    |                                    +++
```

To reproduce:

`cargo +stable build`

PR #656, opened 2025-11-12
Fix mismatched_lifetime_syntaxes warning

jschwe:
Fixes https://github.com/servo/mozjs/issues/652

Not sure if this fixes all instances, since there are still many bindgen related warnings polluting the output, where I can't seem to find the option that can silence them.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -46,7 +46,7 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
         unsafe { Handle::from_marked_location(self.as_ptr()) }
     }
 
-    pub fn handle_mut(&mut self) -> MutableHandle<'_, T> {
+    pub fn handle_mut(&'_ mut self) -> MutableHandle<'_, T> {
         unsafe { MutableHandle::from_marked_location(self.as_ptr()) }
     }
 


mozjs/src/gc/collections.rs
```diff
@@ -125,7 +125,7 @@ where
     Heap<T>: Traceable + 'static,
     T: GCMethods + Copy,
 {
-    pub fn handle(&self) -> Handle<T> {
+    pub fn handle(&self) -> Handle<'_, T> {
```
sagudev: Shouldn't all lifetimes be explicit by lint?

```suggestion
    pub fn handle(&'_ self) -> Handle<'_, T> {
```

mozjs/src/gc/root.rs
```diff
@@ -42,7 +42,7 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
         Handle::new(&self)
     }
 
-    pub fn handle_mut(&mut self) -> MutableHandle<T> {
+    pub fn handle_mut(&'a mut self) -> MutableHandle<'a, T> {
```
jschwe: Hmm, this seems to cause borrow-checker errors. The lifetime `'a`should be correct, given that `handle()` also uses it? 
I don't have time to look deeper into this today though.

sagudev: Ah I know where is the problem. we have impl<'a> do you get 'a from that, but you want lifetime from &self so you need to introduce new lifetime (preferably with different name). That's also one of the reasons why I prefer explicit and not elided lifetimes as they make stuff less ambiguous.

jschwe: Yeah, but rust didn't like 'b either (since it can be elided), so I'm back to elided now. But that had me confused for a while 🤣

mozjs/src/gc/custom.rs
```diff
@@ -141,7 +141,7 @@ impl<'a, T: 'a + CustomTrace> CustomAutoRooterGuard<'a, T> {
         unsafe { Handle::from_marked_location(&raw const (self.rooter.data)) }
     }
 
-    pub fn handle_mut(&mut self) -> MutableHandle<T>
+    pub fn handle_mut(&'a mut self) -> MutableHandle<'a, T>
```
sagudev: I think this is wrong, we should use separate lifetime.

jschwe: hmm, I think you are right. Wouldn't that mean that the regular `handle` functions should also be `'_` and not `'a`?

mozjs/src/gc/root.rs
```diff
@@ -46,7 +46,7 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
         unsafe { Handle::from_marked_location(self.as_ptr()) }
     }
 
-    pub fn handle_mut(&mut self) -> MutableHandle<T> {
+    pub fn handle_mut(&mut self) -> MutableHandle<'_, T> {
```
sagudev: Is there no mismatched lifetime sytaxes here? Because:

```suggestion
    pub fn handle_mut(&'_ mut self) -> MutableHandle<'_, T> {
```

mozjs/src/gc/custom.rs
sagudev: yes

BTW in some cases we even do: https://github.com/servo/mozjs/blob/b01d302510d31cb0b25690b7a014ccb23dcac06e/mozjs/src/gc/root.rs#L401-L403

mozjs/src/gc/root.rs
jschwe: There is no warning both with and without your suggestion (using rust 1.92). `&'_ self` is equivalent with `&self` isn't it? But it also doesn't hurt to add it, so let's do it.

mozjs/src/gc/custom.rs
jschwe: Can I do this in a follow-up to keep this one scoped to pure warning-fixes?

sagudev: I am not really 100% sure we need `'a: 'b` part TBH. Its just something I want to bring attention to. This will need more investigation.

PR #657, opened 2025-11-13
Fixup global method on realm and add reborrow variant

sagudev:
In #650 I did implemented global wrongly and cast hide the problem (`*const` != `*const *mut`). To fix things properly I added `JSObject *const * JS::CurrentGlobal(JSContext* cx)` to jsapi which returns `JSObject *const *` (`CurrentGlobalOrNull` already deref the handle to return the object directly, which makes it harder to return handle again).

Also existing code made use of Handle very hard as we cannot use &mut realm as it's already borrowed by handle, but the only reason we bound the lifetime to handle is to not outlive the realm. So by creating `global_and_reborrow(&'_ mut self) -> (Handle<'_, *mut JSObject>, &'_ mut Self)` we return "reborrowed" realm, handle will not outlive original realm borrow and we can still use &mut realm.

The callback test is using these new global_and_reborrow so we know it's working now.

### DIFF
--- a/mozjs/src/realm.rs
+++ b/mozjs/src/realm.rs
@@ -143,7 +143,7 @@ impl<'cx> AutoRealm<'cx> {
     /// }
     /// ```
     pub fn global_and_reborrow(&'_ mut self) -> (Handle<'_, *mut JSObject>, &'_ mut Self) {
-        // SAFETY: This is ok because we bound handle will still be bounded to original lifetime
+        // SAFETY: This is ok because the handle will still be bound to original lifetime.
         (unsafe { std::mem::transmute(self.global()) }, self)
     }
 
@@ -236,7 +236,7 @@ impl<'cx> CurrentRealm<'cx> {
         unsafe { Handle::from_marked_location(CurrentGlobal(self)) }
     }
 
-    /// Obtain the handle to the global object of the this realm and reborrow the realm.
+    /// Obtain the handle to the global object of this realm and reborrow the realm.
     ///
     /// ```
     /// use mozjs::context::JSContext;
@@ -254,7 +254,7 @@ impl<'cx> CurrentRealm<'cx> {
     /// }
     /// ```
     pub fn global_and_reborrow(&'_ mut self) -> (Handle<'_, *mut JSObject>, &'_ mut Self) {
-        // SAFETY: This is ok because we bound handle will still be bounded to original lifetime
+        // SAFETY: This is ok because the handle will still be bound to original lifetime.
         (unsafe { std::mem::transmute(self.global()) }, self)
     }
 


mozjs/src/realm.rs
```diff
@@ -100,10 +100,51 @@ impl<'cx> AutoRealm<'cx> {
...
+    /// fn g(realm: &'_ mut AutoRealm, global: Handle<'_, *mut JSObject>) {
+    /// }
+    ///
+    /// fn f(realm: &mut AutoRealm) {
+    ///     let (global, realm) = realm.global_and_reborrow();
+    ///     g(realm, global);
+    /// }
+    /// ```
+    pub fn global_and_reborrow(&'_ mut self) -> (Handle<'_, *mut JSObject>, &'_ mut Self) {
+        // SAFETY: This is ok because we bound handle will still be bounded to original lifetime
```
jdm:
```suggestion
        // SAFETY: This is ok because the handle will still be bound to original lifetime.
```

mozjs/src/realm.rs
```diff
@@ -170,10 +211,51 @@ impl<'cx> CurrentRealm<'cx> {
...
+    /// ```
+    ///
+    /// instead use [CurrentRealm::global_and_reborrow].
     pub fn global(&'_ self) -> Handle<'_, *mut JSObject> {
         // SAFETY: object is rooted by realm
-        unsafe { Handle::from_marked_location(CurrentGlobalOrNull(&*self) as _) }
+        unsafe { Handle::from_marked_location(CurrentGlobal(self)) }
+    }
+
+    /// Obtain the handle to the global object of the this realm and reborrow the realm.
```
jdm:
```suggestion
    /// Obtain the handle to the global object of this realm and reborrow the realm.
```

mozjs/src/realm.rs
```diff
@@ -170,10 +211,51 @@ impl<'cx> CurrentRealm<'cx> {
...
+    /// fn g(realm: &'_ mut CurrentRealm, global: Handle<'_, *mut JSObject>) {
+    /// }
+    ///
+    /// fn f(realm: &mut CurrentRealm) {
+    ///     let (global, realm) = realm.global_and_reborrow();
+    ///     g(realm, global);
+    /// }
+    /// ```
+    pub fn global_and_reborrow(&'_ mut self) -> (Handle<'_, *mut JSObject>, &'_ mut Self) {
+        // SAFETY: This is ok because we bound handle will still be bounded to original lifetime
```
jdm:
```suggestion
        // SAFETY: This is ok because the handle will still be bound to original lifetime.
```

jdm (approved):

PR #663, opened 2025-11-18
build: Unify flags

jschwe:
Refactor cc_flags() to return Vec<String>, so that we can also add dynamic flags here.
This makes it easier to keep the flags in sync across bindings and compilation


jschwe:
Superceded

PR #666, opened 2025-11-18
Fix OHOS define

jschwe:
Extracted from #583. 

`target_is_ohos` was not working as expected. This presumably originated in a seemingly [harmless suggestion](https://github.com/servo/mozjs/pull/491#discussion_r1703960490),
but surprisingly the expected return type / values for these functions is `True | None`, and not `True | False` (The other functions like `target_is_xx` don't return anything explicitly if the `if` does not match, which results in a `None` implicitly I guess). Returning `False` does not give the expected result. I noticed this while working on #583 - normally logs are still visible in servo, since we setup a thread that redirects output from stdout / stderr to the logging system on ohos (except in production mode). 

Fixing the ohos define also revealed some issues with the patches:

- In the current version of SM, `LogLevel` defined in `hilog/log.h` collides with a `LogLevel` in the `js` namespace. Some of the header files we included hilog in, have this collision, and are also used from C files, so we can't just namespace the import (without the patch getting quite ugly).To avoid these issues we just copy the function declaration, and use raw integers instead of the enum in the function signature.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,6 +227,7 @@ jobs:
     strategy:
       matrix:
         target: ["aarch64-unknown-linux-ohos", "x86_64-unknown-linux-ohos"]
+        features: [ "", "--features debugmozjs"]
     steps:
       - uses: actions/checkout@v4
       - name: Setup OpenHarmony SDK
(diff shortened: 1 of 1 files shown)


jdm (approved):

jschwe:
Todo: add ci configuration for ohos with debugmozjs

jschwe:
This PR did change a bit since the initial review, so it might make sense to be reviewed again. I'm not super happy with inlining the function declaration in every file it is used, but I don't think it would make the patching easier, if I wrapped the logging in a new header file.

PR #669, opened 2025-11-19
Fix cross-compiling to ohos on macos 26

### LINKED ISSUE #667: Cross-compiling (configuring?) for OHOS from source is broken on macos 26
The only error I can see in my logs is:

```
 dyld[76899]: __DATA_CONST segment missing SG_READ_ONLY flag in /Users/jschwender/Dev/mozjs/target/aarch64-unknown-linux-ohos/debug/build/mozjs_sys-fcf429334e10f692/out/build/config/nsinstall
  dyld[76899]: __DATA_CONST segment missing SG_READ_ONLY flag
  make[4]: *** [../../dist/include/ProfilingCategoryList.h] Abort trap: 6
  make[3]: *** [mozglue/baseprofiler/export] Error 2
  make[3]: *** Waiting for unfinished jobs....
  make[2]: *** [export] Error 2
  make[1]: *** [default] Error 2
  make: *** [all] Error 2
```

Googling the error pointed me to https://github.com/leanprover/lean4/issues/5649, which suggests that LLVM 16+ is needed. This might mean we either need to use a newer clang when cross-compiling (the SDK clang is clang-15), or check if the spidermonkey build-system knows the concept of Host compiler (where we could use a newer clang version)

jschwe:
Appending the cross-compile toolchain to path instead of prepending fixes the issue of `nsinstall` being unable to run.
Probably we should have appended in the first place anyway. 

Fixes #667

### DIFF
--- a/ohos-build
+++ b/ohos-build
@@ -75,7 +75,7 @@ def create_environment_for_build() -> Dict[str, str]:
 
     llvm_toolchain = os.path.join(ndk_root, "llvm")
 
-    env['PATH'] = os.pathsep.join([os.path.join(llvm_toolchain, "bin"), env["PATH"]])
+    env['PATH'] = os.pathsep.join([env["PATH"], os.path.join(llvm_toolchain, "bin")])
 
     set_toolchain_binaries_in_env(llvm_toolchain, target_triple, env)
 


sagudev (approved):
If it works, it works.

PR #671, opened 2025-11-25
Reapply "Fix OHOS define"

jschwe:
This PR re-applies #666, which was reverted due to linking failing. 
The original issue was that when declaring the function `OH_LOG_Print` `__attribute__((visibility("default")))` needs to be added, since (parts of) mozjs are compiled with `-fvisibility=hidden`.
To catch such issues in CI in the future, this PR also adds a final step to the ohos CI which compiles the unit-tests, which will invoke the linker, and thus catch similar link errors in our CI.

Companion PR in servo: https://github.com/servo/servo/pull/40855

### DIFF
--- a/mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
+++ b/mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
@@ -10,22 +10,6 @@ index e0e14bf60..179de2592 100644
 
 
  @depends(target)
-diff --git a/config/system-headers.mozbuild b/config/system-headers.mozbuild
-index 471599f17..a89f0ca7a 100644
---- a/config/system-headers.mozbuild
-+++ b/config/system-headers.mozbuild
-@@ -1367,6 +1367,11 @@ if CONFIG["OS_TARGET"] == "FreeBSD":
-         "sys/capsicum.h",
-     ]
-
-+if CONFIG["OHOS"]:
-+    system_headers += [
-+       "hilog/log.h",
-+    ]
-+
- if CONFIG["MOZ_APP_SYSTEM_HEADERS"]:
-     include("../" + CONFIG["MOZ_BUILD_APP"] + "/app-system-headers.mozbuild")
-
 diff --git a/memory/mozalloc/mozalloc_abort.cpp b/memory/mozalloc/mozalloc_abort.cpp
 index 3cfc92533..9c487ac45 100644
 --- a/memory/mozalloc/mozalloc_abort.cpp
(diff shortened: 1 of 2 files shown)


mozjs-sys/etc/patches/0036-Redirect_errors_to_hilog_on_OpenHarmony.patch
```diff
@@ -61,50 +64,58 @@ diff --git a/mfbt/Assertions.h b/mfbt/Assertions.h
...
+@@ -88,6 +88,11 @@ MOZ_END_EXTERN_C
  #endif
  #ifdef ANDROID
  #  include <android/log.h>
-+#elif defined(OHOS)
-+#  include <hilog/log.h>
++#elif defined(XP_OHOS)
++MOZ_BEGIN_EXTERN_C
++int OH_LOG_Print(unsigned int type, unsigned int level, unsigned int domain, const char *tag, const char *fmt, ...)
++    __attribute__((__format__(os_log, 5, 6))) __attribute__((visibility("default")));
```
mukilan, line 58: Perhaps instead of adding the visibility attribute here, we just need to add `hilog/log.h` to the system headers [list](https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/config/system-headers.mozbuild)?

mukilan: Here [are some notes](https://github.com/servo/servo/pull/32769#issuecomment-2253857177) about the the system headers and visibility from the last time I had to debug a linker issue in SM, in case it is related to this issue.

jschwe: > Perhaps instead of adding the visibility attribute here, we just need to add hilog/log.h to the system headers [list](https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/config/system-headers.mozbuild)?

It is / was actually. The (new) problem is we can't use the `hilog/log.h` header, since the `LogLevel` enum defined there causes `ambiguous symbol` errors, due to an enum with the same name defined in mozjs.  We also can't simply namespace the `#include`, since some of the header files are also included in regular c files.
We could define a new header file that only provides this declaration, but the question is where to put it, and if that would make the patch simpler or not. 
At least, I am still leaning to keeping the duplicated declarations, as to not modify the build system more than necessary.

mukilan: I see. I hadn't read the previous PR, so I missed the part where the explicit prototype was added to avoid the symbol collision. 

If we are duplicating the prototype inline now, should the header be removed from the system headers [list](https://github.com/servo/mozjs/blob/main/mozjs-sys/mozjs/config/system-headers.mozbuild#L1397)?  I didn't realize we already have it there.

jschwe: I removed the system headers patch as suggested.

PR #675, opened 2025-11-27
mozjs: Allow using mozjs-sys from crates.io

jschwe:
Adding both version and path is the common way to support both local workspace development, and crates.io usage.

### DIFF
--- a/mozjs/Cargo.toml
+++ b/mozjs/Cargo.toml
@@ -26,7 +26,7 @@ crown = ["mozjs_sys/crown"]
 encoding_rs = "0.8.35"
 libc.workspace = true
 log = "0.4"
-mozjs_sys = { version = "0.140.5-6", path = "../mozjs-sys" }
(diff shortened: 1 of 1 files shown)


mozjs/Cargo.toml
```diff
@@ -26,7 +26,7 @@ crown = ["mozjs_sys/crown"]
 encoding_rs = "0.8.35"
 libc.workspace = true
 log = "0.4"
-mozjs_sys = { path = "../mozjs-sys" }
+mozjs_sys = { version = "0.140.5-6", path = "../mozjs-sys" }
```
sagudev:
```suggestion
mozjs_sys = { version = "=0.140.5-6", path = "../mozjs-sys" }
```

sagudev (approved):
Yes, this way I can finally publish mozjs too. Do you know if cargo complains if version mismatches? (one forgets to do the bump).

PR #502, opened 2024-09-10
CI: Hack around DCO not in MQ

sagudev:
Same as in servo: https://github.com/servo/servo/pull/33290

DO not forget to also allow DCO check from any source per: https://github.com/servo/servo/pull/33308#issuecomment-2327919183

PR #676, opened 2025-12-04
CI: Refactor publish step to have a dedicated publish workflow

jschwe:
It is a recommended best practice to use a dedicated workflow file for publishing, to reduce the attack surface.
We also move the artifact attestation to the publish workflow, since it makes the most sense there and allows attesting all artifacts in one step (possible by updating the attest action to v2 or newer).
This is a preparation for adding automated releases to crates.io.

[test run](https://github.com/jschwe/mozjs/actions/runs/19926672517/job/57129961585)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,10 @@ on:
         # Keep the default in sync with `rust-toolchain.toml` and servo's MSRV!
         default: "1.85.0"
         type: string
+      verify_artifacts:
+        required: false
+        default: true
+        type: boolean
 
 env:
   RUST_BACKTRACE: 1
@@ -369,7 +373,7 @@ jobs:
   verify-release-artifacts:
     name: Verify release artifacts
     needs: ["android", "linux", "mac", "ohos", "windows"]
-    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
+    if: ${{  inputs.verify_artifacts && !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
     uses: ./.github/workflows/release-check.yml
     with:
       release-tag: ''
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,6 +8,9 @@ on:
 jobs:
   build:
     uses: ./.github/workflows/build.yml
+    with:
+      # We verify in this workflow after creating the github release.
+      verify_artifacts: false
 
   publish-github-release:
     name: Check version and publish release


.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
+name: Publish
+
+on:
+  push:
+    branches: [main, trusted_publishing]
```
jschwe: will be reverted to only main before merging

.github/workflows/build.yml
sagudev: Is this even used anymore?

.github/workflows/build.yml
```diff
@@ -389,38 +355,6 @@ jobs:
...
-      release-tag: ${{ steps.check-tag.outputs.RELEASE_TAG }}
-
-  verify-release:
-    name: Verify release
-    needs: publish-release
-    if: ${{ !contains(needs.*.result, 'failure') && !contains(needs.*.result, 'cancelled') }}
-    uses: ./.github/workflows/release-check.yml
-    with:
-      release-tag: ${{ needs.publish-release.outputs.release-tag }}
-      rust_version: ${{ inputs.rust_version }}
```
sagudev, line 436: Here is the thing, we also do verfy-release on PRs/MQ, where instead of using actual release we use newly generated artifacts: https://github.com/servo/mozjs/blob/4c928791be0effe46d6810edbf63d4a728b13821/.github/workflows/release-check.yml#L42-L44

This ensures that no PR breaks the artifacts (which would be caught after release in this PR).

jschwe: I added the verify job back.

.github/workflows/build.yml
jschwe: Indeed, after moving the attestation it is not needed anymore - removed.

.github/workflows/main.yml
sagudev: > It is a recommended best practice to use a dedicated workflow file for publishing, to reduce the attack surface.

I think the reason why we use one workflow is because MQ/PR checks need to have same entry point workflow and we need to run full CI on for caching purposes too.

With this PR we would run CI on main twice, which might be a bit too much. Maybe we can remove push here? Or maybe we can move publish stuff here, but we would still kept build unprivileged.

.github/workflows/main.yml
```diff
@@ -13,42 +13,28 @@ on:
...
-    if: ${{ github.event_name == 'push' }}
-    permissions:
-      contents: write
-      id-token: write
-      attestations: write
-    uses: ./.github/workflows/build.yml
-    with:
-      release: true
-    secrets: inherit
   normal:
```
sagudev:
```suggestion
  build:
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,49 @@
...
+          RELEASE_TAG=mozjs-sys-v$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "mozjs_sys") | .version')
+          git fetch --tags --quiet
+          if ! git show-ref --tags --verify --quiet "refs/tags/${RELEASE_TAG}" ; then
+            gh release create ${RELEASE_TAG} ./*.tar.gz
+          fi
+          echo "RELEASE_TAG=${RELEASE_TAG}" >> ${GITHUB_OUTPUT}
+    outputs:
+      release-tag: ${{ steps.check-tag.outputs.RELEASE_TAG }}
+
+  verify-release:
```
sagudev, line 45: In this workflow file verify-release actually happens twice, once as part of build (we should skip it) and second time here.

sagudev (changes requested):

.github/workflows/main.yml
jschwe: > I think the reason why we use one workflow is because MQ/PR checks need to have same entry point workflow 

I don't think this is accurate. A `Result` job can only access the status in its own workflow, but one can specify arbitrary jobs as required CI checks, so having multiple Results jobs / multiple workflows is possible for the MQ / PR. This would need a change in the settings of this repository though (and it might only be possible to add the new check, after it was merged to main, since afaik they don't show up there before they have not run once on main).

> Maybe we can remove push here?

I did hat for now. We can always change it back if we later decide to switch to a different release trigger.

sagudev: > > I think the reason why we use one workflow is because MQ/PR checks need to have same entry point workflow
> 
> I don't think this is accurate. A `Result` job can only access the status in its own workflow, but one can specify arbitrary jobs as required CI checks, so having multiple Results jobs / multiple workflows is possible for the MQ / PR. This would need a change in the settings of this repository though (and it might only be possible to add the new check, after it was merged to main, since afaik they don't show up there before they have not run once on main).

Yes you can have more required checks, but checks that are in MQ also need to be in PR. Actually, we have relaxed rules in repos, so we just need to have same checks name available under both PR and MQ, this is also why we need this: https://github.com/servo/mozjs/pull/502

Anyway I think there should be no problem with this PR.

PR #677, opened 2025-12-04
Add CI check to ensure mozjs is bumped

jschwe:
For purposes of publishing to crates.io, we should also bump the `mozjs` version number, and update the dependency on mozjs-sys. `cargo` will give an error, if we update the mozjs-sys version in `mozjs-sys/Cargo.toml`, but forget to update the dependency in `mozjs/Cargo.toml` since we depend on an exact version by using the `=` specifier. 
The only thing we need to check in CI is that the mozjs version is also bumped, since that would prevent publishing to crates.io (without further changes).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -389,6 +389,19 @@ jobs:
             exit 1
           fi
 
+      - name: Ensure mozjs version is bumped if mozjs-sys version bumped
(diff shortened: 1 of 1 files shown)


sagudev:
Hm, actually it is safe to not bump mozjs version for any change (but we need this for mozjs-sys because of artifacts), so maybe we do not need this PR? I can see how it could be useful (to not forget bump if one wants to use this in servo), but also not if we want to land multiple PRs (in short span of time) and not release version for each one (I am worried we will spam to many version - and our crates are big).

PR #678, opened 2025-12-04
Use crates.io Trusted publishing to publish releases to crates.io

jschwe:
Depends on #676 and #677

See the [official crates.io trusted publishing documentation](https://crates.io/docs/trusted-publishing).

This requires a crates.io owner of mozjs to allow trusted-publishing!

>  Configure your crate on crates.io:
>    Go to your crate's Settings → Trusted Publishing
>    Click the "Add" button and select your platform (GitHub or GitLab)
>    Fill in the platform-specific fields and save the configuration

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,14 +42,10 @@ jobs:
           git fetch --tags --quiet
           if ! git show-ref --tags --verify --quiet "refs/tags/${RELEASE_TAG}" ; then
             gh release create ${RELEASE_TAG} ./*.tar.gz
-            echo "CREATE_RELEASE=true" >> ${GITHUB_OUTPUT}
-          else
-            echo "CREATE_RELEASE=false" >> ${GITHUB_OUTPUT}
           fi
           echo "RELEASE_TAG=${RELEASE_TAG}" >> ${GITHUB_OUTPUT}
     outputs:
       release-tag: ${{ steps.check-tag.outputs.RELEASE_TAG }}
-      create-release: ${{ steps.check-tag.outputs.CREATE_RELEASE }}
 
   verify-release:
     name: Verify release
@@ -61,7 +57,7 @@ jobs:
       rust_version: stable
 
   # Our CI testing ensures that mozjs-sys and mozjs are always bumped together.
-  # We always publish both versions, since mozjs depends on a speciifc version
+  # We always publish both versions, since mozjs depends on a specific version
   # of mozjs-sys
   publish-crates-io:
     name: Publish to crates.io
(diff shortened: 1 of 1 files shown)


jschwe:
> I can see how it could be useful (to not forget bump if one wants to use this in servo), but also not if we want to land multiple PRs (in short span of time) and not release version for each one (I am worried we will spam to many version - and our crates are big).

Let's continue the discussion from #677 here. If we switch to consistently using crates.io releases, then technically we don't need to build and release a prebuilt version of mozjs-sys for every commit on main, but just for every published release. 
So we could tag and release on-demand if wanted. Practically, we probably still want to release often, due to the tight coupling of servo and mozjs(-sys). 
Maybe we could parse the commit message and skip publishing if we match some regex that lets us skip the release? That would allow us to default to releasing, and skipping if we know we will be merging multiple related PRs quickly after each other.

I've been wondering about the versioning too - perhaps we should just use a workspace version (following the spidermonkey version, as we already do for mozjs-sys), and not version mozjs independently? That would make version bumps simpler, and we could always just release both libraries. It would be a bit wasteful (in terms of crates.io space usage) if mozjs-sys didn't actually change though.

.github/workflows/publish.yml
```diff
@@ -50,3 +59,29 @@ jobs:
...
+  # of mozjs-sys
+  publish-crates-io:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    needs:
+      - publish-github-release
+      - verify-release
+    permissions:
+      id-token: write
+    if: ${{ needs.publish-github-release.outputs.create-release }}
```
sagudev: cargo publish does nothing (it does not even build) if version number does not change, so maybe we can just remove if guard and let publish happen on any version bump (this should also handle bumping just mozjs).

.github/workflows/publish.yml
```diff
@@ -50,3 +59,29 @@ jobs:
     with:
       release-tag: ${{ needs.publish-github-release.outputs.release-tag }}
       rust_version: stable
+
+  # Our CI testing ensures that mozjs-sys and mozjs are always bumped together.
+  # We always publish both versions, since mozjs depends on a speciifc version
```
sagudev:
```suggestion
  # We always publish both versions, since mozjs depends on a specific version
```

.github/workflows/publish.yml
jschwe: I just checked and cargo publish will exit with `101` when trying to publish an already published version. 
I pushed a commit which uses grep to check the error message, and will keep the pipeline greep if the failure is due to the release already existing.

PR #680, opened 2025-12-10
Remove wrong handling of `*const Handle` in wrappers

sagudev:
Handling of `*const Handle` was wrong in `wrap!` macro. It generated code like this:
```rust
SM(&(*arg).into())
```
but this is wrong as we get ptr to invalid place (args are temporary values) cause crashing. I am not sure why rust was not complaining, but I suspect it has something to do with the fact that we change ref to ptrs. The solution is to do this:
```rust
let t = (*arg).into();
SM(&t);
SM(&raw const t);
```
but this is hard to do in macro, so I simply removed handling of this case and wrote a manual wrapper. Servo currently uses nonwrapped function call as wrapped would cause crashes.

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -1529,7 +1529,7 @@ pub mod wrappers2 {
                 id.into(),
                 v.into(),
                 receiver.into(),
-                &ownDesc,
+                &raw const ownDesc,
                 result,
             )
         } else {


sagudev:
What I think is happens here that we get temporary & that is then deref into ptr and this does not trigger error, but using `&raw const` (which is more correct anyway) will complain as expected:
```
error[E0745]: cannot take address of a temporary
    --> mozjs/src/rust.rs:1532:28
     |
1532 |                 &raw const ownDesc.into(),
     |                            ^^^^^^^^^^^^^^ temporary value

```

PR #688, opened 2025-12-22
Make Handle use a raw pointer instead of an immutable borrow.

jdm:
This ensures that the moving GC cannot trigger UB if we read the value from a handle, move the value it refers to, then read from it again. HandleMut already uses a similar construction.

Fixes the behaviour observed in https://github.com/servo/servo/issues/41190.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -39,7 +39,8 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
     }
 
     pub fn handle(&'a self) -> Handle<'a, T> {
-        Handle::new(&self)
+        // SAFETY: A root is a marked location.
+        unsafe { Handle::from_marked_location(self.as_ptr()) }
     }
 
     pub fn handle_mut(&mut self) -> MutableHandle<T> {
@@ -157,15 +158,11 @@ impl<'a, T> Handle<'a, T> {
         unsafe { *self.ptr }
     }
 
-    pub(crate) fn new(ptr: &'a T) -> Self {
+    pub unsafe fn from_marked_location(ptr: *const T) -> Self {
         Handle {
             ptr,
             _phantom: PhantomData,
         }
-    }
-
-    pub unsafe fn from_marked_location(ptr: *const T) -> Self {
-        Handle::new(&*ptr)
     }
 
     pub unsafe fn from_raw(handle: RawHandle<T>) -> Self {
@@ -210,8 +207,9 @@ impl<'a, T> MutableHandle<'a, T> {
         MutableHandle::from_marked_location(handle.ptr)
     }
 
-    pub fn handle(&self) -> Handle<T> {
-        unsafe { Handle::new(&*self.ptr) }
+    pub fn handle(&self) -> Handle<'a, T> {
+        // SAFETY: This mutable handle was already derived from a marked location.
+        unsafe { Handle::from_marked_location(self.ptr) }
     }
 
     pub(crate) fn new(ptr: &'a mut T) -> Self {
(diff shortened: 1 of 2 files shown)


mozjs/src/gc/root.rs
sagudev: I have something similar locally, when I was debugging for https://github.com/servo/mozjs/pull/680 and from what I recall there are some more places that we do &* hop that we should probably remove (like the one above). I will try to recover the local commit later today.

mozjs/src/gc/root.rs
```diff
@@ -153,11 +154,14 @@ impl<'a, T> Handle<'a, T> {
     where
         T: Copy,
     {
-        *self.ptr
+        unsafe { *self.ptr }
     }
 
     pub(crate) fn new(ptr: &'a T) -> Self {
```
gmorenz: Leaving this function signature alone is suspicious. If ptr is being made from an `&T` we can't write through it (according to the memory models) and we presumably end up with the same issue at a distance.

mozjs/src/gc/root.rs
sagudev: I just checked and the only thing missing is this:

```suggestion
        Self {
            ptr,
            anchor: PhantomData,
        }
```

mozjs/src/gc/root.rs
sagudev: I think this is ok for immutable handle, that will live for lifetime of `'a`.

simonwuelker:
Would it make sense to use `NonNull<T>` instead of `*const T` here? That way things like `Option<Handle<T>>` can still use the niche value optimization.

sagudev:
> Probably, but since we're currently only building our handles from `&T`s (which are `NonNull`) and SM only modifies the things the handles point to, not the handles themselves, I think we should be free to do this optimization.

You can also build handles from raw marked location (where location is potentially null ptr) and in cases where handle is returned from SM, this might matter, so we would potentionally need to start using `Option<Handle>` in many more places. With that being said I am not opposed to using more NonZero in mozjs, it's just that we currently have bigger problems to worry about.

mozjs/src/gc/root.rs
gmorenz: Technically I think you're right, since by having a `&'a T` we've already theoretically guaranteed that the GC won't modify `T` for the duration of `'a`... 

But I don't think we ever actually use this function properly - like that. Instead we're using it with manufactured references with fake lifetimes to create `Handle` types that have lifetimes that extend over GC pauses. And the pointers we create like this effectively inherit the restrictions of the shared references - i.e. the following is undefined behaviour in both stack-borrows and tree-borrows:

```rust
fn main() {
    unsafe { 
        let mut rooted = 0i32;
        let root = &mut rooted as *mut i32;
        let shared = (&*root) as *const i32;
        
        // Do gc:
        *root = 1;
        
        // This is undefined behavior
        *shared;
    }
}
```

sagudev: Uf yes, I was thinking of servo dom objects. This method is used for rooting, but we should just use ptrs directly and avoid converting them to reference at all.

One thing I also discovered in https://github.com/servo/mozjs/pull/680 is that `_: * const _ = &* _` is dangerous, as we create ref (that needs to uphold aliasing rules) just to get a ptr. We should instead use `&raw`. I think I read this somewhere at miri repo, but the closest I find are https://github.com/rust-lang/rust/issues/73987 and https://github.com/rust-lang/rust/issues/99437

gmorenz:
> You can also build handles from raw marked location (where location is potentially null ptr)

If this actually happens it seems like a significant problem. Both `MutableHandle::from_marked_location` (which I thought was sound at this point) and `Handle::from_marked_location` implicitly assume it's not null by converting to a reference. And while we could fix that we provide methods like `MutableHandle::get/set` and `Deref` and so on that allow reading and writing through these pointers which assumes that they aren't null.

jdm:
Yeah, I think we're conflating a Handle to a pointer type that can be null (like a `Handle<*must JSObject>`) with the pointer that the Handle holds. The value that Handle stores a pointer to can be a null pointer, but the pointer inside the Handle will never be null.

PR #689, opened 2025-12-26
Remove MutableHandle::new

gmorenz:
`MutableHandle::new` was unused except from `from_marked_location`. It was unsound when called from `from_marked_location` because`from_marked_location` converted the ptr to an `&mut` reference invalidating the pointer the GC holds. This was an oversight in #572. This fixes that by constructing the `MutableHandle` directly in `from_marked_location` and then removing the now unused function.

There is some discussion in https://github.com/servo/mozjs/pull/688#issuecomment-3689737344  that `from_marked_location` might also be being called with null pointers. I don't think that's happening based on a quick test with putting an assert in and running servo's tests, and if it is we should consider changing the API so that it doesn't happen in the future. In the meantime this makes that sound as a side benefit (so long as the `MutableHandle` isn't then used to read or write the null pointer).

Nothing in the public API changes.

### DIFF
--- a/mozjs/src/gc/root.rs
+++ b/mozjs/src/gc/root.rs
@@ -199,7 +199,10 @@ impl<'a, T> Deref for Handle<'a, T> {
 
 impl<'a, T> MutableHandle<'a, T> {
     pub unsafe fn from_marked_location(ptr: *mut T) -> Self {
-        MutableHandle::new(&mut *ptr)
+        Self {
+            ptr,
+            anchor: PhantomData,
+        }
     }
 
     pub unsafe fn from_raw(handle: RawMutableHandle<T>) -> Self {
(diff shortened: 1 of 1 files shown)


sagudev:
> I don't think that's happening based on a quick test with putting an assert in and running servo's tests

Did you test both Handle and MutableHandle?

sagudev:
> I did. Nothing in servo's unit tests for sure.

Not sure if unit test exercise any relevant code paths.

> I think nothing in servo's wpt tests but even on a clean build of servo I get so many differences running those locally from the blessed CI version that I don't want to swear I'm not missing something there.

Can you do try run in CI with `mach try wpt`, that's how we usually get useful results.

gmorenz:
> Can you do try run in CI with `mach try wpt`, that's how we usually get useful results.

I don't think I can? Or if I can I don't know how to. I'm not part of the servo organization and have no permissions.

I've pushed the branch here if you want to run it (not that it's a complicated change to reproduce): https://github.com/gmorenz/servo/tree/check_null

Pointing to the mozjs commit here: https://github.com/gmorenz/mozjs/tree/check_null

gmorenz:
Thanks @jdm for the pointer - I'd seen the page but missed the flag to run it against my own remote (and my habit is to make `origin` the upstream repo).

The linux-wpt tests pass with the asserts in place. https://github.com/gmorenz/servo/actions/runs/20531474798 if anyone wants to verfiy.

PR #694, opened 2026-01-02
Make `clamp_to` compliant with the WebIDL specification

Taym95:
This PR fixes the implementation of WebIDL clamp numeric conversions to match the behavior required by the WebIDL specification and [Firefox](https://searchfox.org/firefox-main/rev/aee7c0f24f488cd7f5a835803b48dd0c0cb2fd5f/dom/bindings/PrimitiveConversions.h#225-269).

Companion PR: https://github.com/servo/servo/pull/41640

### DIFF
--- a/mozjs/src/conversions.rs
+++ b/mozjs/src/conversions.rs
@@ -46,6 +46,7 @@ use crate::rust::{ToBoolean, ToInt32, ToInt64, ToNumber, ToUint16, ToUint32, ToU
 use libc;
 use log::debug;
 use mozjs_sys::jsgc::Rooted;
+use num_traits::PrimInt;
 use std::borrow::Cow;
 use std::mem;
 use std::ptr::NonNull;
@@ -235,7 +236,7 @@ where
 /// integer WebIDL type. Using it with non-integer types would be incorrect.
 fn clamp_to<D>(d: f64) -> D
 where
-    D: Number + As<f64>,
+    D: Number + PrimInt + As<f64>,
     f64: As<D>,
 {
     // NaN maps to zero.
@@ -267,7 +268,7 @@ where
         // number or we have an odd number but the number we want is one closer to
         // 0. So just unconditionally masking out the ones bit should do the trick
         // to get us the value we want.
-        truncated = (((truncated.cast() as i64) & !1) as f64).cast();
+        truncated = truncated & !D::one();
     }
 
     truncated
@@ -324,7 +325,7 @@ unsafe fn convert_int_from_jsval<T, M>(
     convert_fn: unsafe fn(*mut JSContext, HandleValue) -> Result<M, ()>,
 ) -> Result<ConversionResult<T>, ()>
 where
-    T: Number + As<f64>,
+    T: Number + As<f64> + PrimInt,
     M: Number + As<T>,
     f64: As<T>,
 {
--- a/mozjs/Cargo.toml
+++ b/mozjs/Cargo.toml
@@ -27,6 +27,7 @@ encoding_rs = "0.8.35"
 libc.workspace = true
 log = "0.4"
 mozjs_sys = { version = "=0.140.5-7", path = "../mozjs-sys" }
+num-traits = "0.2"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 criterion = { version = "0.6", default-features = false, features = ["html_reports"] }


Taym95:
> > Please restore the PR template and also explain what the fix is that you made in the PR title and description.
> 
> Apologies, `mozjs` does not have a PR template, but I think it's helpful to write more than "fix" in the PR title and description. Please describe the problem and how these changes fix the problem.

The underlying issue is already fully discussed in the referenced issue, which is linked from this PR.

This mozjs PR is scoped to the mozjs change, I opened a corresponding Servo [PR](https://github.com/servo/servo/pull/41640) respecting the PR template and additional context.

I’ve also added more detail to this PR description.

mozjs/src/conversions.rs
```diff
@@ -232,15 +232,39 @@ where
...
+    let max = D::MAX.cast();
+    let min = D::MIN.cast();
+
+    // clamp
+    if d >= max {
+        return D::MAX;
+    }
+    if d <= min {
+        return D::MIN;
     }
```
mrobinson: This can just be:

```rust
    if d.is_nan() {
       return D::ZERO.
   }
    if d >= D:MAX.cast() {
        return D::MAX;
    }
    if d < D::MAX.cast() {
        return D::MIN;
    }

File: mozjs/src/conversions.rs [hunk 2]
```diff
@@ -232,15 +232,39 @@ where
...
+    let mut truncated: D = to_truncate.cast();
+
+    // Tie case: exact 0.5
+    if truncated.cast() == to_truncate {
+        // go through f64 -> i64 -> even -> f64 -> D
+        let i = to_truncate as i64;
+        let even = i & !1;
+        truncated = (even as f64).cast();
+    }
+
```
mrobinson, line 273: Since this function is now a direct port of the Gecko code, can you please put a permalink to the functionin the rustdoc of the method. Something like:

```rust
///
/// This function is a port of the `PrimitiveConversionTraits_Clamp` [found in Gecko].
/// [found in Gecko]: https://searchfox.org/firefox-main/rev/aee7c0f24f488cd7f5a835803b48dd0c0cb2fd5f/dom/bindings/PrimitiveConversions.h#226
```

In addition, please preserve the code comments from the original function in Gecko as they are extremely helpful to understand what is going on. All of this is fine because both files have the same open source license.

mrobinson (dismissed):

mozjs/src/conversions.rs
```diff
@@ -232,15 +232,39 @@ where
...
+    let to_truncate = if d < 0.0 { d - 0.5 } else { d + 0.5 };
+
+    // Truncate toward zero
+    let mut truncated: D = to_truncate.cast();
+
+    // Tie case: exact 0.5
+    if truncated.cast() == to_truncate {
+        // go through f64 -> i64 -> even -> f64 -> D
+        let i = to_truncate as i64;
+        let even = i & !1;
```
mrobinson: I think it's fine to do this in one line and avoid temporaries.
```suggestion
        truncated = ((to_truncate as i64 & !1)  as f64).cast();
```

mrobinson (commented):
Oh, I think it also makes sense to link to the specification in the rustdoc as well. Another option would be to use specification text, but whatever you like is fine. There just needs to be an explanation of what's going on here as this is quite tricky code.

mozjs/src/conversions.rs
```diff
@@ -224,44 +224,48 @@ where
...
-        let i = to_truncate as i64;
-        let even = i & !1;
-        truncated = (even as f64).cast();
+        /*
+         * It was a tie (since moving away from 0 by 0.5 gave us the exact integer
+         * we want). Since we rounded away from 0, we either already have an even
+         * number or we have an odd number but the number we want is one closer to
+         * 0. So just unconditionally masking out the ones bit should do the trick
+         * to get us the value we want.
+         */
```
mrobinson: Let's keep the comment style consistent in the function by switching to `//`.

mozjs/src/conversions.rs
```diff
@@ -224,23 +224,51 @@ where
...
 }
 
-/// Try to cast the number to a smaller type, but if it doesn't fit,
-/// round it to the MAX or MIN of the source type before casting it to
-/// the destination type.
+/// WebIDL ConvertToInt (Clamp) conversion.
+/// Spec: https://webidl.spec.whatwg.org/#abstract-opdef-converttoint
+///
+/// This function is ported from Gecko’s `PrimitiveConversionTraits_Clamp`.
+/// [Gecko source]: https://searchfox.org/firefox-main/rev/aee7c0f24f488cd7f5a835803b48dd0c0cb2fd5f/dom/bindings/PrimitiveConversions.h#226
```
mrobinson: The markdown link from my suggestion is now broken as the text inside `[...]` has to match some from a link in the preceding paragraph. You can fix it with:


```suggestion
/// This function is ported from Gecko’s [`PrimitiveConversionTraits_Clamp`].
///
/// [`PrimitiveConversionTraits_Clamp`]: https://searchfox.org/firefox-main/rev/aee7c0f24f488cd7f5a835803b48dd0c0cb2fd5f/dom/bindings/PrimitiveConversions.h#226
```

By the way if you use an IDE like Visual Studio Code or Zed, if you hover over the function name it should give you a formatted view of the rustdoc and let you test clicking on links.

mozjs/src/conversions.rs
```diff
@@ -224,23 +224,51 @@ where
...
+
+    if truncated.cast() == to_truncate {
+        /*
+         * It was a tie (since moving away from 0 by 0.5 gave us the exact integer
+         * we want). Since we rounded away from 0, we either already have an even
+         * number or we have an odd number but the number we want is one closer to
+         * 0. So just unconditionally masking out the ones bit should do the trick
+         * to get us the value we want.
+         */
+        truncated = (((to_truncate as i64) & !1) as f64).cast();
```
mrobinson: Hrm. Shouldn't you be truncating `truncate` here instead of `to_truncate`? `to_truncate` has been rounded, but not truncated via the cast.

```suggestion
        truncated = truncated &= !1;
```

With that change does Servo still pass the tests?


If that doesn't work I also wonder about the as f64 here. Theoretically this function should only be called if `D` is an integral type. Gecko ensures this with a static cast, but that doesn't seem to be possible in Rust unfortunately. If the above doesn't work, does the following?

```suggestion
        truncated = ((to_truncate as i64) & !1).cast();
```

mrobinson (commented):
Sorry. I have a doubt about the code itself that I didn't notice before. Otherwise this seems good as it is passing tests.

mozjs/src/conversions.rs
Taym95: No — to_truncate is intentional here and matches the Gecko implementation,the tie check must compare against the value after the ±0.5 adjustment and truncation step, using the already cast value would lose the information needed to detect exact .5 ties.

I ran the Servo tests before committing, and they are passing with this change.

mozjs/src/conversions.rs
Taym95: I followed Gecko and copied the comments as is, I will switch them to // for consistency.

Taym95:
> Sorry. I have a doubt about the code itself that I didn't notice before. Otherwise this seems good as it is passing tests.

I don’t think there is any remaining doubt about the code. It follows the Gecko implementation and the WPT tests are passing.

mozjs/src/conversions.rs
mrobinson: Here is the Gecko code:

```c++
 T truncated = static_cast<T>(toTruncate);

  if (truncated == toTruncate) {
    /*
     * It was a tie (since moving away from 0 by 0.5 gave us the exact integer
     * we want). Since we rounded away from 0, we either already have an even
     * number or we have an odd number but the number we want is one closer to
     * 0. So just unconditionally masking out the ones bit should do the trick
     * to get us the value we want.
     */
    truncated &= ~1;
  }
```

The value that is modified here is `truncated` which has already been cast to the final type. I believe that Gecko code is trying to compare the rounded value to the rounded+cast value and then modify the rounded+cast value if they are equal. In your case you are comparing the rounded value to the rounded+cast value and then modifying the rounded (but not cast value). Instead you cast to i64, which I think is redoing the ~~rounding~~ cast but to a type of a potentially different size than `D`.

mrobinson: FWIW, I think this might work fine when doing this operation from `f64` to `i64`, but I suspect it will not work exactly the same was as Gecko in the general case. That could potentially be why this still passes tests.

Taym95: I think I see your concern, In Gecko the low bit is cleared on the already cast value, while in the Rust version the even adjustment is computed via an i64 intermediate and then cast back to D, right?

given the WebIDL bounds, this produces the same result today, but I agree the difference can be confusing since the comments are copied from Gecko. I can adjust the code to operate directly on truncated if that’s clearer.

mrobinson: Yes, I think we should just use the truncated value here and then maybe put a warning in the rustdoc saying that the function should only be used when the target is an integer type. Thank you!

mozjs/src/conversions.rs
```diff
@@ -224,23 +224,53 @@ where
...
+
+    let mut truncated: D = to_truncate.cast();
+
+    if truncated.cast() == to_truncate {
+        // It was a tie (since moving away from 0 by 0.5 gave us the exact integer
+        // we want). Since we rounded away from 0, we either already have an even
+        // number or we have an odd number but the number we want is one closer to
+        // 0. So just unconditionally masking out the ones bit should do the trick
+        // to get us the value we want.
+        truncated = (((truncated.cast() as i64) & !1) as f64).cast();
```
mrobinson: So this is:

1. Converting an integer to a float (I *think* this is what `truncated.cast()` does)
2. Converting the float back to an integer. (`as i64`)
3. Doing a bitwise operation on the integer. (`& !1`)
4. Converting the integer back to a float. (`as f64`)
5. Converting that float back to an integer. (`.cast()`)

In the Gecko code it directly does the bitwise operation on the integer. 😦 I wish we could do the same here. Two options:

1. Do the truncation and bit flipping in i64 and then convert at the end. We cannot even do this easily as the various integer types (such as `u8`) do not implement `As<i64>` This could be done with `TryInto` but I don't know the performance implications or how performance sensitive this code is. 
2. Ensure that `D` implements the necessary bit operations as well as `num-traits` crate's `One` trait. This might work, but I have not tried it.

Maybe someone with a better grasp of Rust generics can help here.

mrobinson: In the meantime I've dismissed my review so it doesn't block this change.

Taym95: Yes `truncated = truncated & !D::one();` works here.

sagudev:
@Taym95 It is useful to link companion servo PR in this PR description.

Taym95:
It seems there isn’t a PR template or a defined process here yet, Having one similar to what we use in the Servo repo, could help set expectations and avoid this kind of back and forth.

sagudev:
> It seems there isn’t a PR template or a defined process here yet, Having one similar to what we use in the Servo repo, could help set expectations and avoid this kind of back and forth.

I've opened https://github.com/servo/mozjs/pull/696

PR #697, opened 2026-01-15
Support using rooted macro with vectors.

jdm:
We created our `RootedVec` type back before we supported storing arbitrary traceable types in SpiderMonkey's `Rooted` class. Now that we can do that, we can replicate SpiderMonkey's `RootedVec`, which is just a Rooted wrapper around a vector.

Testing: New unit test added.
Servo PR: https://github.com/servo/servo/pull/41921

### DIFF
--- a/mozjs/Cargo.toml
+++ b/mozjs/Cargo.toml
@@ -26,7 +26,7 @@ crown = ["mozjs_sys/crown"]
 encoding_rs = "0.8.35"
 libc.workspace = true
 log = "0.4"
-mozjs_sys = { version = "=0.140.5-8", path = "../mozjs-sys" }
+mozjs_sys = { version = "=0.140.5-9", path = "../mozjs-sys" }
 num-traits = "0.2"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]


mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        // Safety: No GC occurs during take call
+        unsafe { self.as_mut().push(value) }
+    }
+
+    pub fn extend(&mut self, iterator: impl Iterator<Item = T>) {
+        // Safety: No GC occurs during take call
+        unsafe { self.as_mut().extend(iterator) }
+    }
+
+    pub fn set_index(&mut self, index: usize, value: T) {
```
jdm, line 152: The fact that we don't implement DerefMut for RootedGuard (#572) makes `Rooted<Vec<T>>`  less ergonomic than `RootedVec<T>`, which makes me think we should keep RootedVec around and document that it should only be used with non-SpiderMonkey types which require more complete `Vec<T>` API access.

mozjs/src/gc/collections.rs
```diff
@@ -49,6 +49,15 @@ impl From<&RootedVec<'_, JSVal>> for JS::HandleValueArray {
...
 }
 
+impl From<&super::RootedGuard<'_, Vec<JSVal>>> for JS::HandleValueArray {
+    fn from(vec: &super::RootedGuard<'_, Vec<JSVal>>) -> JS::HandleValueArray {
+        JS::HandleValueArray {
+            length_: vec.len(),
+            elements_: vec.deref().as_ptr(),
+        }
+    }
+}
```
sagudev, line 59: I think it would make more sense to `impl RootedGuard<'_, Vec<JSVal>>` as that's how we do other stuff.

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        // Safety: No GC occurs during take call
+        unsafe {
+            *self.as_mut().index_mut(index) = value;
+        }
+    }
+
+    pub fn handle_at(&self, index: usize) -> Handle<'_, T> {
```
sagudev: This already happens, but I want to be explicit:
```suggestion
    pub fn handle_at(&'_ self, index: usize) -> Handle<'_, T> {
```

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        assert!(index < self.len());
+        // Safety: Any value within this rooted vector is marked automatically.
+        unsafe { Handle::from_marked_location(self.deref().as_ptr().add(index)) }
+    }
+
+    pub fn handle_mut_at(&mut self, index: usize) -> MutableHandle<'_, T> {
```
sagudev:
```suggestion
    pub fn handle_mut_at(&'_ mut self, index: usize) -> MutableHandle<'_, T> {
```

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
     }
 }
 
+impl<'a, T> RootedGuard<'a, Vec<T>>
+where
+    Vec<T>: RootKind,
+{
+    pub fn push(&mut self, value: T) {
+        // Safety: No GC occurs during take call
```
sagudev:
```suggestion
        // Safety: No GC occurs during this call
```

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        unsafe { self.as_mut().push(value) }
+    }
+
+    pub fn extend(&mut self, iterator: impl Iterator<Item = T>) {
+        // Safety: No GC occurs during take call
```
sagudev:
```suggestion
        // Safety: No GC occurs during this call
```

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        unsafe { self.as_mut().extend(iterator) }
+    }
+
+    pub fn set_index(&mut self, index: usize, value: T) {
+        // Safety: No GC occurs during take call
```
sagudev:
```suggestion
        // Safety: No GC occurs during this call
```

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
...
+        unsafe {
+            *self.as_mut().index_mut(index) = value;
+        }
+    }
+
+    pub fn handle_at(&self, index: usize) -> Handle<'_, T> {
+        assert!(index < self.len());
+        // Safety: Any value within this rooted vector is marked automatically.
```
sagudev:
```suggestion
        // Safety: Values within this rooted vector are traced.
```

mozjs/src/gc/root.rs
sagudev: Hm, I wonder if RootedVec is actually safe/sound. I have not reviewed it's code yet.

sagudev (approved):

mozjs/src/gc/root.rs
jdm: It should be; I believe we're only concerned about the rooted values moving around in memory, and non-GC values won't do that.

mozjs/src/gc/root.rs
```diff
@@ -85,6 +85,40 @@ where
     }
 }
 
+impl<'a, T> RootedGuard<'a, Vec<T>>
+where
+    Vec<T>: RootKind,
+{
+    pub fn push(&mut self, value: T) {
```
Taym95, line 142: I think this is what I was trying to do here https://github.com/servo/mozjs/pull/630/changes#diff-737e12000758c48503126ab75c14db2c9d0854a95487ebdd0a51d99f991822efR289 but I ended up creating a whole new `struct DynamicValueArray` and `struct FixedValueArray`.

mozjs/src/gc/collections.rs
jdm: Can you give an example of what you mean? There's a matching impl for RootedVec right above this one.

jdm: Similarly, https://github.com/servo/mozjs/blob/369880c95717069a18fc221ac985c453d1330ec3/mozjs-sys/src/jsimpls.rs#L142-L167

sagudev: Above is impl for `RootedVec`, but `RootedGuard` usually have `handle()` and `handle_mut()` methods so here it would be smth like:
```rust
impl<'a,> RootedGuard<'a, Vec<JSVal>>
{
fn handle_value_array(&self) -> JS::HandleValueArray {
        JS::HandleValueArray {
            length_: self.len(),
            elements_: self.deref().as_ptr(),
        }
    }
}
```

sagudev: Hm, given that HandleValueArray has no lifetime associated it can outlive the root so it should be unsafe (thus all from conversions are unsound).

jdm: For the purposes of this PR, I'd rather match the existing HandleValueArray creation code that I linked in https://github.com/servo/mozjs/pull/697/changes/BASE..60cd1d696448318e06ec28f103b30641c9c8ffa6#r2700733569. If we want to change that, I think we should change all of them together.

sagudev: fair

PR #698, opened 2026-01-15
fix: make windows-arm64 build

burhankhanzada:
- enable aarch64 on windows on arm

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,15 +140,6 @@ jobs:
         run: |
           curl -SL "https://github.com/servo/servo-build-deps/releases/download/msvc-deps/moztools-4.0.zip" --create-dirs -o target/dependencies/moztools.zip
           cd target/dependencies && unzip -qo moztools.zip -d .
-      # - name: Install LLVM 19
-      #   run: |
-      #     # MSVC has headers that require clang-19.
-      #     # remove when windows runner updates to llvm 19
-      #     iwr -useb get.scoop.sh -outfile 'install.ps1'
-      #     .\install.ps1 -RunAsAdmin
-      #     scoop install llvm@19.1.7 --global
-      #     # Prepend to path so we override the default LLVM installation in priority
-      #     echo "C:\ProgramData\scoop\apps\llvm\current\bin;" + (Get-Content $env:GITHUB_PATH -Raw) | Set-Content $env:GITHUB_PATH
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.9
       - name: Build Windows


.github/workflows/build.yml
```diff
@@ -148,19 +150,19 @@ jobs:
       - name: Build Windows
         shell: cmd
         run: |
-          cargo +${{ steps.toolchain.outputs.name }} build --verbose --target ${{ matrix.target }} --features ${{ matrix.features }}
+          cargo +${{ steps.toolchain.outputs.name }} build --verbose --target ${{ matrix.platform.target }} --features ${{ matrix.features }}
       - name: Test Windows
-        if: ${{ !contains(matrix.target, 'aarch64') }}
+        if: ${{ !contains(matrix.platform.target, 'aarch64') }}
```
sagudev: we should just remove this if as we can run arm64 test on arm64 runner

.github/workflows/release-check.yml
sagudev, line 30: we should add arm64 here too, given that we now make artifacts

.github/workflows/build.yml
```diff
@@ -148,19 +150,19 @@ jobs:
...
+          cargo +${{ steps.toolchain.outputs.name }} test --tests --examples --verbose --target ${{ matrix.platform.target }} --features ${{ matrix.features }}
 
       - name: Upload artifact
-        if: ${{ !contains(matrix.target, 'aarch64') && matrix.features != 'debugmozjs' && env.NEW_RUST_CHECK == 'false' }}
+        if: ${{ matrix.features != 'debugmozjs' && env.NEW_RUST_CHECK == 'false' }}
         uses: actions/upload-artifact@v4
         with:
-          path: ./target/${{ matrix.target }}/libmozjs-x86_64-pc-windows-msvc.tar.gz
-          name: libmozjs-x86_64-pc-windows-msvc.tar.gz
+          path: ./target/libmozjs-${{ matrix.platform.target }}.tar.gz
```
sagudev: IIRC if we pass `--target` to cargo it will always put stuff in `target/${target}` folder

burhankhanzada: well i think i test this on my fork and its put it on `target/*` like also android and other builds.

sagudev: it does not work: https://github.com/servo/mozjs/actions/runs/21094308964/job/60669761566#step:8:30

you can also see this by reading cargo invocations: https://github.com/servo/mozjs/actions/runs/21094308964/job/60669761566#step:6:1137

burhankhanzada: yeah you were right

burhankhanzada:
@sagudev do you know why its not showing any log for https://github.com/servo/mozjs/actions/runs/21094818107/job/60671635602?pr=698#logs

sagudev:
you also need to bump mozjs version er else CI will not be happy

.github/workflows/build.yml
```diff
@@ -138,33 +136,32 @@ jobs:
...
-          echo "C:\ProgramData\scoop\apps\llvm\current\bin;" + (Get-Content $env:GITHUB_PATH -Raw) | Set-Content $env:GITHUB_PATH
+      # - name: Install LLVM 19
+      #   run: |
+      #     # MSVC has headers that require clang-19.
+      #     # remove when windows runner updates to llvm 19
+      #     iwr -useb get.scoop.sh -outfile 'install.ps1'
+      #     .\install.ps1 -RunAsAdmin
+      #     scoop install llvm@19.1.7 --global
+      #     # Prepend to path so we override the default LLVM installation in priority
+      #     echo "C:\ProgramData\scoop\apps\llvm\current\bin;" + (Get-Content $env:GITHUB_PATH -Raw) | Set-Content $env:GITHUB_PATH
```
sagudev: We can remove this now.

PR #700, opened 2026-01-16
Add safe methods to handles/rootedguard and simplify NoGC

sagudev:
Firstly we simplify `NoGC` to be proper ZST without any inner lifetime. One should now use `no_gc: &NoGC` (this gives us compiler reborrow support, instead of manual reborrows we need to do for handles). This allows us to impl deref from `&JSContext` so `&JSContext` and `&NoGC` can be used interchangeably.

Then we add `as_ref` and `as_mut_ref` as the safe way (as they require `&NoGC`, aliasing rules are uphold by &/&mut of handles/root) of obtaining reference to rooted value. This should eventually replace unsound derefs and unsafe as_mut, but until then documentation notice will suffice.

Testing: compile doc tests
Servo PR: https://github.com/servo/servo/pull/41971

This is not reviewable per commit.

### DIFF
--- a/mozjs/tests/rootable.rs
+++ b/mozjs/tests/rootable.rs
@@ -61,7 +61,11 @@ fn rooting() {
         assert_eq!(some_container.val.to_object(), prototype_proto.get());
         assert_eq!(
             some_container.val,
-            some_optional_container.as_ref(cx).as_ref().unwrap().val
+            some_optional_container
+                .as_ref(context)
+                .as_ref()
+                .unwrap()
+                .val
         );
     }
 }


mozjs/src/gc/root.rs
```diff
@@ -54,7 +55,56 @@ impl<'a, T: 'a + RootKind> RootedGuard<'a, T> {
...
+    /// fn f(cx: &mut JSContext, obj: *mut JSObject) {
+    ///     rooted!(&in(cx) let mut root = obj);
+    ///     let r = root.as_ref(cx);
+    ///     gc(cx); // cannot call gc while r (thus cx borrow) is alive
+    ///     drop(r); // otherwise rust automatically drops r before gc call
+    /// }
+    /// ```
+    pub fn as_ref<'s: 'r, 'cx: 'r, 'r>(&'s self, _no_gc: &'cx NoGC) -> &'r T
+    where
+        'a: 's,
```
sagudev, line 77: for those who always forget this syntax (like me): `'a: 'b` is `'a >= 'b`

sagudev:
this will need companion PR in servo because as_ref, as_mut_ref can shadow methods that come via deref.

PR #705, opened 2026-02-15
Return a slice when querying contents of a typed array

simonwuelker:
In rust there is no need to pass `(*mut T, usize)` around, as that is just a `*mut [T]`. This makes out-of-bounds access to typed arrays harder to do by accident.

Testing: Behaviour should be unchanged, no new tests are written
Companion PR: https://github.com/servo/servo/pull/42688

### DIFF
--- a/mozjs/Cargo.toml
+++ b/mozjs/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mozjs"
 description = "Rust bindings to the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.16.4"
+version = "0.17.0"
 authors = ["The Servo Project Developers"]
 license.workspace = true
 edition.workspace = true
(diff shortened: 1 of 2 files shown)


sagudev (approved):
LGTM, but such change should get companion PR in servo before landing (so we do not land something that borks servo).

CI is blocked by https://github.com/jidicula/clang-format-action/issues/267

sagudev:
> > LGTM, but such change should get companion PR in servo before landing (so we do not land something that borks servo).
> > CI is blocked by [jidicula/clang-format-action#267](https://github.com/jidicula/clang-format-action/issues/267)
> 
> Servo currently depends on the crates.io version of mozjs and I think we want to keep it that way - is releasing a new version for this change okay?
> 
> This is a breaking change, but it doesn't require any changes to servo.

Yes, that's okay.

jschwe:
Looks like this PR fell of the radar, should we try and get this merged again?

PR #718, opened 2026-03-15
impl jsstr_to_string and latin1_to_string using &JSContext

sagudev:
Testing: Change is covered by existing tests.

### DIFF
--- a/mozjs/src/conversions.rs
+++ b/mozjs/src/conversions.rs
@@ -597,7 +597,7 @@ impl FromJSValConvertible for f64 {
 
 /// Converts a `JSString`, encoded in "Latin1" (i.e. U+0000-U+00FF encoded as 0x00-0xFF) into a
 /// `String`.
-pub fn latin1_to_string_safe(cx: &crate::context::JSContext, s: NonNull<JSString>) -> String {
+pub fn latin1_to_string(cx: &crate::context::JSContext, s: NonNull<JSString>) -> String {
     assert!(unsafe { JS_DeprecatedStringHasLatin1Chars(s.as_ptr()) });
 
     let mut length = 0;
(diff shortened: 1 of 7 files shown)


mozjs/src/conversions.rs
```diff
@@ -620,18 +621,37 @@ pub unsafe fn latin1_to_string(cx: *mut JSContext, s: NonNull<JSString>) -> Stri
...
-    let char_vec = slice::from_raw_parts(chars, length as usize);
+    let char_vec = unsafe { slice::from_raw_parts(chars, length as usize) };
     String::from_utf16_lossy(char_vec)
 }
 
+/// Converts a `JSString`, encoded in "Latin1" (i.e. U+0000-U+00FF encoded as 0x00-0xFF) into a
+/// `String`.
+///
+/// Use [`latin1_to_string_safe`] if possible as this function will be eventually removed.
+pub unsafe fn latin1_to_string(cx: *mut JSContext, s: NonNull<JSString>) -> String {
```
jschwe: Should we perhaps add a `deprecated` annotation on it, or call it `_legacy` to make the comment about "will eventually be removed" more obvious at the usage sites?

sagudev: The way we are currently doing is just adding comments without renames/deprecation notice for less friction in servo. But eventually we should start doing this.

Gae24: There are only a couple of places left where we can't switch to the safe version yet; so with a companion pr that migrates the other instances to the safe version, we can start deprecating the unsafe one.

sagudev: I did this now, but I do not have time for companion PR.

mozjs/src/conversions.rs
```diff
@@ -596,12 +597,12 @@ impl FromJSValConvertible for f64 {
 
 /// Converts a `JSString`, encoded in "Latin1" (i.e. U+0000-U+00FF encoded as 0x00-0xFF) into a
 /// `String`.
-pub unsafe fn latin1_to_string(cx: *mut JSContext, s: NonNull<JSString>) -> String {
-    assert!(JS_DeprecatedStringHasLatin1Chars(s.as_ptr()));
+pub fn latin1_to_string(cx: &crate::context::JSContext, s: NonNull<JSString>) -> String {
```
jschwe: Technically I guess this function should still be unsafe, because `s` could be a dangling pointer.

Gae24:
Also need to do a version bump.

jschwe (commented):
Sorry, this PR fell off my radar completely. My concern is that I'm not familiar enough with the bindings to be able to judge if assuming that `NonNull<JSString>` will always be live is an acceptable tradeoff we can make. Maybe someone more familiar with the bindings can review.

mozjs/src/conversions.rs
sagudev: I fully agree. We want to be strict.

jschwe (approved):

jdm (approved):

PR #723, opened 2026-03-23
jsglue: add a method to retrieve ErrorInfo from the pending exception stack

Gae24:
Attempt to retrieve `ErrorInfo` from the stack of the pending exception, to avoid empty messages on uncaught exceptions.

Testing: Need to test it with servo.
Servo PR: [servo/43632](https://github.com/servo/servo/pull/43632)

### DIFF
--- a/mozjs-sys/src/jsglue.cpp
+++ b/mozjs-sys/src/jsglue.cpp
@@ -1232,6 +1232,42 @@ bool DescribeScriptedCaller(JSContext* cx, char* buffer, size_t buflen,
   return true;
 }
 
+typedef void (*StringCallback)(const char* ptr, size_t len, void* target);
+
+bool PendingExceptionStackInfo(JSContext* cx, StringCallback callback,
+                               void* message_target, void* filename_target,
+                               uint32_t* line, uint32_t* col,
+                               JS::MutableHandleValue dest) {
+  JS::ExceptionStack stack(cx);
+  JS::ErrorReportBuilder builder(cx);
+  if (JS::StealPendingExceptionStack(cx, &stack) &&
+      builder.init(cx, stack, JS::ErrorReportBuilder::WithSideEffects)) {
+    JSErrorReport* aReport = builder.report();
+
+    if (!aReport || aReport->isWarning()) {
+      return false;
+    }
+
+    const char* message = aReport->message().c_str();
+    if (!message) {
+      message = builder.toStringResult().c_str();
+    }
+
+    callback(message, strlen(message), message_target);
+
+    const char* filename = aReport->filename.c_str();
+    callback(filename, strlen(filename), filename_target);
+
+    *line = aReport->lineno;
+    *col = aReport->column.oneOriginValue();
+    dest.set(stack.exception());
+
+    return true;
+  }
+
+  return false;
+}
+
 void SetDataPropertyDescriptor(JS::MutableHandle<JS::PropertyDescriptor> desc,
                                JS::HandleValue value, uint32_t attrs) {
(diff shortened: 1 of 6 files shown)


mozjs/src/rust.rs
```diff
@@ -1056,6 +1057,38 @@ pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCal
...
+pub unsafe fn error_info_from_exception_stack(cx: *mut JSContext) -> Option<ErrorInfo> {
+    let mut message_buf = [0; 1024];
+    let mut filename_buf = [0; 1024];
+    let mut line = 0;
+    let mut col = 0;
+    if !PendingExceptionStackInfo(
+        cx,
+        message_buf.as_mut_ptr(),
+        filename_buf.as_mut_ptr(),
+        1024,
```
Taym95: magic number

mozjs-sys/src/jsglue.cpp
```diff
@@ -643,6 +643,44 @@ bool ShouldMeasureObject(JSObject* obj, nsISupports** iface) {
...
+      return false;
+    }
+
+    const char* message = aReport->message().c_str();
+    const char* filename = aReport->filename.c_str();
+
+    if (!message) {
+      message = builder.toStringResult().c_str();
+    }
+    *message_len = std::min<size_t>(1024, strlen(message));
```
sagudev: I think this might reasonable strategy for filename, but not for message. Also we are doing two copies, one to message_buffer and one with to_string. Maybe we should impl this with callbacks, that receive len and char* and create rust string from it.

We might also skip utf8 validation if I understand https://searchfox.org/firefox-main/source/js/public/CharacterEncoding.h#135 correctly.

Gae24: Sorry could you expand on it? Is the issue that the selected length could be insufficient?

sagudev: Yes and the fact that implementation is not optimal we copy it twice, firstly into rust array then into rust String.

mozjs/src/rust.rs
```diff
@@ -1056,6 +1058,56 @@ pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCal
...
+
+unsafe extern "C" fn fill_string_callback(ptr: *const c_char, len: usize, target: *mut c_void) {
+    let target = &mut *(target as *mut String);
+
+    if ptr.is_null() {
+        return;
+    }
+
+    let slice = slice::from_raw_parts(ptr as *const u8, len);
+    *target = str::from_utf8_unchecked(slice).to_owned();
```
sagudev: This looks sus, we could already have alloced string and this would replace it with new alloc (String::new is without alloc tho). But wouldn't https://doc.rust-lang.org/std/string/struct.String.html#method.push_str just be nicer and more explicit.

Need to think about this tomorrow.

mozjs/src/rust.rs
```diff
@@ -1056,6 +1058,56 @@ pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCal
...
+    let mut filename = String::new();
+
+    let mut line = 0;
+    let mut col = 0;
+
+    if !PendingExceptionStackInfo(
+        cx,
+        Some(fill_string_callback),
+        &mut message as *mut String as *mut c_void,
+        &mut filename as *mut String as *mut c_void,
```
```suggestion
        &raw mut message as *mut c_void,
        &raw mut filename as *mut c_void,
```

sagudev (approved):

PR #729, opened 2026-04-08
Security bump SpiderMonkey to 140.9

sagudev:
Bump SpiderMonkey to 140.9 (changeset: [cca515379f569261e903f6e84dcdb0ecdbbe821e](https://github.com/servo/mozjs/releases/download/mozjs-source-cca515379f569261e903f6e84dcdb0ecdbbe821e))


mozjs-sys/mozjs/config/milestone.txt
```diff
@@ -10,4 +10,4 @@
 # hardcoded milestones in the tree from these two files.
 #--------------------------------------------------------
 
-140.8.0
+140.10.0
```
jschwe, line 13: Is this 140.10, or is the number here misleading?

jdm (approved):

mozjs-sys/mozjs/config/milestone.txt
sagudev: Hm, this is weird, the commit from COMMIT file has different content: https://hg-edge.mozilla.org/releases/mozilla-esr140/file/cca515379f569261e903f6e84dcdb0ecdbbe821e/config/milestone.txt

sagudev: Ah, the problem is that commits are on separate branches (it's dot release) so we should not just walking back on pushes is not enough ...

sagudev:
I need to fix bump script again.

PR #731, opened 2026-04-12
ci: build aarch64-linux in CI

### LINKED ISSUE #730: Add prebuilt arm64 linux mozjs
This is useful when using docker (devcontainers) on an arm mac host.

jschwe:
Having the prebuilt mozjs artifact available is mainly useful for users using docker on arm Macs.

Testing: This modifies CI workflow files. 
Fixes: #730

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,11 +72,14 @@ jobs:
       RUSTC_WRAPPER: "sccache"
       CCACHE: sccache
       SCCACHE_GHA_ENABLED: "true"
-    runs-on: ubuntu-22.04 # Needed for artifacts to work on older systems (due to glibc)
     strategy:
       fail-fast: false
       matrix:
         features: ["debugmozjs", ""]
+        platform:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
sagudev:
```suggestion
```

sagudev: We should remove arm64 here.

sagudev (approved):
otherwise LGTM

.github/workflows/build.yml
jschwe: Not sure why, but for some reason i couldn't apply your suggestion directly. manually worked though.

sagudev: It also rendered incorrectly so thats why I wrote text in new comment

sagudev:
we didn't bump version numbers, so we did not get new releases: https://github.com/servo/mozjs/actions/runs/24312618806/job/70987127256; but we can wait for real PRs to do bumps

jschwe:
Yeah I noticed that too - but then again I can build from source just fine, so it's not urgent from my point of view either.

PR #732, opened 2026-04-18
build: Ignore CLANGFLAGS

jschwe:
CLANGFLAGS was only applied to bindgen.
cc-rs is not documented to use it, and SM also doesn't reference it. In servo I also couldn't see us setting this variable. This is a preparation for unifying our common build flags across cc-rs, SM build and bindgen.

Testing: Not tested.


jschwe:
I'm not sure if it's worth it to merge this standalone, or if I should just put it into the bigger PR.

sagudev (approved):

PR #734, opened 2026-04-19
build: Unify bindings and glue code flags with cc-rs

jschwe:
Follow-up to #732. 

Refactor cc_flags to get_common_cc, which returns a cc builder. Any bindgen specific flags are moved to build_bindings.
Removed the explicit `O0` override, this will be configured by cc-rs based on opt-level (for the glue code, and ignored by bindgen).

Testing: *Describe how this pull request is tested or why it doesn't require tests*
Servo PR: *Link to a companion pull request in the Servo repository. Remove this line if there is no companion PR.*


jschwe:
Superceded by https://github.com/servo/mozjs/pull/789

PR #735, opened 2026-04-19
build: Workaround make jobserver bug

### LINKED ISSUE #375: Parallel compilation is broken with GNU Make 4.4
GNU Make 4.4 breaks the parallel builds in mozjs and behaves as though `-j1` was given, leading to long build times.

Relevant logs from `cargo build -vv`:
```
[mozjs_sys 0.68.2] make -f Makefile
[mozjs_sys 0.68.2] make[1]: Entering directory '/home/mukilan/dev/mozjs/target/debug/build/mozjs_sys-094547916dc1fb70/out/build'
[mozjs_sys 0.68.2] make[2]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
[mozjs_sys 0.68.2] make[2]: Entering directory '/home/mukilan/dev/mozjs/target/debug/build/mozjs_sys-094547916dc1fb70/out/build'
[mozjs_sys 0.68.2] make recurse_pre-export
```

Th release notes for [GNU Make 4.4](https://lists.gnu.org/archive/html/help-make/2022-10/msg00020.html) and the linked bugs note that when using 'old-style' anonymous pipes for jobserver, the file descriptors are closed when parent make invokes a recursive sub-make when the sub-make call is *[not direct](https://www.gnu.org/software/make/manual/make.html#MAKE-Variable)* 
```
* When the pipe-based jobserver is enabled and GNU Make decides it is invoking
  a non-make sub-process and closes the jobserver pipes, it will now add a new
  option to the MAKEFLAGS environment variable that disables the jobserver.
  This prevents sub-processes that invoke make from accidentally using other
  open file descriptors as jobserver pipes.  For more information see
  https://savannah.gnu.org/bugs/?57242 and https://savannah.gnu.org/bugs/?62397
```

SpiderMonkey's Makefile seem to implement recursive invocations that are [not direct](https://searchfox.org/mozilla-central/source/config/rules.mk#233) $(MAKE) invocations, so these sub-makes fail to open the closed jobserver pipes and fallback to -j1 mode.

This issue can be fixed by using a version of [Cargo that has support](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-169-2023-04-20) for the new 'named fifo' based jobserver as GNU Make 4.4 by introduced them to allow  any process to be integrate with the jobserver and thus should work with both direct and indirect sub-make invocations.

This affects Servo (on distros that ship GNU Make 4.4 such as Fedora 39, Arch,  NixOS  23.05 etc) since it currently uses 2023-02-01 build of rust toolchain where the Cargo doesn't have support for named fifo jobserver.

jschwe:
This works around the following bug, where the invocation below fails with the pipe style and works fine with the new fifo jobserver. Since there is no sign of cargo going to support fifo jobservers, and single threaded compilation is horribly slow, we don't pass makeflags on, when using make 4.4. We use a crude version check, which checks the presence of the flag - since new make versions are rare, this should be fine for a while.
If we detect the flag (i.e. make 4.4 or newer), we pass `-j <NUM_JOBS>` to make instead of inheriting MAKEFLAGS and the jobserver from cargo. This means we get an upper bound of 2x NUM_JOBS jobs, since the cargo jobserver  and makes jobserver coexist.
We can remove this workaround once cargo [supports fifo style jobservers](https://github.com/rust-lang/cargo/issues/13483).

Minimum reproducer:

`gmake -j2 --jobserver-style=pipe -f Makefile`

Makefile:
```make
define RECURSE
+$(MAKE) -f sub.mk
endef

all:
	$(RECURSE)
```

sub.mk:

```
all:
	@:
```

Output: `gmake[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.`

Testing: This is specific to make 4.4, and hence not covered by automated tests
Fixes: #375 
Servo PR: https://github.com/servo/servo/pull/44346

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -222,8 +222,14 @@ fn build_spidermonkey(build_dir: &Path) {
     cppflags.push(get_cc_rs_env_os("CPPFLAGS").unwrap_or_default());
     cmd.env("CPPFLAGS", cppflags);
 
-    if let Some(makeflags) = env::var_os("CARGO_MAKEFLAGS") {
-        cmd.env("MAKEFLAGS", makeflags);
+    // With make 4.4 a new jobserver style was added, which cargo doesn't support yet.
+    // Unfortunately, the old pipe jobserver that cargo exposes, has a new bug in 4.4,
+    // which is exposed by spidermonkeys makefiles and makes compilation largely
+    // single-threaded.
+    if !is_buggy_make_version() {
+        if let Some(makeflags) = env::var_os("CARGO_MAKEFLAGS") {
+            cmd.env("MAKEFLAGS", makeflags);
+        }
     }
 
     let mut cxxflags = vec![];
@@ -287,6 +293,26 @@ fn cbindgen_bidi(build_dir: &Path) {
       .write_to_file(root_to_bidi(build_dir).join("unicode_bidi_ffi_generated.h"));
 }
 */
+
+fn is_buggy_make_version() -> bool {
+    if let Ok(output) = Command::new("gmake")
+        .arg("--help")
+        .output()
+        .or_else(|_| Command::new("make").arg("--help").output())
+    {
+        let Ok(output) = String::from_utf8(output.stdout) else {
+            println!("cargo:warning=Output from make was not valid utf-8. Can't determine version");
+            return false;
+        };
+        // --jobserver-style was added in Make 4.4, and hence can tell us if we have a version
+        // of make that is incompatible with cargos pipe jobserver (in spidermonkey).
+
+        output.contains("--jobserver-style")
+    } else {
+        println!("cargo:warning=Couldn't invoke make --help to determine make version");
+        false
+    }
+}
 
 fn build(build_dir: &Path, target: BuildTarget) {
     let mut build = cc::Build::new();
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -226,6 +226,8 @@ fn build_spidermonkey(build_dir: &Path) {
     // Unfortunately, the old pipe jobserver that cargo exposes, has a new bug in 4.4,
     // which is exposed by spidermonkeys makefiles and makes compilation largely
     // single-threaded.
+    // The workaround can be removed / adapted once cargo supports the new jobserver:
+    // <https://github.com/rust-lang/cargo/issues/13483>
     if is_buggy_make_version() {
         // worst-case we'll get 2xNUM_JOBS jobs, since we will have two independant job servers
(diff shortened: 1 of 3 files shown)


jschwe:
The downside is of course increased system load, since now you have two make jobservers that don't know about each other. It mostly shouldn't affect servo, since most builds will be pre-compiled. We could expose an environment variable to limit parallelism again for SM if people do run into memory / load issues.

mozjs-sys/build.rs
```diff
@@ -222,8 +222,19 @@ fn build_spidermonkey(build_dir: &Path) {
     cppflags.push(get_cc_rs_env_os("CPPFLAGS").unwrap_or_default());
     cmd.env("CPPFLAGS", cppflags);
 
-    if let Some(makeflags) = env::var_os("CARGO_MAKEFLAGS") {
-        cmd.env("MAKEFLAGS", makeflags);
+    // With make 4.4 a new jobserver style was added, which cargo doesn't support yet.
```
mukilan, line 225: Maybe you can link to the issue here?

mukilan (commented):
I don't think this PR should close #375. Either we keep that open or we should create a new issue to remove this workaround when it becomes possible to use a single job server.

mozjs-sys/build.rs
jschwe: sure, done

jschwe:
> I don't think this PR should close #375. Either we keep that open or we should create a new issue to remove this workaround when it becomes possible to use a single job server.

I think I'd prefer opening a new one, since the problem changes from "No parallism" to "potentially too much parallelism"

sagudev (approved):
LGTM, but needs updated PR description to reflect last changes (because these gets into commit message on land)

sagudev:
Uf, could you also redo try run in servo (in github hosted runners - personal fork) to make sure we do not fail there (due too more jobs).

jschwe:
> Uf, could you also redo try run in servo (in github hosted runners - personal fork) to make sure we do not fail there (due too more jobs).

https://github.com/jschwe/servo/actions/runs/24789316855/job/72542152621 - The windows MSRV build is failing, but thats expected / an existing issue because it isn't able to build mozjs from source (missing bootstrap). All other jobs are done, or past mozjs-sys, so looking good.

sagudev (approved):

sagudev:
> Uf, could you also redo try run in servo (in github hosted runners - personal fork) to make sure we do not fail there (due too more jobs).

Ugh I am stupid, CI uses old make so it will not really test this so this is not the problem at all (at least for now).

jschwe:
Ah, I didn't think of that either. But perhaps by the time we update to ubuntu 26.04 in CI, cargo might have already added the new jobserver support. (we just recently updated to 24.04 after all).

PR #736, opened 2026-04-19
Add a safe wrapper around `ForOfIterator`

simonwuelker:
`ForOfIterator` implements `for .. of` iteration in spidermonkey. It holds stack roots and is a pain to use directly due to some weird shenanigans where bindgen adds extra padding to the struct for some LLVM versions. Its also hard to use in a crash-proof manner, because `ForOfIterator` must not be moved. We'd have to `Pin<Box<>>` it to be safe, which adds overhead.

This change instead adds `mozjs::rust::for_of`, a function that takes a callback and uses `ForOfIterator` correctly so users don't have to worry about it.

Testing: The iterator is used during conversion of IDL sequences, and its covered by WPT. I'm also using it for web animations locally.
Part of https://github.com/servo/servo/issues/36950
Servo PR: https://github.com/servo/servo/pull/44356

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -9,6 +9,7 @@ use std::char;
 use std::default::Default;
 use std::ffi::{c_char, c_void, CStr, CString};
 use std::marker::PhantomData;
+use std::mem;
 use std::mem::MaybeUninit;
 use std::ops::{Deref, DerefMut};
 use std::ptr::{self, NonNull};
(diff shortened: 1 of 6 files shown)


mozjs-sys/build.rs
```diff
@@ -459,6 +459,7 @@ fn link_bindgen_static_lib_binaries(build_dir: &Path) {
 /// Check env variable conditions to decide if we need to link pre-built archive first.
 /// And then return bool value to notify if we need to build from source instead.
 fn should_build_from_source() -> bool {
+    return true;
```
simonwuelker: This is temporary to make try runs work. I assume we need to build and publish a new mozjs archive before merging this change.

sagudev: Hm, just bumping versions should work (as there are no artifacts).

simonwuelker: Ah I tried this before bumping the version, that makes sense.

simonwuelker:
Hmm need to investigate why this crashes in the `vec_conversion` test on windows.

simonwuelker:
This is ready for review now. Some jobs are not picked up by the runners, not sure what's up with that.

sagudev:
> Some jobs are not picked up by the runners, not sure what's up with that.

We are limited by number of concurrent mac runners (org wise), so it usually takes some time for them to get running. That is also way I canceled old CI runs, as they result will be thrown away anyway.

mozjs/src/rust.rs
```diff
@@ -1267,6 +1270,105 @@ impl<'a> Handle<'a, StackGCVector<*mut JSString, js::TempAllocPolicy>> {
...
+///
+/// The callback returns `Err()` to indicate a pending exception or `Ok()` containing a boolean
+/// value that is `true` if the iterator should continue iterating.
+pub fn for_of<Callback, OtherError>(
+    cx: *mut JSContext,
+    iterable: HandleValue<'_>,
+    mut callback: Callback,
+) -> Result<(), ForOfIterationFailure<OtherError>>
+where
+    Callback: FnMut(HandleValue<'_>) -> Result<bool, ForOfIterationFailure<OtherError>>,
```
```suggestion
    Callback: FnMut(HandleValue<'_>) -> Result<ControlFlow, ForOfIterationFailure<OtherError>>,
```

bool is often used for success code in SM code so that could cause confusion (it did for me) but I think we would benefit from more descriptive type. Luckily std has type for this exact situation: https://doc.rust-lang.org/beta/std/ops/enum.ControlFlow.html

> This is used when exposing things (like graph traversals or visitors) where you want the user to be able to choose whether to exit early. Having the enum makes it clearer – no more wondering “wait, what did false mean again?” – and allows including a value.

sagudev (approved):
Otherwise LGTM

mozjs/src/rust.rs
simonwuelker: Ah I knew I'd seen `Controlflow` somewhere before, but couldn't find it anymore. Thank you!

sagudev: I consider it std's hidden 💎

Issue #737, opened 2026-04-21
Build is broken with crown analysis enabled


jdm:
This was not caught by Servo's CI because we always run against downloaded binaries, so the crown build for mozjs never actually occurs in CI.

```
error: Type must be rooted.
   --> /Users/jdm/src/mozjs/mozjs-sys/src/jsgc.rs:344:27
    |
344 |     pub fn boxed(v: T) -> Box<Heap<T>>
    |                           ^^^^^^^^^^^^
    |
    = note: `#[deny(crown::unrooted_must_root)]` on by default

error: Expression of type std::boxed::Box<jsgc::Heap<T/#0>, std::alloc::Global> must be rooted.
   --> /Users/jdm/src/mozjs/mozjs-sys/src/jsgc.rs:348:13
    |
348 |         let boxed = Box::new(Heap::default());
    |             ^^^^^

error: Type must be rooted, use #[crown::unrooted_must_root_lint::must_root] on the struct definition to propagate.
     --> /Users/jdm/src/servo/target/debug/build/mozjs_sys-a12c30867b0fe764/out/build/jsapi.rs:14866:13
      |
14866 |             pub expando: root::JS::Heap<root::JS::Value>,
      |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

sagudev:
> This was not caught by Servo's CI because we always run against downloaded binaries

This does not sounds correct, only C/C++ artifacts of SM and rust source code generated by bindgen get's downloaded, so all rust code should be checked in repo. Maybe invalidation is not working again? Maybe try running `./mach check --use-crown` on clean target.

jdm:
Probably caused by https://github.com/servo/servo/issues/44630.

PR #738, opened 2026-04-21
Fix build with crown feature enabled.

jdm:
Per #737 this doesn't get detected in either this repository's CI or Servo's CI. `Heap<T>` is marked `must_root`, and method that return a `Heap<T>` need to suppress that warning. 

Testing: Servo build succeeded with MOZJS_FROM_SOURCE=1, `--use-crown`, and a local Cargo override.

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -846,6 +846,7 @@ impl BuildTarget {
                 "JS::dbg::Builder_Object",
                 "JS::dbg::Builder_Object_Base",
                 "JS::dbg::BuilderOrigin",
+                "JS::ExpandoAndGeneration",
                 "JS::RootedTuple",
                 "mozilla::external::AtomicRefCounted",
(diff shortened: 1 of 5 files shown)


mozjs-sys/build.rs
```diff
@@ -814,6 +814,7 @@ impl BuildTarget {
                 "JS::dbg::Builder_Object",
                 "JS::dbg::Builder_Object_Base",
                 "JS::dbg::BuilderOrigin",
+                "JS::ExpandoAndGeneration",
```
sagudev, line 849: this change seems unrelated

sagudev (approved):
Otherwise LGTM.

Ideally we would run crown in this repo, but this is complicated.

mozjs-sys/build.rs
jdm: It contains a Heap member so it needs annotations, but we don't actually use it.

PR #739, opened 2026-04-21
Allow extracting the box out of a RootedTraceableBox.

jdm:
This method is important for making it easier to use typed arrays more safely in Servo. We can pass around RootedTraceableBox values into functions that return heap-allocated objects that need to store the inner `Heap` instead of the RootedTraceableBox.

Testing: Added unit test.
Servo PR: https://github.com/servo/servo/pull/44464

### DIFF
--- a/mozjs/src/gc/collections.rs
+++ b/mozjs/src/gc/collections.rs
@@ -6,6 +6,7 @@ use mozjs_sys::jsgc::GCMethods;
 use mozjs_sys::jsval::JSVal;
 use mozjs_sys::trace::Traceable;
 use std::ops::{Deref, DerefMut};
+use std::ptr;
 
 /// A vector of items to be rooted with `RootedVec`.
 /// Guaranteed to be empty when not rooted.
@@ -131,6 +132,18 @@ impl<T: Traceable + 'static> RootedTraceableBox<T> {
     pub unsafe fn ptr(&self) -> *mut T {
         self.ptr
     }
+
+    pub fn into_box(mut self) -> Box<T> {
+        self.cleanup()
+    }
+
+    fn cleanup(&mut self) -> Box<T> {
+        unsafe {
+            let ptr = std::mem::replace(&mut self.ptr, ptr::null_mut());
+            RootedTraceableSet::remove(ptr);
+            Box::from_raw(ptr)
+        }
+    }
 }
 
 impl<T> RootedTraceableBox<Heap<T>>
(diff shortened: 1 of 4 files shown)


jdm:
b997fd3aaa2b213c497abf479f7d4e74f34b3ec9 is part of this because it's also needed for https://github.com/servo/servo/pull/44464 for identical reasons as RootedTraceableBox::into_box.

mozjs/src/gc/collections.rs
```diff
@@ -131,6 +132,18 @@ impl<T: Traceable + 'static> RootedTraceableBox<T> {
     pub unsafe fn ptr(&self) -> *mut T {
         self.ptr
     }
+
+    pub fn into_box(mut self) -> Box<T> {
+        self.cleanup()
+    }
```
sagudev, line 141: We should write doc comment here that T is not rooted anymore.

I think we currently do not mark such things as unsafe everytime, but eventually we should (except if we pass &NoGc and bound it's lifetime to return type).

mozjs/src/gc/collections.rs
```diff
@@ -131,6 +132,18 @@ impl<T: Traceable + 'static> RootedTraceableBox<T> {
...
+
+    fn cleanup(&mut self) -> Box<T> {
+        unsafe {
+            let ptr = std::mem::replace(&mut self.ptr, ptr::null_mut());
```
sagudev:
```suggestion
            let ptr = std::mem::take(&mut self.ptr);
```

mozjs/src/gc/root.rs
```diff
@@ -167,6 +167,11 @@ where
         // Safety: Values within this rooted vector are traced.
         unsafe { MutableHandle::from_marked_location(self.as_mut().as_mut_ptr().add(index)) }
     }
+
+    pub fn take(&'_ mut self) -> Vec<T> {
+        // Safety: No GC can run during this call.
+        std::mem::take(unsafe { self.as_mut() })
+    }
```
sagudev, line 174: This is ok because in it's place we left empty vector not null value which would cause problem in tracing.

I wonder if it would make sense to have like into variant that would consumed the root (at least based on usage in https://github.com/servo/servo/pull/44464/changes#diff-aa673782e2e3f68b9c853c6294633094e49c64c6baed9bb37fbc66b89d9979aaR1250 it would)

sagudev (approved):
LGTM

mozjs/src/gc/root.rs
jdm: I think I can imagine how to write that. I'll try adding that next time I run into a situation like this one.

mozjs/src/gc/collections.rs
jdm: Turns out this does not compile.

sagudev: This is weird, because while reviewing this PR I wrote something like this in playground (don't mind me reusing some weird types from some previous experiment): https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c1667e8a367d2ec8d0d0973468de6e3d

PR #741, opened 2026-04-22
security bump: Fix script, and verify version

jschwe:
This PR attempts to fix the issue identified in https://github.com/servo/mozjs/pull/729#discussion_r3050185183.
We add a function to verify that the milestone version matches our expected version, so that failures are visible.
To fix `get_previous_pushes` we also check ancestors on `hg.mozilla.org` and use that to filter the time-based pushes from treeherder.
I'm pretty confident in the "verification" part of the PR, less so for the treeherder part, since I'm not familiar.

Also change the version to `<ESR>.<minor>.<patch>-0`, since otherwise there is no room for the patch version. 
For sorting use a tuple of minor, patch. In practice `float` should be fine, but in theory if we actually had 10 patch releases, then the sorting would be wrong. Changed it anyway, to be on the correct side.

The sm-security-bump script already pushed 140.10.0 to servo releases, so it's required to comment out the unneeded parts for testing.

Testing: Not part of any automated testing.

### DIFF
--- a/mozjs-sys/etc/get_taskcluster_mozjs.py
+++ b/mozjs-sys/etc/get_taskcluster_mozjs.py
@@ -2,12 +2,15 @@
 
 import sys
 import os
+import tarfile
+
 import requests
+from urllib.parse import quote
 from urllib.request import urlretrieve
 
 ESR = 140
 REPO = f"mozilla-esr{ESR}"
-HEADERS = {"User-Agent": "Mozilla/5.0 (X11; Linux x86_64) mozjs-sys/1.0"}
+HEADERS = {"User-Agent": "mozjs-sys/1.0 (https://github.com/servo/mozjs)"}
 
 
 def download_artifact(task_id: str, artifact_name: str, dl_name: str, i=0):
@@ -55,7 +58,21 @@ def find_sm_pkg_and_hazard_task_in_push(push_id: int) -> tuple[str | None, str |
     return sm_pkg_task_id, hazard_task_id
 
 
+def get_ancestor_revisions(commit: str, limit: int) -> set[str]:
+    revset = quote(f"ancestors({commit})")
+    response = requests.get(
+        f"https://hg.mozilla.org/releases/{REPO}/json-log"
+        f"?rev={revset}&limit={limit}",
+        headers=HEADERS,
+    )
+    response.raise_for_status()
+    return {entry["node"] for entry in response.json().get("entries", [])}
+
+
 def get_previous_pushes(commit: str, count: int):
+    # Multiple changesets can map to the same push, so we fetch more.
+    ancestors = get_ancestor_revisions(commit, limit=count * 20)
+
     response = requests.get(
         f"https://treeherder.mozilla.org/api/project/{REPO}/push/?revision={commit}",
         headers=HEADERS,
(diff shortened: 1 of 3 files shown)


mozjs-sys/etc/get_taskcluster_mozjs.py
```diff
@@ -2,12 +2,15 @@
...
+import tarfile
+
 import requests
+from urllib.parse import quote
 from urllib.request import urlretrieve
 
 ESR = 140
 REPO = f"mozilla-esr{ESR}"
-HEADERS = {"User-Agent": "Mozilla/5.0 (X11; Linux x86_64) mozjs-sys/1.0"}
+HEADERS = {"User-Agent": "mozjs-sys/1.0 (https://github.com/servo/mozjs)"}
```
jschwe, line 13: This was required to access `hg.mozilla.org`, apparently it doesn't like us using `Mozilla` prefix.

sagudev:
Did you test this on 140.9.X or just for latest 140.10? AFAIK 140.10 should not have any problems with old code. The only problem were patch releases for SM which uses different branches (like 140.9.1).

jschwe:
Ah, no unfortunately latest was already 140.10 when I tested. I'll need to have another look and see if i can test 140.9.1 specifically.

jschwe:
The downloading and verification part seems to also work for `140.10.1`. Any recommendations on what else to test?

sagudev:
> The downloading and verification part seems to also work for `140.10.1`. Any recommendations on what else to test?

140.10.1 also worked in current script (I think they did not do new branch). I would temporary ignore 140.10.* and check that 140.9.1 works correctly.

jschwe:
Okay, I filtered out the `minor == "10"` and it seems it worked. Note I still commented out the release checking and creating part.

Output: 
```
python3 mozjs-sys/etc/sm-security-bump.py
Latest tag: FIREFOX_140_9_1esr_RELEASE, changeset: cca515379f569261e903f6e84dcdb0ecdbbe821e
Checking initial push 1876727 (cca515379f56)...
No SpiderMonkey tasks found, checking previous push 1876216 (e0769af456c6)...
No SpiderMonkey tasks found, checking previous push 1876094 (7e4d14775e84)...
Found tasks in push 1876094 (7e4d14775e84)
Spidermonkey package task id LIv19IZvTqmZGoihP8V_9Q
Downloading: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/LIv19IZvTqmZGoihP8V_9Q/runs/0/artifacts/public/build/mozjs-140.9.1.tar.xz
Hazard task id YK8XWZE-SISTkajM3GuFHg
Downloading: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/YK8XWZE-SISTkajM3GuFHg/runs/0/artifacts/public/build/allFunctions.txt.gz
Downloading: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/YK8XWZE-SISTkajM3GuFHg/runs/0/artifacts/public/build/gcFunctions.txt.gz
Milestone verified: tarball is SpiderMonkey 140.9.1
```

mozjs-sys/etc/get_taskcluster_mozjs.py
sagudev: Interesting, I am pretty sure I had problems without using browser like UA.

mozjs-sys/etc/get_taskcluster_mozjs.py
```diff
@@ -64,13 +81,56 @@ def get_previous_pushes(commit: str, count: int):
...
+        f"https://treeherder.mozilla.org/api/project/{REPO}/push/"
+        f"?push_timestamp__lte={push_timestamp}&count={count * 4}",
         headers=HEADERS,
     )
     response.raise_for_status()
-    return response.json()["results"]
+    all_pushes = response.json()["results"]
+
+    # Keep only pushes whose tip changeset is an ancestor of `commit`.
+    on_branch = [p for p in all_pushes if p["revision"] in ancestors]
```
sagudev, line 93: I though of getting pushes of ancestors, but that would require many request, this way we do everything in just one big request. While this could potentially not work I think it will be OK in practice.

sagudev (approved):
Let's do it!

PR #744, opened 2026-04-29
Security bump SpiderMonkey to 140.10

sagudev:
Bump SpiderMonkey to 140.10 (changeset: [a24e9e115e02794f72dab9ef6081244403b0183a](https://github.com/servo/mozjs/releases/download/mozjs-source-a24e9e115e02794f72dab9ef6081244403b0183a))


mozjs-sys/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "mozjs_sys"
 description = "System crate for the Mozilla SpiderMonkey JavaScript engine."
 repository.workspace = true
-version = "0.140.10-2"
+version = "0.140.10-0"
```
jschwe, line 5: In general how should we deal with such point version updates? Would it perhaps make sense to just go to `140.10.1-0` and drop the 0 major? All releases would still be "pre-release" from cargos standpoint, so we aren't really changing anything on the semver side.

jdm: That sounds pretty reasonable to me.

sagudev: > In general how should we deal with such point version updates? Would it perhaps make sense to just go to `140.10.1-0` and drop the 0 major? All releases would still be "pre-release" from cargos standpoint, so we aren't really changing anything on the semver side.

Hm, I guess you are right.

jschwe:
Manually merged via https://github.com/servo/mozjs/pull/745

PR #748, opened 2026-05-21
Implement JSContext::get_from_thread

sagudev:
This is unofficially perma cx. Sometimes it is needed: https://github.com/servo/servo/pull/45040#discussion_r3278839426 and https://doc.servo.org/mozjs/rust/struct.Runtime.html#method.get is planned to be removed.

### DIFF
--- a/mozjs/src/context.rs
+++ b/mozjs/src/context.rs
@@ -101,7 +101,7 @@ impl JSContext {
     /// SAFETY:
     /// - only one [JSContext] can be alive and it should not outlive [Runtime].
     pub unsafe fn get_from_thread() -> Option<JSContext> {
-        unsafe { crate::rust::Runtime::get().map(JSContext::from_ptr) }
+        unsafe { crate::rust::Runtime::get().map(|raw_cx| unsafe { JSContext::from_ptr(raw_cx) }) }
     }
 
     /// Returns [NoGC] token bounded to this [JSContext].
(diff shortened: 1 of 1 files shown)


Gae24:
This is needed in particular for `DOMString`, there are three calls to `Runtime::get` that are probably tricky to remove, e.g one is called from Deref trait.

sagudev:
> This is needed in particular for `DOMString`, there are three calls to `Runtime::get` that are probably tricky to remove, e.g one is called from Deref trait.

This should be used carefully. I think usage in deref will be ok as we only need &JSContext, but other places might not be so lucky.

mozjs/src/context.rs
```diff
@@ -96,6 +96,14 @@ impl JSContext {
         JSContext { ptr: cx }
     }
 
+    /// Get the `JSContext` for this thread (thin air). This should be rarely used.
+    ///
+    /// SAFETY:
+    /// - only one [JSContext] can be alive and it should not outlive [Runtime].
+    pub unsafe fn get_from_thread() -> Option<JSContext> {
```
jdm, line 103: Marking a static function with no arguments as unsafe is interesting, because it makes me wonder what kind of safety comments we can write at the call site. Doesn't Runtime::get already uphold the safety comment for this implementation?

sagudev: `Runtime::get` returns raw context (ptr) so one does not need to uphold any invarinats really (but it should still not outlive runtime). Here we return safe jscontext so one needs to ensure that we uphold it's invariants (same as at it's construction).

jdm (approved):

Issue #747, opened 2026-05-19
Ask mozilla to make SM stuff on release


sagudev:
Look at this commit: https://hg-edge.mozilla.org/releases/mozilla-esr140/rev/b295e796caa7f5476197d8412a73e192ecfc03a1, it has SM changes but it is DONTBUILD.

And thus hopping through many pushes we could not find any artifacts for last release: https://github.com/servo/mozjs/actions/runs/26106826562/job/76772893012#step:6:33

So https://github.com/servo/mozjs/pull/727 and thus https://github.com/servo/mozjs/pull/741 are not good.

sagudev:
Hazard task is on all release commits, but sm pkg is not. In worst case we could also run pkg script on our own.

jschwe:
Did you reach out yet?

sagudev:
I yelled to the void on mozilla matrix (both on releases-discussions and spidermonkey channels), but received no response. Maybe others have better contacts within mozilla.

jschwe:
I'm looking into using https://github.com/mozilla-firefox/firefox/blob/main/js/src/make-source-package.py to create the tarball ourselves from source. Seems like that might be simpler / more reliable than downloading from upstream CI.

jschwe:
Can we close this issue, since we now build the tar-ball ourselves, or do you think this is still something worth pursuing?

sagudev:
This is still something worth pursuing not just for us but for other embedders too. But it's not urgent anymore so we do not need to beg 🙃

PR #752, opened 2026-06-07
scripts: Build mozjs.tar.xz from source instead of searching taskcluster

jschwe:
ESR 140.11 is not available on the taskcluster, since CI was skipped. The simplest solutions seems to be to just build the archive ourselves, from source, by checking out the firefox source and running the script ourselves. This allows us to simplify the taskcluster script, which now only needs to download the hazard artifacts, which apparently are always there.

This change got a lot bigger than I initially hoped - a large part is reducing the complexity again in the hazard task download script, by removing the part checking multiple pushes. Instead, we do check multiple runs of one push, since the first one doesn't always have the hazard tasks (probably since CI can fail). Maybe reverse order first would have already been sufficient, but at that point i had already written the code and it's probably safer to iterate through all runs anyway until we find the hazard tasks.

We can let CI create the release after this PR is merged.

Testing: Manual testing only - `get_latest_mozjs.py` and `sm-security-bump.py` tested locally (the latter by commenting out the parts with permanent side-effects, like creating a github release)
Addresses #747 (in a way).

### DIFF
--- a/mozjs-sys/etc/get_git_mozjs.py
+++ b/mozjs-sys/etc/get_git_mozjs.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+"""
+Build mozjs.tar.xz locally from upstream.
+Since not all upstream ESR releases build mozjs.tar.xz in CI,
+this script checks out the firefox source code from upstream at the relevant tag
+and builds the archive using the upstream script.
+"""
+
+import argparse
+import os
+import shutil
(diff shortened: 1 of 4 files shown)


jdm (approved):
Very nice!

jschwe:
MQ failed, one job had a network error and the other one had an OOM. no point on waiting for the remaining mac jobs.

PR #756, opened 2026-06-12
Fix mozjs_sys prebuilt library

jschwe:
The windows-latest image was upgraded to VS 2026 in June, which seems to cause linking errors when attempting to link the VS 2026 prebuilt library into servo (built in CI with VS 2022). Pin the old image instead, to stay in sync with servo.

Testing: This will need a prebuilt library first, before we can test this in servo, since build from source doesn't reproduce the problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,7 +128,10 @@ jobs:
       matrix:
         features: ["debugmozjs", ""]
         platform:
-          - { target: x86_64-pc-windows-msvc, os: windows-latest }
+          # Note: the image should be kept in sync with the servo image,
+          # otherwise there might be ABI issues with the prebuilt mozjs
+          # if the VS major changes.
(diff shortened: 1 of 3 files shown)


sagudev (approved):
I guess there is no backwards compatibility, but is there at least forward compatibility or does that one also breaks? If so we should do some MSVC detection in build script and fallback to build when MSVC from prebuilds does not match.

jschwe:
I'm not sure, but I would hope so. Can we test that after merging this PR, by doing a try run in servo (fork) with windows-latest? Not sure how to test it easily before the release.

jschwe:
https://learn.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-180

Seems that as long as the linker is newer than the newest used toolchain, artifacts can be combined, so we should be good.

PR #760, opened 2026-06-17
CI: Validate `-lts` version suffix on the LTS branch

jschwe:
To keep the servo LTS version using a recent SM ESR version, without needing to backport `mozjs` API changes, we add a new branch `servo-lts-v0.1`, for use in the servo v0.1 LTS release based on mozjs 0.15.4. 
To allow duplicate releases (so that publishing to crates.io. / github releases to work), we require an `-lts` suffix to mozjs_sys versions. For `mozjs` itself this isn't necessary, since we already upgrade the `minor` version, so we can continue bumping x in `0.15.X`.
The `build.rs` doesn't require any changes, since it consumes the `version` via `CARGO_PKG_VERSION` and hence will download the lts branch prebuilt release.

ESR updates on this branch would be run by editing `COMMIT` and running `update.py`, since `main` should have received the ESR update first.

Testing: This does not change the mozjs API, and should be invisible for servo (LTS), which just consumes mozjs 0.15.X releases.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,11 +28,37 @@ jobs:
       rust_version: 'beta'
     secrets: inherit
 
+  # The goal is to have an LTS branch of mozjs, where the rust API (mozjs and mozjs_sys glue code)
+  # is stable, and only spidermonkey is updated.
+  # We still keep the regular versioning, but add an `-lts` suffix, so we can have "duplicate" releases
+  # of the same ESR (to crates.io and github releases).
+  verify_lts_version:
+    runs-on: ubuntu-22.04
+    if: github.event_name == 'pull_request' && startsWith(github.base_ref, 'servo-lts-v')
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        # Check if we added the `-lts` suffix to the version, so we can create an independent crates.io release for
+        # the LTS.
+      - name: Verify LTS version suffix
+        run: |
+          MOZJS_SYS_VERSION=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "mozjs_sys") | .version')
+          # A new LTS servo version is expected to have an updated SM ESR, hence the major version would differ,
+          # and we don't need to version the lts suffix itself to achieve distinct tags for the period of time
+          # when the old LTS branch overlaps with the new one.
+          if [[ "${MOZJS_SYS_VERSION}" == *-lts ]]; then
+            echo "Version ${MOZJS_SYS_VERSION} is okay and has the expected suffix"
+          else
+            echo "You forgot to add the `-lts` suffix to the mozjs_sys version number. This is required on the lts branch"
+            exit 1
+          fi
+
 
   build_result:
     name: Result
     runs-on: ubuntu-latest
-    needs: ["build"]
+    needs: ["build", "verify_lts_version"]
     if: ${{ always() }}
     steps:
       - name: Mark the job as successful
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,10 +4,9 @@ name: Publish
 
 # Workflow triggers should be strictly controlled,
 # as this will trigger publishing to crates.io.
-# We only want to publish on the main branch.
 on:
   push:
-    branches: [main]
+    branches: ["main", "servo-lts-v*"]
 
 # dispatches build workflow with different permissions
 jobs:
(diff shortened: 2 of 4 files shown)


sagudev:
> ESR updates on this branch would be run by editing COMMIT and running update.py, since main should have received the ESR update first.

Not sure about that, given that we can get new ESR on main while LTS could still remain on old one for few months more.

.github/workflows/main.yml
```diff
@@ -28,11 +30,37 @@ jobs:
       rust_version: 'beta'
     secrets: inherit
 
+  # This job is only added to the LTS branch.
+  # It should be cherrypicked to the next LTS branch, once we make a new LTS servo major.
+  # The goal is to have an LTS branch of mozjs, where the rust API (mozjs and mozjs_sys glue code)
+  # is stable, and only spidermonkey is updated.
+  # We still keep the regular versioning, but add an `-lts` suffix, so we can have "duplicate" releases
+  # of the same ESR (to crates.io and github releases).
```
sagudev: Usually stuff also goes on main but in disabled state (for next LTS).

sagudev (approved):

sagudev:
I still need to sort out branch setting but that is non trivial thing apparently.

jschwe:
> > ESR updates on this branch would be run by editing COMMIT and running update.py, since main should have received the ESR update first.
> 
> Not sure about that, given that we can get new ESR on main while LTS could still remain on old one for few months more.

True, although all we really need is for the github release with the source tarball to have been created, which we could do by manually running the script (and backporting the latest sm-security-bump.py changes if necessary).

.github/workflows/main.yml
jschwe: You mean, add this, but comment the job on main?

jschwe:
Are you okay with landing this as is (on the new branch)? I guess I could also ask @jdm to protect the branch, he's done that recently in servo too.

.github/workflows/main.yml
sagudev: We could just detect if branch name contains "LTS" and do additional stuff.

sagudev:
> > > ESR updates on this branch would be run by editing COMMIT and running update.py, since main should have received the ESR update first.
> > 
> > 
> > Not sure about that, given that we can get new ESR on main while LTS could still remain on old one for few months more.
> 
> True, although all we really need is for the github release with the source tarball to have been created, which we could do by manually running the script (and backporting the latest sm-security-bump.py changes if necessary).

I suspect that we should be running scheduled run of SM bump independently on LTS branches.

sagudev:
> Are you okay with landing this as is (on the new branch)? I guess I could also ask @jdm to protect the branch, he's done that recently in servo too.

Yes I am okay with landing. I also looked into the protection of the branch and I think the problem is that no CI runs happened so some options are unavailable. I triggered a CI run there and hopefully this will allow all options we need.

jschwe:
> I suspect that we should be running scheduled run of SM bump independently on LTS branches.

[Scheduled runs only run on the default branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows), so that would complicate things a bit. Although I guess we could adapt the workflow on main to also explicitly target the LTS branch. But I think I'd prefer to defer that work into next month, I guess it should be sufficient if its ready before / around the time we do the ESR upgrade (which is in ~4 weeks)

.github/workflows/main.yml
jschwe: Thats an interesting idea. I'll give that a shot

sagudev:
> > I suspect that we should be running scheduled run of SM bump independently on LTS branches.
> 
> [Scheduled runs only run on the default branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows), so that would complicate things a bit. Although I guess we could adapt the workflow on main to also explicitly target the LTS branch. But I think I'd prefer to defer that work into next month, I guess it should be sufficient if its ready before / around the time we do the ESR upgrade (which is in ~4 weeks)

I see, but we can manually dispatch the workflow from the LTS (tho this currently does not handle LTS branches correctly).

sagudev:
> I also looked into the protection of the branch and I think the problem is that no CI runs happened so some options are unavailable. I triggered a CI run there and hopefully this will allow all options we need.

Re this: I've enabled branch protection like on [servo-lts-v0.1](https://github.com/servo/mozjs/tree/servo-lts-v0.1), but it will not work on the rest of LTS branches as we cannot enable MQ in rules that apply many branches (have wildcard pattern). One possible solution for this is to not have MQ there (we do not expect many PR there anyway) and select rule that all PRs needs to be rebased and check passing (thus effectively enforcing benefits of MQ without actual MQ).

jschwe:
> One possible solution for this is to not have MQ there (we do not expect many PR there anyway) and select rule that all PRs needs to be rebased and check passing (thus effectively enforcing benefits of MQ without actual MQ).

I'd be fine with that, we do the same for servo/servo LTS branches.

PR #765, opened 2026-06-26
CI: Validate `-lts` version suffix on the LTS branch

jschwe:
This applies the CI verification needed for the LTS branch to main (#760), so we don't forget it when doing the ESR upgrade.
Same as on main, we expect to always merge to the LTS branch via pull-requests, so validating on pull-requests should be sufficient.

Testing: This is a CI change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,11 +28,39 @@ jobs:
       rust_version: 'beta'
     secrets: inherit
 
+  # The goal is to have an LTS branch of mozjs, where the rust API (mozjs and mozjs_sys glue code)
+  # is stable, and only spidermonkey is updated.
+  # We still keep the regular versioning, but add an `-lts` suffix, so we can have "duplicate" releases
+  # of the same ESR (to crates.io and github releases).
+  # When targetting `main` (or any other non lts branch) this job does nothing.
+  verify_lts_version:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        if: github.event_name == 'pull_request' && startsWith(github.base_ref, 'servo-lts-v')
+        # Check if we added the `-lts` suffix to the version, so we can create an independent crates.io release for
+        # the LTS.
+      - name: Verify LTS version suffix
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -28,11 +28,39 @@ jobs:
...
+  # We still keep the regular versioning, but add an `-lts` suffix, so we can have "duplicate" releases
+  # of the same ESR (to crates.io and github releases).
+  # When targetting `main` (or any other non lts branch) this job does nothing.
+  verify_lts_version:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        if: github.event_name == 'pull_request' && startsWith(github.base_ref, 'servo-lts-v')
```
sagudev: Can move the if guard one level up to the job (not step) level. It should work per https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions for more info.

.github/workflows/main.yml
```diff
@@ -28,11 +28,37 @@ jobs:
       rust_version: 'beta'
     secrets: inherit
 
+  # The goal is to have an LTS branch of mozjs, where the rust API (mozjs and mozjs_sys glue code)
+  # is stable, and only spidermonkey is updated.
+  # We still keep the regular versioning, but add an `-lts` suffix, so we can have "duplicate" releases
+  # of the same ESR (to crates.io and github releases).
+  verify_lts_version:
+    runs-on: ubuntu-22.04
+    if: github.event_name == 'pull_request' && startsWith(github.base_ref, 'servo-lts-v')
```
sagudev, line 37: Any reason why we limit this to PRs?

sagudev (approved):

.github/workflows/main.yml
jschwe: Otherwise the condition would get more complicated. For PRs, we are interested in the target branch (hence base_ref). We don't have a push trigger for this workflow file. And for manual workflow triggers I'm not sure if we would have a branch name.

sagudev: Ah indeed.

PR #767, opened 2026-06-29
Allow configuring SM optimization level for debugmozjs builds

### LINKED ISSUE #733: Should debug-mozjs feature force `O0`
`--feature debug-mozjs` enables debug assertions, but also forces the optimization level to 0. 
IMHO these two are orthogonal, and we shouldn't need to force O0.

jschwe:
This PR allows configuring the optimisation level of spidermonkey by honoring `OPT_LEVEL` (which is set by cargo).
For the prebuilt CI artifact, we default to `O3` (i.e. the default for the `--release` profile), which allows us to have a somewhat performant SM version with debug assertions, which hopefully would allow us to run `wpt` tests with this debug-mozjs version (once existing assertion failures in wpt are fixed).
This does mean that when debugging issues in SM, one might need to build mozjs from source more often, but since the fallback to from source build is automatic, this shouldn't be too bad. 
We could consider adding a `O0` debugmozjs build for Linux + arm mac, since those are probably the only two platforms actively used by developers working on mozjs, but that could be done in a follow-up if necessary.

Testing: We don't test optimisation levels
Servo PR: TODO
Closes: #733

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,8 +63,8 @@ jobs:
         env:
           DEBUGMOZJS_OPT: ${{ matrix.features == 'debugmozjs' && '--config profile.dev.package.mozjs_sys.opt-level=3' || '' }}
         run: |
-          cargo +${{ steps.toolchain.outputs.name }} build --verbose --features "${{ matrix.features }}" ${DEBUGMOZJS_OPT}
-          cargo +${{ steps.toolchain.outputs.name }} test --tests --examples --verbose --features "${{ matrix.features }}" ${DEBUGMOZJS_OPT}
+          cargo +${{ steps.toolchain.outputs.name }} build --verbose --features "${{ matrix.features }}" ${{ env.DEBUGMOZJS_OPT }}
+          cargo +${{ steps.toolchain.outputs.name }} test --tests --examples --verbose --features "${{ matrix.features }}" ${{ env.DEBUGMOZJS_OPT }}
       - name: Upload artifact
         if: ${{ env.NEW_RUST_CHECK == 'false' }}
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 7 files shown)


sagudev:
> This does mean that when debugging issues in SM, one might need to build mozjs from source more often, but since the fallback to from source build is automatic, this shouldn't be too bad.

Actually this is what worries me, sometimes I get fallback from source even with no changes (in cases where I am offline). But given that we want O3 in CI I only see two options:
- use `OPT_LEVEL` only when explicitly set `MOZJS_FROM_SOURCE=1` and use `--release` in CI
- expose new env var `SM_OPT_LEVEL`, if not provided we default to O3; if overriden to something other then O3 it causes rebuild from source

I am starting to prefer second option.

jschwe:
> But given that we want O3 in CI I only see two options:

There is also the third option, that we additionally prebuild `O0` versions for x86 linux and arm64 mac, since those 2 should probably cover the needs of the day-to-day usage of the people likely to be working on mozjs with `debugmozjs` enabled. (If the feature is disabled, we don't respect `OPT_LEVEL` anyway). That would be a compromise between additional CI time, and targeting most needs imho.

sagudev:
> we additionally prebuild O0 versions for x86 linux and arm64 mac

We could also make this work under `SM_OPT_LEVEL`.

jschwe:
What makes me a bit concerned about `SM_OPT_LEVEL` is that it breaks the assumption that a simple `./mach build --dev --debugmozjs` will give you a build you can use for gdb / lldb debugging and is yet another thing one needs to remember, otherwise one gets "wierd" behavior in the debugger.

.github/workflows/build.yml
```diff
@@ -57,15 +57,20 @@ jobs:
...
       - name: Build
+        # Build the debug-mozjs prebuilt artifact with -O3 (matching the
+        # `checked-release` profile in servo) so that we have a fast prebuilt version with
+        # debug assertions for running WPT in servo with debug-mozjs.
+        env:
+          DEBUGMOZJS_OPT: ${{ matrix.features == 'debugmozjs' && '--config profile.dev.package.mozjs_sys.opt-level=3' || '' }}
         run: |
-          cargo +${{ steps.toolchain.outputs.name }} build --verbose --features "${{ matrix.features }}"
-          cargo +${{ steps.toolchain.outputs.name }} test --tests --examples --verbose --features "${{ matrix.features }}"
+          cargo +${{ steps.toolchain.outputs.name }} build --verbose --features "${{ matrix.features }}" ${DEBUGMOZJS_OPT}
```
sagudev: Instead of using shell replacement it's better to use yaml replacement:
```suggestion
          cargo +${{ steps.toolchain.outputs.name }} build --verbose --features "${{ matrix.features }}" ${{ env.DEBUGMOZJS_OPT }}
```

sagudev (approved):
That is actually perfect, I overlooked the dev/release profile part.

jschwe:
I edited the description in the readme a bit, so hopefully it's a bit clearer now.

PR #773, opened 2026-07-06
Add safer cell types

sagudev:
`JSRefCell` is based on work done in https://github.com/servo/servo/pull/46050 and `JSCell` is based on https://github.com/servo/mozjs/issues/569#issuecomment-2764639269, but we do not offer special handling for Copy types as they should be stored in normal Cell instead and instead of requiring &mut JSContext we require &mut NoGC when we need exclusivity.

Testing: Many doctests and also some regular unit tests.
Servo PR: https://github.com/servo/servo/pull/46447

### DIFF
--- a/mozjs/src/cell.rs
+++ b/mozjs/src/cell.rs
@@ -30,6 +30,16 @@ impl<T> JSRefCell<T> {
         }
     }
 
+    /// Returns a reference to the underlying [`RefCell`].
+    ///
+    /// This is not strictly unsafe, but it is not recommended to use this method unless you know what you are doing.
+    ///
+    /// Mutable borrows from it can cause panics if a GC happens while the borrow is active,
+    /// which is why [`JSRefCell::borrow_mut`] should be used instead.
+    pub fn as_refcell(&self) -> &RefCell<T> {
+        &self.value
+    }
+
     /// Immutably borrows the wrapped value.
     ///
     /// The borrow lasts until the returned `Ref` exits scope. Multiple
--- a/mozjs/src/cell.rs
+++ b/mozjs/src/cell.rs
@@ -1,0 +1,168 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+//! A shareable mutable container.
+
+pub use std::cell::{Ref, RefMut};
+use std::cell::{RefCell, UnsafeCell};
+
+use mozjs_sys::trace::Traceable;
+
+use crate::context::{JSContext, NoGC};
+use crate::jsapi::JSTracer;
+
+/// [`std::cell::RefCell`] that statically prevents borrow_mut panics
+/// from occurring by borrowing due to GC using the [`NoGC`] marker.
+///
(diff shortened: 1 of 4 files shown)


sagudev:
idea: `&mut NoGC` as affinity guard (akin to exclusivesnatchguard in wgpu)

mozjs/src/cell.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// ```
+    /// use mozjs::context::{JSContext, NoGC};
+    /// use mozjs::cell::JSRefCell;
+    /// fn f(no_gc: &NoGC, cell: &JSRefCell<String>) {
+    ///     let mut mutably_borrowed = cell.borrow_mut(no_gc);
+    /// }
+    /// ```
+    ///
+    /// But in more complex cases, method might trigger a GC, and thus require a [`&mut JSContext`][crate::context::JSContext].
```
jdm:
```suggestion
    /// But in more complex cases, a method might trigger a GC, and thus require a [`&mut JSContext`][crate::context::JSContext].
```

mozjs/src/cell.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+}
+
+impl<T: Default> JSRefCell<T> {
+    /// Takes the wrapped value, leaving `Default::default()` in its place.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value is currently borrowed.
+    pub fn take(&self, _no_gc: &NoGC) -> T {
```
jdm: Why do we need a NoGC argument here?

mozjs/src/cell.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+unsafe impl<T: Traceable> Traceable for JSRefCell<T> {
+    unsafe fn trace(&self, trc: *mut JSTracer) {
+        unsafe { (*self).borrow().trace(trc) };
+    }
+}
+
+/// A cell for interior mutability, that (ab)uses [NoGC]
+/// as a borrow token for ensuring safety.
```
jdm, line 148: Huh. So this is effectively working around how we can't have `&mut self`, but we use `&mut NoGC` for the same purpose and perform unsafe operations under the hood?

jdm (approved):
It's subtle, but very interesting!

mozjs/src/cell.rs
sagudev: Yes!

mozjs/src/cell.rs
sagudev: Ah, indeed we don't need it. I got confused with RefCell vs. Cell (because `RefCell::take` internally uses `RefCell:borrow_mut`).

mozjs/src/cell.rs
sagudev: Fun fact: I've been thinking to allow `&mut self` on dom methods (and forbid them from calling GC so they got no cx), but that might be too restrictive as here we can have more fine grained no_gc section. That might be still worth perusing in the future (but it would need some kind of interior mut on the self). That would work well on webgpu because the only mutators are `SetLabel`.

Problem: We cannot pass `&mut NoGC` and `&JSContext` at the same time, but I think it should be ok to go from `&NoGC` to `&JSContext`

PR #776, opened 2026-07-09
Mark VecDeque as rootable.

jdm:
This allows using VecDeque values with the root! macro.


sagudev (approved):

sagudev:
Needs bump in https://github.com/servo/mozjs/blob/fb1847a2db4c2d32ca1dcd0f05dd3bb4694e0fd6/mozjs/Cargo.toml#L30 too which also causes CI to require bump for mozjs_sys IIRC.

PR #777, opened 2026-07-09
Expose safe typed array slices.

jdm:
We can remove a bunch of unsafe usage related to typed arrays since we can now tie the slice lifetime to a NoGC argument.

Servo PR: https://github.com/servo/servo/pull/46416

### DIFF
--- a/mozjs/src/typedarray.rs
+++ b/mozjs/src/typedarray.rs
@@ -9,6 +9,7 @@
 use crate::conversions::ConversionResult;
 use crate::conversions::FromJSValConvertible;
 use crate::conversions::ToJSValConvertible;
+use crate::context::NoGC;
 use crate::glue::GetFloat32ArrayLengthAndData;
 use crate::glue::GetFloat64ArrayLengthAndData;
 use crate::glue::GetInt16ArrayLengthAndData;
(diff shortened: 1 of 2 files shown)


mozjs/src/typedarray.rs
```diff
@@ -207,6 +208,14 @@ impl<T: TypedArrayElement, S: JSObjectStorage> TypedArray<T, S> {
         &*self.data()
     }
 
+    pub fn as_slice_safe<'a>(&self, _no_gc: &'a NoGC) -> &'a [T::Element] {
+        // SAFETY: The slice can only be invalidated by invoking JS engine
+        //         behaviour that detaches the underlying typed array.
+        //         The slice's lifetime is bounded by the provided NoGC token,
+        //         which prevents any JS engine interaction.
+        unsafe { &*self.data() }
```
sagudev: we should also assert that we do not have null ptr here, see #775

sagudev:
Oh and we should probably mark unsafe variants as deprecated.

mozjs/src/typedarray.rs
```diff
@@ -187,6 +188,7 @@ impl<T: TypedArrayElement, S: JSObjectStorage> TypedArray<T, S> {
     }
 
     /// Retrieves an owned data that's represented by the typed array.
+    #[allow(deprecated)]
```
sagudev, line 191: why?

jdm: It's a method that calls as_slice and doesn't take a cx argument, and passing in a cx argument is unnecessary since we already guarantee that the slice does not outlive the stack frame.

sagudev: I just brainfarted here, I read this is `#[deprecated]`.

sagudev (approved):

PR #778, opened 2026-07-12
More raw `JSContext` deprecations

Gae24:
To fully remove unsafe bindings we need to introduce safe variants of some existing mozjs APIs, starting with `describe_scripted_caller` and `error_info_from_exception_stack`.
Deprecate `wrappers` module too.

Testing: It compiles
Servo PR: https://github.com/servo/servo/pull/46461

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -1062,20 +1062,12 @@ pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCal
     })
 }
 
-pub fn describe_scripted_caller_safe(
-    cx: &mut crate::context::JSContext,
-) -> Result<ScriptedCaller, ()> {
+pub fn describe_scripted_caller_safe(cx: &crate::context::JSContext) -> Result<ScriptedCaller, ()> {
     let mut buf = [0; 1024];
     let mut line = 0;
     let mut col = 0;
     if unsafe {
-        !DescribeScriptedCaller(
-            cx.raw_cx(),
-            buf.as_mut_ptr(),
-            buf.len(),
-            &mut line,
-            &mut col,
-        )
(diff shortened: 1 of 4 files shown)


mozjs/src/rust.rs
```diff
@@ -1078,6 +1104,39 @@ unsafe extern "C" fn fill_string_callback(ptr: *const c_char, len: usize, target
...
+    rval: MutableHandleValue,
+) -> Option<ErrorInfo> {
+    let mut message = String::new();
+    let mut filename = String::new();
+
+    let mut line = 0;
+    let mut col = 0;
+
+    unsafe {
+        if !PendingExceptionStackInfo(
```
sagudev: glue functions are also available in wrappers2: https://doc.servo.org/mozjs/rust/wrappers2/fn.DescribeScriptedCaller.html

mozjs/src/rust.rs
```diff
@@ -1061,6 +1062,31 @@ pub unsafe fn describe_scripted_caller(cx: *mut JSContext) -> Result<ScriptedCal
...
 }
 
+pub fn describe_scripted_caller_safe(
+    cx: &mut crate::context::JSContext,
+) -> Result<ScriptedCaller, ()> {
+    let mut buf = [0; 1024];
+    let mut line = 0;
+    let mut col = 0;
+    if unsafe {
+        !DescribeScriptedCaller(
```
sagudev: Similarly we have https://doc.servo.org/mozjs/rust/wrappers2/fn.DescribeScriptedCaller.html. Tho because this is glue function bindings will default to `&mut JSContext` but I just checked it's impl and it cannot GC, so it might make sense to add it here: https://github.com/servo/mozjs/blob/3b2ebf2dbd45dc0a8e8a858113eff74496f2fdb5/mozjs/src/generate_wrappers.py#L131-L132

sagudev (approved):

PR #780, opened 2026-07-14
Don't panic when retrieving slices from detached typed arrays

jdm:
By using `Option<NonNull<[..]>>` instead of `*mut [..]` we can ensure that we always check for detached buffers and don't need to panic.

Testing: New unit test for detached ArrayBuffer.
Servo PR: https://github.com/servo/servo/pull/46504

### DIFF
--- a/mozjs/src/typedarray.rs
+++ b/mozjs/src/typedarray.rs
@@ -126,14 +126,17 @@ pub enum CreateWith<'a, T: 'a> {
     Slice(&'a [T]),
 }
 
+#[derive(Clone, Copy)]
+enum ArrayData<T> {
+    NotYetComputed,
+    Detached,
+    Computed(NonNull<[T]>)
+}
+
 /// A typed array wrapper.
 pub struct TypedArray<T: TypedArrayElement, S: JSObjectStorage> {
     object: S,
-    /// The cached computed data for this typed array. A None value means
-    /// the data has not been computed. A Some(None) means the underlying
-    /// buffer is detached. A Some(Some(..)) is the computed, non-detached
-    /// buffer contents.
-    computed: Cell<Option<Option<NonNull<[T::Element]>>>>,
+    computed: Cell<ArrayData<T::Element>>,
 }
 
 unsafe impl<T> CustomTrace for TypedArray<T, *mut JSObject>
@@ -161,18 +164,22 @@ impl<T: TypedArrayElement, S: JSObjectStorage> TypedArray<T, S> {
 
             Ok(TypedArray {
                 object: S::from_raw(unwrapped),
-                computed: Cell::new(None),
+                computed: Cell::new(ArrayData::NotYetComputed),
             })
         }
     }
 
     fn data(&self) -> Option<NonNull<[T::Element]>> {
-        if let Some(data) = self.computed.get() {
-            return data;
+        if let ArrayData::Computed(data) = self.computed.get() {
+            return Some(data);
         }
 
         let data = unsafe { T::length_and_data(self.object.as_raw()) };
-        self.computed.set(Some(data));
+        self.computed.set(if let Some(data) = data {
+            ArrayData::Computed(data)
+        } else {
+            ArrayData::Detached
+        });
         data
     }
 
@@ -320,7 +327,7 @@ impl<T: TypedArrayElementCreator + TypedArrayElement, S: JSObjectStorage> TypedA
 /// and various functions required to manipulate typed arrays of that element type.
 pub trait TypedArrayElement {
     /// Underlying primitive representation of this element type.
-    type Element;
+    type Element: Copy;
     /// Unwrap a typed array JS reflector for this element type.
     unsafe fn unwrap_array(obj: *mut JSObject) -> *mut JSObject;
     /// Retrieve the length and data of a typed array's buffer for this element type.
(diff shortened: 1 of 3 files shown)


sagudev:
BTW Here are some notes about this from other PR: https://github.com/servo/mozjs/pull/775#discussion_r3542306704

jdm:
I believe this is addressing that note; we now always return an empty slice for detached buffers, and https://github.com/servo/servo/pull/46069 helps centralize that logic even more.

mozjs/src/typedarray.rs
```diff
@@ -127,7 +129,11 @@ pub enum CreateWith<'a, T: 'a> {
 /// A typed array wrapper.
 pub struct TypedArray<T: TypedArrayElement, S: JSObjectStorage> {
     object: S,
-    computed: Cell<Option<*mut [T::Element]>>,
+    /// The cached computed data for this typed array. A None value means
+    /// the data has not been computed. A Some(None) means the underlying
+    /// buffer is detached. A Some(Some(..)) is the computed, non-detached
+    /// buffer contents.
+    computed: Cell<Option<Option<NonNull<[T::Element]>>>>,
```
sagudev: Code instead of comments plz. Maybe something like this:
```rust
enum ArrayState {
    NotYetComputed,
    Detached,
    Computed(NonNull<[T::Element]>)
```
or maybe two different enums. Anything is better then nested Option.

jdm: Fair.

sagudev:
> I believe this is addressing that note; we now always return an empty slice for detached buffers, and [servo/servo#46069](https://github.com/servo/servo/pull/46069) helps centralize that logic even more.

My only worry is that we collapse detached/not computed into empty slice, which can hide logical problems elsewhere (like using detached buffer) which should not occur if we follow the spec (and the spec is correctly written).

jdm:
> > I believe this is addressing that note; we now always return an empty slice for detached buffers, and [servo/servo#46069](https://github.com/servo/servo/pull/46069) helps centralize that logic even more.
> 
> My only worry is that we collapse detached/not computed into empty slice, which can hide logical problems elsewhere (like using detached buffer) which should not occur if we follow the spec (and the spec is correctly written).

I'm confused, since https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy already collapses that result. Is the concern that code that _doesn't_ use that algorithm may have logic that treats detached array buffers differently, and we won't be able to implement that correctly?

sagudev:
> > > I believe this is addressing that note; we now always return an empty slice for detached buffers, and [servo/servo#46069](https://github.com/servo/servo/pull/46069) helps centralize that logic even more.
> > 
> > 
> > My only worry is that we collapse detached/not computed into empty slice, which can hide logical problems elsewhere (like using detached buffer) which should not occur if we follow the spec (and the spec is correctly written).
> 
> I'm confused, since https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy already collapses that result. Is the concern that code that _doesn't_ use that algorithm may have logic that treats detached array buffers differently, and we won't be able to implement that correctly?

Yeah, but maybe I am just paranoid. Wasn't there a website where one could find all [IsDetachedBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-isdetachedbuffer) in all specs?

jdm:
I'm trying to find out what that site is on Matrix, because I have forgotten it. However, I did come across https://w3c.github.io/IndexedDB/#convert-a-value-to-a-key

jdm:
And https://html.spec.whatwg.org/multipage/canvas.html#dom-context2d-putimagedata-common ;_;

jdm:
Perhaps it's enough to add an API to determine if an array buffer is detached, though.

sagudev:
> Perhaps it's enough to add an API to determine if an array buffer is detached, though.

but this bring us again to

> can hide logical problems elsewhere (like using detached buffer) which should not occur if we follow the spec (and the spec is correctly written).

I consider such crashes as good as one could easily see what is wrong (missing detached buffer check) and how it needs to be fixed.

jdm:
I don't like relying on a panic for this, because it only manifests if test code or a webpage triggers it. If we want to force callers to consider this case, we should change as_slice/as_slice_mut to return an enum instead. https://github.com/servo/servo/pull/46069 means that won't be a huge annoyance for callsites any more.

sagudev:
> I don't like relying on a panic for this, because it only manifests if test code or a webpage triggers it.

Fair.

> If we want to force callers to consider this case, we should change as_slice/as_slice_mut to return an enum instead. [servo/servo#46069](https://github.com/servo/servo/pull/46069) means that won't be a huge annoyance for callsites any more.

That would be ideal.

TimvdLippe:
> Yeah, but maybe I am just paranoid. Wasn't there a website where one could find all [IsDetachedBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-isdetachedbuffer) in all specs?

Yes that website is https://dontcallmedom.github.io/webdex/i.html although I don't see it listed

sagudev:
> > Yeah, but maybe I am just paranoid. Wasn't there a website where one could find all [IsDetachedBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-isdetachedbuffer) in all specs?
> 
> Yes that website is https://dontcallmedom.github.io/webdex/i.html although I don't see it listed

Thanks, bookmarked. It's here: https://dontcallmedom.github.io/webdex/i.html#IsDetachedBuffer%40%40ecmascript%25%25abstract-op although references are very vague.

sagudev (approved):

PR #782, opened 2026-07-16
Derive raw pointers instead of creating mutable references to root stacks.

### LINKED ISSUE #781: `JS::AutoGCRooter` creates a mutable reference to a shared object
Hello! I'm with the [BorrowSanitizer](https://github.com/borrowSanitizer/bsan) team. We are building an experimental tool for finding violations of Rust's Tree Borrows aliasing model in multilanguage programs. Earlier this year, @jdm reached out to us about testing `mozjs` with our tool. We have just reached the point where we can execute test programs of this magnitude (albeit quite slowly).

We found the following aliasing violation in the latest version of `mozjs-sys`, when running the test `callback`:
```
error: Undefined Behavior: read access through <20114129>(unprotected) at alloc57837[0x70] is forbidden
    --> mozjs/mozjs-sys/src/jsimpls.rs:416:17
     |
 416 | assert!(*self.stack == self as *mut _ as usize as _);
     |
     = help: the accessed tag <20114129>(unprotected) has state Disabled which forbids this child read access
help: the accessed tag <20114129>(unprotected) was created here, in the initial state Reserved
    --> mozjs/mozjs-sys/src/jsimpls.rs:423:9
     |
 423 | &mut (*rooting_cx).stackRoots_.0[kind] as *mut _ as *mut _
     |
help: the accessed tag <20114129>(unprotected) later transitioned to Disabled due to a foreign write access
 at offsets [0x70..0x78]
    --> mozjs_sys/../include/js/RootingAPI.h:1174:18
     |
1174 | *this->stack = this;
     |
help: this transition corresponds to a loss of read permissions

stack backtrace:
0: <mozjs_sys::jsgc::RootedBase>::remove_from_root_stack
      at .../mozjs/mozjs-sys/src/jsimpls.rs:416:17
1: <mozjs_sys::jsgc::Rooted<mozjs_sys::generated::root::JS::Value>>::remove_from_root_stack
      at .../mozjs/mozjs-sys/src/jsimpls.rs:449:19
2: <mozjs::gc::root::RootedGuard<mozjs_sys::generated::root::JS::Value> as core::ops::drop::Drop>::drop
      at .../mozjs/mozjs/src/gc/root.rs:198:26
```
The underlying problem seems to be this mutable reference here:
```rust
  let autoGCRooters: *mut _ = {
     let rooting_cx = cx as *mut JS::RootingContext;
     &mut (*rooting_cx).autoGCRooters_.0[self.kind_ as usize]
  };
  self.stackTop = autoGCRooters as *mut *mut _;
```
The pointer `autoGCRooters` was cast from a mutable reference, so it needs to have unique ownership. If the same memory is accessed again through a different pointer, then `self.stackTop` will become invalid. The solution seems to be changing `&mut` to `&raw mut`, so that you get the same pointer without asserting unique ownership.

icmccorm:
Fixes #781.

`JS:AutoGCRooter` and `RootedBase` have methods that create mutable references to objects that are shared across the C/C++ boundary, leading to an aliasing bug found by [BorrowSanitizer](https://github.com/borrowSanitizer/bsan). This PR switches to using `&raw mut` within the affected APIs.

### DIFF
--- a/mozjs-sys/src/jsimpls.rs
+++ b/mozjs-sys/src/jsimpls.rs
@@ -200,7 +200,7 @@ impl JS::AutoGCRooter {
         #[allow(non_snake_case)]
         let autoGCRooters: *mut _ = {
             let rooting_cx = cx as *mut JS::RootingContext;
-            &mut (*rooting_cx).autoGCRooters_.0[self.kind_ as usize]
+            &raw mut (*rooting_cx).autoGCRooters_.0[self.kind_ as usize]
         };
         self.stackTop = autoGCRooters as *mut *mut _;
         self.down = *autoGCRooters as *mut _;
@@ -420,7 +420,7 @@ impl RootedBase {
     unsafe fn get_root_stack(cx: *mut JSContext, kind: JS::RootKind) -> *mut *mut RootedBase {
         let kind = kind as usize;
         let rooting_cx = Self::get_rooting_context(cx);
-        &mut (*rooting_cx).stackRoots_.0[kind] as *mut _ as *mut _
+        &raw mut (*rooting_cx).stackRoots_.0[kind] as *mut _ as *mut _
     }
 
     unsafe fn get_rooting_context(cx: *mut JSContext) -> *mut JS::RootingContext {


sagudev (approved):
Thank you! I always suspected we should use `&raw mut`/`&raw const` in more places.

PR #789, opened 2026-08-10
build: Unify bindings and glue code flags with cc-rs

jschwe:
This PR revives #734. To reduce duplication between bindings and glue code, we use a shared cc-rs builder, and apply flags uniformly to glue code build and binding generation. The spidermonkey build itself is out-of-scope as of now.

Beyond the pure refactor, this PR also has the following changes:

- Use cc-rs to determine the cxx stdlib flags, instead of of just considering `CXXSTDLIB` and only add our fallback path if the environment doesn't define a CXXSTDLIB to use.
- bindgen did not receive a `--target` argument before this PR, but gets it now via cc-rs. This can affect binding generation when cross-compiling, but running bindgen for the target triple is definitly what we want.

Testing: Testing via companion servo PR.
Servo PR: https://github.com/servo/servo/pull/47131

### DIFF
--- a/mozjs-sys/build.rs
+++ b/mozjs-sys/build.rs
@@ -319,22 +319,11 @@ fn is_buggy_make_version() -> bool {
 }
 
 fn build(build_dir: &Path, target: BuildTarget) {
-    let mut build = cc::Build::new();
-    build.cpp(true).file(target.path());
-
-    for flag in cc_flags(false) {
-        build.flag_if_supported(flag);
-    }
+    let mut build = get_common_cc(build_dir, target);
+    build.file(target.path());
 
     if let Ok(android_api) = env::var("ANDROID_API_LEVEL").as_deref() {
         build.define("__ANDROID_MIN_SDK_VERSION__", android_api);
-    }
-
-    build.flag(include_file_flag(build.get_compiler().is_like_msvc()));
-    build.flag(&js_config_path(build_dir));
-
-    for path in target.include_paths(build_dir) {
-        build.include(path);
     }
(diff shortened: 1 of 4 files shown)


sagudev:
re windows failures, IIRC libclang uses normal clang driver and it's arguments while rest invocations uses clang-cl style

jschwe:
Hmm, I guess I will need to try this out on a real windows machine.

mozjs-sys/build.rs
```diff
@@ -515,75 +523,90 @@ fn minimum_rust_target() -> RustTarget {
...
+            .std("c++17")
+            .flag_if_supported("-Zi")
+            .flag_if_supported("-GR-")
+            .define("WIN32", None)
             // Don't use reinterpret_cast() in offsetof(),
             // since it's not a constant expression, so can't
             // be used in static_assert().
-            "-D_CRT_USE_BUILTIN_OFFSETOF",
-        ]);
+            .define("_CRT_USE_BUILTIN_OFFSETOF", None);
```
jschwe, line 543: `./src/jsapi.cpp: warning C4117: macro name '_CRT_USE_BUILTIN_OFFSETOF' is reserved, '#define' ignored`

I can see (when the build fails), presumably otherwise the warning is hidden. Maybe we should remove this while we are at it.

jschwe:
Try run: https://github.com/jschwe/servo/actions/runs/31407322992 - Looks like all jobs that support compiling mozjs from source passed.

sagudev (approved):

PR #795, opened 2026-08-12
ci: Restrict security bump to servo/mozjs.

### LINKED ISSUE #792: CI: Disable security bump workflow in forks.

jschwe:
The action shouldn't run in forks, and will fail there.

Testing: Not tested, but trivial. We already do the same checks in servo/servo for many workflows.
Fixes #792

### DIFF
--- a/.github/workflows/security-bump.yml
+++ b/.github/workflows/security-bump.yml
@@ -9,6 +9,8 @@ on:
 jobs:
     bump:
         runs-on: ubuntu-latest
+        # We don't want to run the security bump in forks.
+        if: github.repository == 'servo/mozjs'
         permissions:
             contents: write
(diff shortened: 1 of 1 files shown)


sagudev (approved):

sagudev:
Do not rebase as I want to see that "Require branches to be up to date before merging" works.

sagudev:
> Do not rebase as I want to see that "Require branches to be up to date before merging" works.

It worked (pr did not land), but it did not really give any concrete error msg.

jschwe:
> It worked (pr did not land), but it did not really give any concrete error msg.

Yeah, I think that is how it works. I think the only hint is that the button changes from "Enable automege" to merge now if the status checks are actually met already.

PR #796, opened 2026-08-14
Use cargo-ohos to build for OpenHarmony with LLVM-19

jschwe:
In preparation for the upcoming ESR-153, switch to using `cargo-ohos` instead of our custom script.
Supporting a non-SDK LLVM ends up being quite a bit more complex, and needing to duplicate that logic across every project that needs to build for ohos doesn't make sense. Hence, use cargo-ohos to wrap all the environment variables, and toolchain downloading etc. for an openharmony build.

Testing: Covered by existing tests in servo
Servo PR: https://github.com/servo/servo/pull/47421


sagudev:
BTW Rust 1.88 (so the MSRV) uses clang 20, because I know you are interested in LTO.

> For android it might also make sense to switch to cargo-ndk, instead of rolling our own script, but thats another topic.

Yeah, I am not sure why did not try that in the first place.

jschwe:
> BTW Rust 1.88 (so the MSRV) uses clang 20, because I know you are interested in LTO.

I'm going with clang-19 for now mainly because that's an official prebuilt "preview" version I found. It seems like compiling clang-20 from source with ohos support also works, so I might build / host that myself after some more investigation.

.github/workflows/build.yml
```diff
@@ -243,15 +245,15 @@ jobs:
         env:
           OHOS_SDK_NATIVE: ${{ steps.setup_sdk.outputs.ohos_sdk_native }}
         run: |
-          ./ohos-build cargo +${{ steps.toolchain.outputs.name }} build --target="${{ matrix.target }}" ${{ matrix.features }}
+          cargo +${{ steps.toolchain.outputs.name }} ohos build --target="${{ matrix.target }}" ${{ matrix.features }} --download-prebuilt=19
```
sagudev: I am not sure how cargo-ohos work, but would 
```suggestion
          cargo +${{ steps.toolchain.outputs.name }} ohos --download-prebuilt=19 build --target="${{ matrix.target }}" ${{ matrix.features }}
```
work? Just so that we separate cargo build and ohos args.

.github/workflows/build.yml
```diff
@@ -243,15 +245,15 @@ jobs:
...
       - name: Upload artifact
         uses: actions/upload-artifact@v7
         if: ${{ env.NEW_RUST_CHECK == 'false' && matrix.features == '' }}
         with:
           path: ./target/libmozjs-${{ matrix.target }}.tar.gz
           name: libmozjs-${{ matrix.target }}.tar.gz
       - name: Test linking (build unit-tests)
-        run: ./ohos-build cargo +${{ steps.toolchain.outputs.name }} build --target="${{ matrix.target }}" ${{ matrix.features }} --tests
+        run: cargo +${{ steps.toolchain.outputs.name }} ohos build --target="${{ matrix.target }}" ${{ matrix.features }} --tests --download-prebuilt=19
```
sagudev:
```suggestion
        run: cargo +${{ steps.toolchain.outputs.name }} ohos --download-prebuilt=19 build --target="${{ matrix.target }}" ${{ matrix.features }} --tests
```

sagudev (approved):

.github/workflows/build.yml
jschwe: I pushed a commit which uses an environment variable instead to configure `cargo-ohos`. Is that also okay?

sagudev (approved):

.github/workflows/build.yml
sagudev: This is even better.

sagudev:
Any reason for not landing this yet?

jschwe:
The companion PR isn't approved yet, and ideally both should land at the same time (since the precompiled artifact with clang-19 wouldn't work with servo-main as of now).

PR #801, opened 2026-08-22
Port the remaining public APIs to `&mut JSContext`

Gae24:
This PR completes mozjs's public API migration to `&mut JSContext`.
Only `for_of` remains, since the raw pointer is a field of `ForOfIterator` and is not used to call any methods, it can be left as is.

Testing: It compiles
Servo PR: https://github.com/servo/servo/pull/47446

### DIFF
--- a/mozjs/src/rust.rs
+++ b/mozjs/src/rust.rs
@@ -20,9 +20,9 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use self::wrappers2::{
     BuildStackString, CaptureCurrentStack, CreateRootedIdVector, CreateRootedObjectVector,
-    StackGCVectorStringAtIndex, StackGCVectorStringLength, StackGCVectorValueAtIndex,
-    StackGCVectorValueLength, ToInt32Slow, ToInt64Slow, ToNumberSlow, ToStringSlow, ToUint16Slow,
-    ToUint32Slow, ToUint64Slow,
+    JS_DefineFunctions, JS_DefineProperties, StackGCVectorStringAtIndex, StackGCVectorStringLength,
+    StackGCVectorValueAtIndex, StackGCVectorValueLength, ToInt32Slow, ToInt64Slow, ToNumberSlow,
+    ToStringSlow, ToUint16Slow, ToUint32Slow, ToUint64Slow,
 };
 use crate::consts::{JSCLASS_GLOBAL_SLOT_COUNT, JSCLASS_RESERVED_SLOTS_MASK};
 use crate::consts::{JSCLASS_IS_DOMJSCLASS, JSCLASS_IS_GLOBAL};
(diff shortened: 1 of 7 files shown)


sagudev:
> Only `for_of` remains, since the raw pointer is a field of `ForOfIterator` and is not used to call any methods, it can be left as is.

But it is used internally by `next`, which can GC. Basically ForOfIterator effectively stores &mut JSContext and it's method can at any time trigger GC. One solution to model this with ForOfIterator is simply for next to take `&mut JSContext`, but given that it's not used anywhere in servo it would be enough to fix just for_of abstraction. It should take &mut JSContext and it should pass it down to callback too. Or to be more flexible take generic `C: AsMut<JSContext>` like we are doing in https://github.com/servo/servo/pull/45684/changes

Gae24:
> > Only `for_of` remains, since the raw pointer is a field of `ForOfIterator` and is not used to call any methods, it can be left as is.
> 
> But it is used internally by `next`, which can GC. Basically ForOfIterator effectively stores &mut JSContext and it's method can at any time trigger GC. One solution to model this with ForOfIterator is simply for next to take `&mut JSContext`, but given that it's not used anywhere in servo it would be enough to fix just for_of abstraction. It should take &mut JSContext and it should pass it down to callback too. Or to be more flexible take generic `C: AsMut<JSContext>` like we are doing in https://github.com/servo/servo/pull/45684/changes

The thing is both `init` and `next` are defined inside `mozjs-sys`, we could maybe design some safer abstraction on top of it.

sagudev:
> > > Only `for_of` remains, since the raw pointer is a field of `ForOfIterator` and is not used to call any methods, it can be left as is.
> > 
> > 
> > But it is used internally by `next`, which can GC. Basically ForOfIterator effectively stores &mut JSContext and it's method can at any time trigger GC. One solution to model this with ForOfIterator is simply for next to take `&mut JSContext`, but given that it's not used anywhere in servo it would be enough to fix just for_of abstraction. It should take &mut JSContext and it should pass it down to callback too. Or to be more flexible take generic `C: AsMut<JSContext>` like we are doing in https://github.com/servo/servo/pull/45684/changes
> 
> The thing is both `init` and `next` are defined inside `mozjs-sys`, we could maybe design some safer abstraction on top of it.

for_of was designed as safer abstraction for them: https://github.com/servo/mozjs/pull/736 We could have another abstraction in between, but I think that is not needed.

mozjs/src/rust.rs
```diff
@@ -1400,7 +1401,7 @@ where
     }
 
     let mut done = false;
-    rooted!(in(cx) let mut value = UndefinedValue());
+    rooted!(in(raw_cx) let mut value = UndefinedValue());
```
```suggestion
    rooted!(&in(cx) let mut value = UndefinedValue());
```

sagudev (approved):
LGTM but needs companion PR before landing

PR #802, opened 2026-08-23
Drop deprecated code and old wrappers 

Gae24:
Testing: It compiles
Servo PR: https://github.com/servo/servo/pull/47482


mozjs/src/conversions.rs
```diff
@@ -134,7 +135,7 @@ impl_num!(f64, 0.0, f64::MIN, f64::MAX);
 /// A trait to convert Rust types to `JSVal`s.
 pub trait ToJSValConvertible {
     /// Convert `self` to a `JSVal`. JSAPI failure causes a panic.
-    fn safe_to_jsval(&self, cx: &mut crate::context::JSContext, rval: MutableHandleValue);
+    fn safe_to_jsval(&self, cx: &mut JSContext, rval: MutableHandleValue);
```
sagudev: we should also rename safe_${real_name} into ${real_name}

Gae24: Should I do `wrappers2` too?

sagudev: We could, but we can also just keep it as it's just one letter and usually only in import statements.

sagudev (approved):
LGTM, but can we wait for ESR bump to land first. Because I really do not want to do another hard rebase.

Gae24:
> LGTM, but can we wait for ESR bump to land first. Because I really do not want to do another hard rebase.

Sure!

PR #803, opened 2026-08-24
Update SpiderMonkey to 153_0esr_RELEASE

### LINKED ISSUE #687: Notes on ESR153
Continuation of #476.

- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0029-fix-ld64-detection.patch is still in progress: https://bugzilla.mozilla.org/show_bug.cgi?id=1844694
- New MozillaBuild "needs" to be packed into new moztools (not sure if we really need it, but it's nice to have)
- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0032-macos-sdk-min-14-2.patch
- https://github.com/servo/mozjs/blob/main/mozjs-sys/etc/patches/0045-python-deprecated.patch will be dropped
- Various changes to JobQueue (announcement: https://spidermonkey.dev/blog/2026/01/15/job-responsibility.html) will require changes to glue and in servo

sagudev:
There are some changes in hazard packaging in TC so we repackage to older format to avoid more work.

SM now unconditionally depends on some rust code: vendored&patched icu crates and some glue crates that bridge icu through rust to C. Because recursive cargo invocation are no go we extract rust crates from tar and import them in our tree. This means we will publish all those mozilla and patched icu crates under mozjs- prefix. We also need to provide headers from cbindgen to SM, so we reuse same pattern as we used for zlib for this.

Then there are some boring changes like requiring newer clang, NDK.

We fixed and added some more patches and some glue. The most important parts are changes to JobQueue (for microtask changes) and some additional moduleloading code. For their use see corresponding commits in servo companion PR.

Reviewable per commits. During working on this I squashed many commits otherwise rebase would be hellish.

Testing: Covered by existing tests.
Servo PR: https://github.com/servo/servo/pull/47489
Fixes #687


jschwe:
Whats the reason for removing the Linux armv7 builds? Upstream dropped support?

sagudev:
> Whats the reason for removing the Linux armv7 builds? Upstream dropped support?

No, we would just need to update cross images, but my rationale was that armv7 as an arch is covered by android and arm64-linux means that it will probably also work on linux. Ideally we should be able to run/compile it on arm64 runners anyway.

atouchet:
https://github.com/servo/mozjs/blob/main/README.md instructions are still for 140. Should that be updated to 153 as well?

mozjs-extracted-crates/collator_glue/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+license = "MPL-2.0"
+links = "collator_glue"
+name = "mozjs_collator_glue"
+version = "153.0.0"
+edition = "2021"
+rust-version.workspace = true
```
jschwe, line 9: I believe `crates.io` requires a description field. We probabluy also want to add the repository field and link it to here.

mozjs-sys/etc/update.py
```diff
@@ -26,6 +29,204 @@ def extract_tarball(tarball, commit):
...
+                cargo_toml_contents = re.sub(
+                    rf'^{original_crate} = ".*"',
+                    f'{original_crate} = {{ package = "{patched_crate}", path = "../{original_crate}" }}',
+                    cargo_toml_contents,
+                    flags=re.MULTILINE,
+                )
+                cargo_toml_contents = re.sub(
+                    rf'^{original_crate} = {{ version = ".*?"',
+                    f'{original_crate} = {{ package = "{patched_crate}", path = "../{original_crate}"',
+                    cargo_toml_contents,
```
jschwe: I think we still need a version field for crates.io publishing

jschwe:
I think we might want to still do something about HOST_CC / HOST_CXX before this lands (could be in a seperate PR that we rebase on). My suggestion would be to not just check `clang` but also search for `clang-XX` so if a user on ubuntu has a newer (non-default) clang in PATH, then we choose that as the host compiler. Ubuntu 24.04 LTS still ships with clang-18 by default, so I think if we don't do anything it would be more breaking than it needs to be. 
What do you think? I could probably make a PR.
I think the code should probably be in our `build.rs` so also embedders benefit from it (they are still free to override HOST_CC).

mozjs-sys/etc/update.py
sagudev: Ugh, you are right. I will update the script to fully support dry run.

sagudev:
> I think we might want to still do something about HOST_CC / HOST_CXX before this lands (could be in a seperate PR that we rebase on). My suggestion would be to not just check `clang` but also search for `clang-XX` so if a user on ubuntu has a newer (non-default) clang in PATH, then we choose that as the host compiler. Ubuntu 24.04 LTS still ships with clang-18 by default, so I think if we don't do anything it would be more breaking than it needs to be. What do you think? I could probably make a PR. I think the code should probably be in our `build.rs` so also embedders benefit from it (they are still free to override HOST_CC).

So you propose we populate HOST_CC with newer clang? I am not sure how I feel about this aditional logic. It would be ok if we could reuse something from cc-rs or bindgen, but I am not sure something version oriented like this exists there.

Anyway I consider those stuff non blocking as most user will use artifact anyway, meaning they will get away with older clang.

jdm:
I will have time tomorrow to review this.

sagudev:
I published dummy version of mozjs extracted crates and invited jdm as co-owner and setup trusted publishing, so it should just work on land.

mozjs-sys/Cargo.toml
```diff
@@ -22,27 +22,42 @@ jit = []
...
-  "icu_properties",
-  "icu_segmenter",
+  "segmenter",
+  "calendar",
 ], optional = true }
+mozjs_unicode_bidi_ffi = { path = "../mozjs-extracted-crates/unicode_bidi_ffi", optional = true }
+mozjs_normalizer_glue = { path = "../mozjs-extracted-crates/normalizer_glue" }
+mozjs_properties_glue = { path = "../mozjs-extracted-crates/properties_glue", optional = true }
+mozjs_collator_glue = { path = "../mozjs-extracted-crates/collator_glue", optional = true }
+mozjs_locale_glue = { path = "../mozjs-extracted-crates/locale_glue", optional = true }
```
jschwe: These also need a `version = ` for publishing mozjs_sys to work. Probably we should specify the version with `=X.Y.Z` for an exact version?

mozjs-extracted-crates/icu_collections/.cargo-checksum.json
jschwe, line 1: These `.cargo` files should be deleted, since publish is supposed to create them.

mozjs-extracted-crates/icu_collections/Cargo.toml.orig
jschwe, line 1: Same for the `Cargo.toml.orig` files, should be deleted.

mozjs-extracted-crates/collator_glue/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+description = "copy of collator_glue for mozjs-sys"
+links = "collator_glue"
+repository = "https://github.com/servo/mozjs"
+name = "mozjs_collator_glue"
+version = "153.0.0"
+edition = "2021"
+rust-version.workspace = true
+
+[dependencies]
+icu_collator = { package = "mozjs_icu_collator", version="153.0.0", path = "../icu_collator", features = ["latin1", "compiled_data", "unstable"] }
```
jschwe: These version speciifcations should be `=` to pin the exact version, or do we guarantee to be forward compatible across this SM major?

sagudev: We need to pin.

mozjs-sys/Cargo.toml
sagudev: Yes

jschwe:
For some reason I can't see `jsglue.cpp` in the github diff, but in `SavedQueue` `cx` is added as a member, and I don't see it being initialized anywhere. (This was reported in an agent review with the claim that it causes `./mach test-devtools` to segfault. I didn't check if it crashes, but I looked at the code, and definitly don't see `cx` being initialized anywhere, so seems legit)

sagudev:
> For some reason I can't see `jsglue.cpp` in the github diff, but in `SavedQueue` `cx` is added as a member, and I don't see it being initialized anywhere. (This was reported in an agent review with the claim that it causes `./mach test-devtools` to segfault. I didn't check if it crashes, but I looked at the code, and definitly don't see `cx` being initialized anywhere, so seems legit)

We run devtools test CI and it doesn't crashes, but it sure looks wrong (C++ compilers are useless).

mozjs-extracted-crates/icu_normalizer_data/Cargo.toml
```diff
@@ -0,0 +1,59 @@
...
+authors = ["The ICU4X Project Developers"]
+build = "build.rs"
+include = [
+    "data/**/*",
+    "src/**/*",
+    "examples/**/*",
+    "benches/**/*",
+    "tests/**/*",
+    "Cargo.toml",
+    "LICENSE",
```
jschwe, line 26: We probably want to preserve this license, I think it's required to include it in the distribution. 

> ... provided that either (a) this copyright and permission notice appear with all copies of the Data
Files or Software, or (b) this copyright and permission notice appear in
associated Documentation.

sagudev: This is copied from https://searchfox.org/firefox-main/source/third_party/rust/icu_normalizer_data (because it is patched) which does not have LICENCE file (it was actually removed as part of https://github.com/mozilla-firefox/firefox/commit/99a7f735b37b1bc5caba492453c00c0075d7ea55#diff-d763a343b4cb05014c03b1fbe09370f63c3d38a9e0887ca525dc5b4a9c5af252)

We do have licence field in cargo tho.

sagudev:
> > For some reason I can't see `jsglue.cpp` in the github diff, but in `SavedQueue` `cx` is added as a member, and I don't see it being initialized anywhere. (This was reported in an agent review with the claim that it causes `./mach test-devtools` to segfault. I didn't check if it crashes, but I looked at the code, and definitly don't see `cx` being initialized anywhere, so seems legit)
> 
> We run devtools test CI and it doesn't crashes, but it sure looks wrong (C++ compilers are useless).

They did crash, but they work now so good catch.

mozjs-sys/etc/patches/0032-macos-sdk-min.patch
```diff
@@ -1,13 +1,13 @@
...
 +++ b/build/moz.configure/toolchain.configure
-@@ -100,7 +100,7 @@ with only_when(host_is_osx | target_is_osx):
+@@ -252,7 +252,7 @@ with only_when(host_is_osx | target_is_osx):
      )
  
      def mac_sdk_min_version():
--        return "15.5"
-+        return "14.2"
+-        return "26.5"
++        return "26.0"
```
jdm: Why is raising the minimum from our previous patch necessary?

jdm (commented):
The changes in this PR look reasonable to me. I'm going to take a look at the servo one next.

mozjs-sys/etc/patches/0032-macos-sdk-min.patch
sagudev: I just lower it for the CI from existing requirements, but we can indeed go lower as done in https://github.com/servo/mozjs/pull/803/commits/9b381d59ddf5f1f2d08c8b22f4633cb0ba26bfdb. Hopefully this will not bite us in this ESR cycle.

jdm (approved):

PR #804, opened 2026-08-24
Fix android linking in CI

jschwe:
- Adjust the script to support macos
- Exit early if missing --target
- Use cross-compile wrappers
- Set the linker, so we can link tests

Testing: This adds linking the unit-tests for android to CI

### DIFF
--- a/android-build
+++ b/android-build
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python3
 # This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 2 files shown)


jschwe:
Okay, cherry-picked onto the 153 branch the tests still build and link, so thats good (only tested with the 28 NDK though).

android-build
sagudev, line 1: Wouldn't this be replaced by cargo-ndk?

jschwe: Yes it would be. I don't think I'll have time for working on a `cargo-ndk` PR in the close future though (it might just work, but chances are it doesn't)

PR #805, opened 2026-08-25
Expose resizable ArrayBuffer view and native Error stack APIs

Taym95:
we need these APIs to inspect RAB maxByteLength, length tracking and out-of-bounds state, and to recreate a view exactly after its buffer has shrunk, we also needs a way to preserve native Error stack text without exposing SM’s SavedFrame, and call the inline proxy check from Rust.

### DIFF
--- a/mozjs-sys/src/jsapi.cpp
+++ b/mozjs-sys/src/jsapi.cpp
@@ -17,6 +17,7 @@
 #include "js/Date.h"
 #include "js/EnvironmentChain.h"
 #include "js/Equality.h"
+#include "js/Exception.h"
 #include "js/ForOfIterator.h"
 #include "js/Id.h"
 #include "js/Initialization.h"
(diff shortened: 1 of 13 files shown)


mozjs-sys/src/jsapi.cpp
```diff
@@ -325,6 +328,21 @@ bool CreateError(JSContext* cx, JSExnType type, JS::HandleObject stack,
       rval);
 }
 
+bool CreateErrorWithOptionalCause(
```
jdm: I would move these into glue.cpp rather than requiring patching the vendored mozjs.

jdm, line 331: The changes in this file could live in jsglue.cpp instead, and don't have a strong reason to go upstream.

jdm (commented):
My main concern here is that this is a lot of new code in actual engine internals that we will need to maintain, and there are lots of parts of it that I never really had to learn anything about. This is the kind of change that I would feel much happier if it was submitted upstream.

Taym95:
> My main concern here is that this is a lot of new code in actual engine internals that we will need to maintain, and there are lots of parts of it that I never really had to learn anything about. This is the kind of change that I would feel much happier if it was submitted upstream.

I will try to make this upstream and see how SM team feels about it.

PR #806, opened 2026-08-28
glue: Use release assert for popping interrupt queue

jschwe:
The assert has side-effects, so doing this only in debug mode is not sufficient.

Testing: Not tested


sagudev:
I think we would need a devtools test that would test SavedQueue and https://github.com/servo/servo/pull/47489#discussion_r3862152060

sagudev:
Offending code was removed as part of https://github.com/servo/mozjs/pull/811

PR #810, opened 2026-09-02
Fixup update.py

sagudev:
Reviewable per commit.

- fixed handling of milestone version
- push changes even if mozilla-source exist
- commit mozjs-extracted-crates stuff
- update mozjs-extracted-crates versions in mozjs-sys Cargo.toml
- use github bot instead of servo bot for PRs (this will require close&reopen to trigger CI run per docs, but one already needs to manually approve PR anyway and current approach was complicated and used my profile for PRs).

example PR: https://github.com/servo/mozjs/pull/815 (because it was triggered from this branch it has wrong base)


sagudev:
Uh oh, I think I did a mistake here. Usually I just remove source tags/release and retrigger CI run from fix branch to test it. But now with immutable releases they got deleted only from UI, but they still exist so CI cannot push them anew. Is disabling immutable releases for this repo temporary possible solution or did I just permanently mess things up? cc @jschwe

jdm:
It's possible to disable it. I've turned it off.

sagudev:
> It's possible to disable it. I've turned it off.

Yeah, but that's just for new releases I think. Old are still immutable: https://github.com/servo/mozjs/actions/runs/33642340739/job/100297584084#step:6:33

jdm:
Yeah, I don't see anything about the release through the `gh` tool either :<

.github/workflows/security-bump.yml
```diff
@@ -23,8 +23,8 @@ jobs:
                   python-version: "3.14"
             - run: pip install requests
             - run: |
-                  git config user.name 'servo-bot'
-                  git config user.email 'ghbot+mozjs-sm-bump@servo.org'
+                  git config user.name "github-actions[bot]"
+                  git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
jschwe, line 27: Is there any specific reason we are changing this?

mozjs-sys/etc/sm-security-bump.py
```diff
@@ -22,59 +22,59 @@
...
 
-try:
-    subprocess.check_call(
-        ["gh", "release", "view", f"mozjs-source-{changeset}", "--repo", "servo/mozjs"],
-    )
-    print(f"Release mozjs-source-{changeset} already exists, skipping SM bumps")
-    sys.exit(0)
-except subprocess.CalledProcessError:
-    pass
-
```
jschwe, line 33: I'm wondering if we could keep this? We could allow `-1` postfixs after the changeset, to allow incrementing the release if we need to do fixups. Might be too complicated though, so not sure.

.github/workflows/security-bump.yml
```diff
@@ -33,8 +33,6 @@ jobs:
                   title: Security bump SpiderMonkey to ${{ steps.bump.outputs.version }}
                   body: |
                       Bump SpiderMonkey to ${{ steps.bump.outputs.version }} (changeset: [${{ steps.bump.outputs.changeset }}](https://github.com/servo/mozjs/releases/mozjs-source-${{ steps.bump.outputs.changeset }}))
-                  branch-token: ${{ secrets.SERVO_BOT }}
-                  token: ${{ secrets.PR_TOKEN }}
-                  push-to-fork: servo-bot/mozjs
-                  author: servo-bot <ghbot+mozjs-sm-bump@servo.org>
+
+                      Close and reopen this PR to trigger CI.
```
jschwe, line 37: Nit: I think if we added `actions: write` to the jobs permissions, we could run `gh workflow run main.yml --ref <branch-name>` here, which would be slightly nicer than having to close and open the PR. but not blocking.

mozjs-sys/etc/sm-security-bump.py
```diff
@@ -83,7 +83,7 @@
 
 os.remove("mozjs.tar.xz")
 
-subprocess.check_call(["git", "add", "mozjs-sys/mozjs"])
+subprocess.check_call(["git", "add", "mozjs-sys/mozjs", "mozjs-extracted-crates"])
```
jschwe: Shouldn't the extracted crates be committed as part of the update SM commit? I guess that would be in `main()`.

.github/workflows/security-bump.yml
sagudev: > use github bot instead of servo bot for PRs (this will require close&reopen to trigger CI run per docs, but one already needs to manually approve PR anyway and current approach was complicated and used my profile for PRs).

plus that's what we are doing in mozangle now so I bring it here too. It's less complicated and it's usually how things are done in github.

mozjs-sys/etc/sm-security-bump.py
sagudev: I do not think we actually need to fixup the source which just contains artifact from mozilla CI + extracted mozjs source. Implementing postfix would be hard so I think it's not worth it as all times I needed rerun mozjs-source was actually alright.

mozjs-sys/etc/sm-security-bump.py
sagudev: I think you are right

.github/workflows/security-bump.yml
sagudev: That would only do workflow_dispatch != PR checks

jschwe (approved):

PR #811, opened 2026-09-05
Store draining in JobQueue and thus simpify JobQueue

sagudev:
Store draining flag in JobQueue allows us to remove interrupt queues.
We also expose obtaining JobQueue from cx.

Testing: Tested in servo.
Servo PR: https://github.com/servo/servo/pull/47827

### DIFF
--- a/mozjs-sys/src/jsglue.cpp
+++ b/mozjs-sys/src/jsglue.cpp
@@ -51,7 +51,7 @@ struct JobQueueTraps {
       JS::MutableHandle<JSObject*> optionalHostDefinedData);
   bool (*getHostDefinedGlobal)(JSContext* cx,
                                JS::MutableHandle<JSObject*> data);
-  void (*runJobs)(RustJobQueue* queue, JSContext* cx);
+  void (*runJobs)(JSContext* cx);
   void (*traceNonGCThingMicroTask)(JSTracer* trc, JS::Value* valuePtr);
 };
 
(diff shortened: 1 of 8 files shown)


jdm (approved):
Excellent.

mozjs-sys/src/jsglue.cpp
```diff
@@ -97,79 +84,39 @@ class RustJobQueue : public JS::JobQueue {
...
+      // assert that the context’s jobQueue hasn’t been changed or cleared with
+      // SetJobQueue(nullptr) or DestroyContext().
+      MOZ_ASSERT(JS::GetJobQueue(cx) == mQueue);
 
-      // Set the current queue to mNewQueue.
-      // We need to take care of this, so that we can save the old queue in the
-      // member initializers above.
-      *mCurrentQueue = mNewQueue;
+      mSavedMicroTaskQueue = JS::SaveMicroTaskQueue(cx);
+      draining = mQueue->draining;
```
jschwe, line 94:
```suggestion
      draining = mQueue->draining;
      mQueue->draining = false;
```

During agent review of your servo PR for fixing more issues from the 153 upgrade this PR was flagged as a having a regression. The reported regression is that if the debugger pauses in a microtask, any microtask enqueued during the pause is dropped (which should cause a crash with debug-mozjs, except we don't have test coverage for this).

AI-generated regression test available in my fork at https://github.com/openharmony-rs/servo-ohos/commit/6415bcde952ecf9e261a66ee43efcdf98890931f

sagudev: Indeed, in firefox this is done in https://searchfox.org/firefox-main/rev/ae0bb53a873d228e4e61b796e184dc3c687ff82e/js/src/vm/JSContext.cpp#960

PR #816, opened 2026-09-16
Measure the null pointer as zero

mxmgorin:
This returns zero for a null pointer instead of passing it to the platform allocator. `JS::AddServoSizeOf` calls the `mallocSizeOf` hook with a null pointer, which on Windows makes `_msize` invoke the invalid parameter handler and terminate the process.

On Windows, this kills Servo when a memory report is taken with a page loaded. This was encountered in servo/servo#48016, whose test takes a memory report.

Testing: New `mozjs/tests/servo_sizes.rs`, which evaluates a small script and calls `CollectServoSizes`. Without this change, it aborts the test process on Windows.


mrobinson (approved):
Looks good to me, but maybe @sagudev should also take a look.

sagudev (approved):

